PR #704, opened 2015-05-15
Fix pixelation during drill down effect / Optimise step back effect

Ohohcakester:
Uses preloading of the zoom in image to execute a drill down without pixelation.

### DIFF
--- a/PowerPointLabs/PowerPointLabs/AutoZoom.cs
+++ b/PowerPointLabs/PowerPointLabs/AutoZoom.cs
@@ -1,8 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
-using System.Text;
-using System.Runtime.InteropServices;
 using PowerPointLabs.Models;
 using Office = Microsoft.Office.Core;
 using PowerPoint = Microsoft.Office.Interop.PowerPoint;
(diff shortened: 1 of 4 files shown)


Ohohcakester:
Made another commit - optimises step-back effect so that it is smooth as well.

nevet:
It does not work on my comp :S It throws COMException at line 715 in
Ribbon1.cs. The message says

Shapes (unknown member): Invalid request. Clipboard is empty or contains
data which may not be pasted here.

This exception is only captured in this commit.

My environment is Win8 ppt 2010.

On Friday, May 15, 2015, Oh <notifications@github.com> wrote:

> Uses preloading of the zoom in image to execute a drill down without
> pixelation.
> ------------------------------
> You can view, comment on, or merge this pull request online at:
>
>   https://github.com/PowerPointLabs/powerpointlabs/pull/704
> Commit Summary
>
>    - Fix pixelation during drill down effect
>
> File Changes
>
>    - *M* PowerPointLabs/PowerPointLabs/AutoZoom.cs
>    <https://github.com/PowerPointLabs/powerpointlabs/pull/704/files#diff-0>
>    (33)
>    - *M* PowerPointLabs/PowerPointLabs/DefaultMotionAnimation.cs
>    <https://github.com/PowerPointLabs/powerpointlabs/pull/704/files#diff-1>
>    (49)
>    - *M* PowerPointLabs/PowerPointLabs/Models/PowerPointDrillDownSlide.cs
>    <https://github.com/PowerPointLabs/powerpointlabs/pull/704/files#diff-2>
>    (25)
>
> Patch Links:
>
>    - https://github.com/PowerPointLabs/powerpointlabs/pull/704.patch
>    - https://github.com/PowerPointLabs/powerpointlabs/pull/704.diff
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/PowerPointLabs/powerpointlabs/pull/704>.
>


-- 
Best Regards,
Huang Da (Mr.)
National University of Singapore, Computer Engineering, Year 4
+65 86552730

damithc:
@Ohohcakester Keep in mind that PPT 2010 API lacks some functions available in PPT2013 API. We have to avoid such functions.

nevet:
The latest commit also triggers the same exception in drill down, but perfect for step back :)

Ohohcakester:
I haven't used any functions that aren't already in the code.
When the exception occurs, is background mode on or off?

nevet:
What do u mean by background mode?

Ohohcakester:
Whether "include slide background" is checked.
I assume it is checked then.

nevet:
Yes it is checked. Hmm very strange. Can you ask @Kai to check if it's reproducible? It's still not working here :S

Ohohcakester:
Could you pinpoint which line the exception occurs in? the multiple try/catch all's and throws makes all exceptions end up in the same place, line 715 of Ribbon.cs

nevet:
Just now I traced, and found you added a line which copies the slide. This method, I mean slide.copy(), does not work on some of the computers, and this is a known issue and was first reported by @kai. We haven't figured out the reason yet, since not all computers have this kind of issue :S A work around would be copy all shapes inside that slide instead of copy the slide.

Ohohcakester:
Noted. I'll make an equivalent method call. The currentslide.copy() call was already there in the code, I just reused it and changed the way it pasted.

Ohohcakester:
Alright, I believe it should work correctly now.

nevet:
It looks better now. However I could observe a noticeable blur->clear transition in some of the slides, but it is within tolerance :)

Ohohcakester:
Yeah, apparently saving the copy pasting the slide as a png does not make a perfect copy of the slide.
If we could copy paste the slide as an image without losing quality, then it would be perfect.

nevet:
@Ohohcakester I think this is an old issue, and was handling by @kai33. You could discuss this issue with him :)

nevet:
Btw, this level of blurriness won't be observable when you project the presentation onto the screen. Sometimes we could tolerant some level of incorrectness. In other words, if some of feature needs to use some kind of approximation, e.g. blur, un-smooth transition, let it be first until we have time to improve them :)

PR #707, opened 2015-05-19
Fix zoom to area on image shapes.

Ohohcakester:
Fix bug where only shapes, not images are replaced by transparent rectangles in zoom to area.

Previously, if you applied autozoom (zoom to area) after selecting images, the graphic of the images will not be changed into the autozoom transparent rectangle, which can be confusing to users (lack of feedback). Users may also wonder why their image is not visible in the presentation after they accidentally apply autozoom on the image and delete the autozoom slides.

### DIFF
--- a/PowerPointLabs/PowerPointLabs/ZoomToArea.cs
+++ b/PowerPointLabs/PowerPointLabs/ZoomToArea.cs
@@ -1,8 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
-using System.Text;
-using System.Runtime.InteropServices;
 using PowerPointLabs.Models;
 using Office = Microsoft.Office.Core;
 using PowerPoint = Microsoft.Office.Interop.PowerPoint;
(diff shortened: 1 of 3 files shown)


nevet:
Oh is zooming to picture an expected scenario, or is it just a corner case that hasn't been handled gracefully?

nevet:
You'd better confirm with @damithc about this issue :P

damithc:
I think this is a corner case, but still it is better to be consistent. It
is better to replace the image with a transparent rectangle.

PowerPointLabs/PowerPointLabs/ZoomToArea.cs
```diff
@@ -1,8 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Diagnostics;
```
nevet: this module is used for?

Ohohcakester:
I occasionally use it for debugging. Accidentally left it in that time (forgot to run an auto-remove usings).

PowerPointLabs/PowerPointLabs/ZoomToArea.cs
Ohohcakester: I occasionally use it for debugging. Accidentally left it in that time (forgot to run an auto-remove usings).

Issue #709, opened 2015-05-20
Text change to default color when trying to highlight bullets


Ohohcakester:
I encountered a few bugs while trying to figure out the Highlight Bullets code.
1. Occasionally, highlight bullets changes all the text to orange (instead of red, then to black). I haven't figured out what exactly causes this issue.
2. When you have a text box containing all of these three, using highlight bullets while the text box (not the text) is selected will cause an exception:
   - lines with bullet points,
   - lines without bullet points
   - empty lines (no text)

Ohohcakester:
I've fixed issue 2 in the Fixes-HighlightText branch, but I can't figure out how to fix issue 1.
Trying to set effect.EffectParameters.Color2.RGB in HighlightBulletsText.cs occasionally does nothing. Must be something going on behind the setter method.

A sequence of steps that reproduces bug 1:
 1. Select text on a slide and apply highlight bullets (not the one with the background)
 2. Undo the change.
 3. Ctrl+C, Ctrl+V on the slide to copy paste the **slide**.
 4. Select the same text on the same slide (not the copy)
 5. Apply highlight bullets on the text.

The text should turn to orange (default colour) instead of red/black. If this does not occur, undo and repeat step 5. It should be orange this time.

damithc:
@Ohohcakester is this fixed?

damithc:
Closing for now.

Issue #688, opened 2015-03-22
Open the animations pane when 'highlight points' is activated


damithc:
When the 'highlight bullets' button is clicked, we should open the animation pane if it is not already open. Otherwise, the user might no see any result of clicking the button.

Ohohcakester:
Fixed by adding a highlight bullets preview animation for feedback to the user.

PR #710, opened 2015-05-21
Fixes for Highlight Text feature

Ohohcakester:
1) Fix bug where newly-spawned highlight shapes in Highlight Text Background is sent to back, behind all other shapes. Now it is sent to just behind the text box.

2) If there are no bullet points in the text box, All the text in the text box is highlighted. (for both highlight points and highlight points background)

3) Fix bug 2 in issue #709.

4) Add feedback when highlight points is used (for issue #688)

### DIFF
--- a/PowerPointLabs/PowerPointLabs/HighlightBulletsBackground.cs
+++ b/PowerPointLabs/PowerPointLabs/HighlightBulletsBackground.cs
@@ -1,9 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
-using System.Linq;
-using System.Text;
 using System.Drawing;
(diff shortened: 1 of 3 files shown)


Ohohcakester:
Added another commit:
5) If you select a slide and click highlight points / highlight points (background), it now only highlights the text with bullet points in it.
Only when there are no text boxes with bullet points will it highlight all text.

The behaviour in the previous commit can be achieved by pressing ctrl+A in the slide and pressing highlight points. (applying highlight points to all text boxes in the slide)

nevet:
I suggest you post a picture that shows the difference in the pull request, so that we could easily tell if the change is appropriate :)

PR #711, opened 2015-05-22
More fixes for ZoomToArea

Ohohcakester:
There appears to be a bug in powerpoint where if you delete a shape programmatically then undo the change, the shape enters a weird state where if you try to access any of its properties programmatically again, a COMException occurs. I think this is mentioned in "common traps" by @kai33.
[Steps to replicate: 1) Use ZoomToArea on a rectangle. 2) Undo. 3) Use Drill Down on the same rectangle]

Apparently this issue comes about due to powerpoint not properly releasing memory after the shape is deleted
The only fix I found for this that works reliably so far is to run some mixture of ReleaseComObject and GC.Collect after the shape is deleted programmatically.

### DIFF
--- a/PowerPointLabs/PowerPointLabs/Models/PowerPointMagnifyingSlide.cs
+++ b/PowerPointLabs/PowerPointLabs/Models/PowerPointMagnifyingSlide.cs
@@ -15,6 +15,7 @@ namespace PowerPointLabs.Models
         private PowerPointMagnifyingSlide(PowerPoint.Slide slide) : base(slide)
         {
             _slide.Name = "PPTLabsMagnifyingSlide" + DateTime.Now.ToString("yyyyMMddHHmmssffff");
+            DeleteHiddenShapes();
         }
 
         new public static PowerPointSlide FromSlideFactory(PowerPoint.Slide slide)
--- a/PowerPointLabs/PowerPointLabs/Models/PowerPointSlide.cs
+++ b/PowerPointLabs/PowerPointLabs/Models/PowerPointSlide.cs
@@ -249,6 +249,15 @@ namespace PowerPointLabs.Models
             }
         }
 
+        public void DeleteHiddenShapes()
+        {
+            _slide.Shapes
+                .Cast<Shape>()
+                .Where(sh => sh.Visible == MsoTriState.msoFalse)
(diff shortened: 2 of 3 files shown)


nevet:
This is a very common situation during the development. The current work around is cut-paste the shape if exception is detected when access shape's property. But your solution looks better. I suggest you write an extension function for Shape that does this memory collection. E.g.

```C#
public static void SafeDelete(this Shape)
{
    System.Runtime.InteropServices.Marshal.ReleaseComObject(selectedShapes);
    GC.Collect();
}
```

Ohohcakester:
I planed to do that once I fully understand the issue. Only after I do that can I properly generalize the solution.
For now, it seems that the solution is to call GC.Collect() after all references to powerpoint objects have been deallocated. Looking back, I think I could do without the ReleaseComObject if I had moved GC.Collect() out of the scope of the local variable selectedShapes (after selectedShapes has been deallocated).

nevet:
Perhaps you could try ```using``` keyword to see if GC will be executed.

Ohohcakester:
It seems the thing works only if we call GC.Collect after all the relevant objects (in this case, selectedShapes) have been deallocated.

Deallocating selectedShapes by using `selectedshapes = null;` doesn't work, probably because the compiler optimizes out the deallocation.

Another solution that works is to place GC.Collect() after the ZoomToArea.AddZoomToArea() method call in Ribbon.cs.

Is there any way to add a post-user-action hook? If we can do that, I can make a SafeDelete(shape) function that adds a cleanup command to be called after the user-action function (e.g. `AddZoomToArea()`) has returned.

kai33:
> Is there any way to add a post-user-action hook? If we can do that, I can make a SafeDelete(shape) function that adds a cleanup command to be called after the user-action function (e.g. AddZoomToArea()) has returned.

This is another issue we're facing. we lack of a common place to intercept the beginning or ending of a feature's call.
Possible sol in the future can be: refactor ribbon1 to have only one method to be called from UI button. based on Office.IRibbonControl name, a factory can create the feature/lab to use.
then a common interceptor can be used at the beginning or ending and ribbon1.cs can be shortened as well

kai33:
How's this pull request? Can merge?

PR #729, opened 2015-06-24
Fix crash when autozoom buttons are pressed too quickly

Ohohcakester:


Ohohcakester:
To replicate bug:
1) click on shape
2) click on any of the AutoZoom buttons multiple times quickly.

PR #756, opened 2015-07-09
Add alternate synchronization for beam

### LINKED ISSUE #700: Renaming a section should not reorganize the agenda side bar
Once the user has organized the items the sidebar, we should try to stick to that organization.

Ohohcakester:
Now a dialog box prompts you about whether you want to reorganise the sidebar.
Fixes #700


damithc:
The title of the PR should be same as the issue it fixes. Just copy paste the entire issue title and the number, as i've done above.

Issue #759, opened 2015-07-09
Typo in error message "By confirm this ..."


damithc:
![image](https://cloud.githubusercontent.com/assets/1673303/8592856/1f1ce954-2665-11e5-98ac-2339e02aa5a1.png)

Ohohcakester:
Fixed in 642b360a55a4bb709f021b6fbe057957781f5e38

Issue #758, opened 2015-07-09
Remove 'beam' from user visible messages


damithc:
'beam' was used before we named it 'sidebar'.

Ohohcakester:
Fixed in 642b360a55a4bb709f021b6fbe057957781f5e38

PR #764, opened 2015-07-10
Typo in error message "By confirm this ..." #759, Remove 'beam' from user visible messages #758

Ohohcakester:
Rephrased some dialogue in TextCollection.cs for grammatical errors, as well as replace "beam" with "sidebar".


damithc:
In the PR description mention 'Fixes #759 Fixes #758'. That way, the relevant issue will auto-close when this is merged back to master.

damithc:
Also, go ahead and delete the branch after merging. We can undo the delete if we need later.

PR #766, opened 2015-07-11
Revise effects lab menu slightly

Ohohcakester:
No more effects lab settings. Now we have two separate buttons, 'blur remainder' and 'blur all except selected'

### DIFF
--- a/PowerPointLabs/PowerPointLabs/Ribbon1.cs
+++ b/PowerPointLabs/PowerPointLabs/Ribbon1.cs
@@ -59,9 +59,6 @@ namespace PowerPointLabs
         public bool EmbedAudioVisible = true;
         public bool RecorderPaneVisible = false;
         
-        // TODO: if this really need public
-        private bool _effectsLabSelectionIsCover = true;
-
         private bool _previewCurrentSlide;
         
         private List<string> _voiceNames;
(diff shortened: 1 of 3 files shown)


PowerPointLabs/PowerPointLabs/Ribbon1.cs
```diff
@@ -2056,13 +2134,13 @@ public void TransparentEffectClick(Office.IRibbonControl control)
...
 
-        public void EffectsLabSettingsButtonClick(Office.IRibbonControl control)
+        /*public void EffectsLabSettingsButtonClick(Office.IRibbonControl control)
         {
             var effectsLabSettingsDialog = new EffectsLabSettings(_effectsLabSelectionIsCover);
             effectsLabSettingsDialog.SettingsHandler += EffectsLabSettingChangedHandler;
             effectsLabSettingsDialog.ShowDialog();
             effectsLabSettingsDialog.Dispose();
-        }
+        }*/
```
damithc: Just delete instead of commenting. Dead code should not remain.

PowerPointLabs/PowerPointLabs/Ribbon1.cs
```diff
@@ -2092,7 +2170,7 @@ private void MagnifyGlassEffect(PowerPoint.Shape shape, float ratio)
             shape.LockAspectRatio = Office.MsoTriState.msoTrue;
         }
 
-        private PowerPointBgEffectSlide GenerateEffectSlide()
+        private PowerPointBgEffectSlide GenerateEffectSlide(bool isCover)
```
damithc: `isCover` was a temporary term. Choose something more self explanatory?

PowerPointLabs/PowerPointLabs/TextCollection.cs
```diff
@@ -98,6 +98,8 @@ internal class TextCollection
         public const string EffectsLabMagnifyGlassSupertip = @"Magnify a small area or detail on the slide.";
         public const string EffectsLabBlurRemainderSupertip = @"Draw attention to an area of the slide by blurring everything else.";
         public const string EffectsLabColorizeRemainderSupertip = @"Recolor an area of a slide to attract attention to it.";
+        public const string EffectsLabBlurBackgroundSupertip = @"Blur everything in the slide except for this shape.";
```
damithc: The feature works for multiple selections (I think). So the wording needs to change. `... except for selected shapes`?

PowerPointLabs/PowerPointLabs/Ribbon1.cs
Ohohcakester: I was keeping it there so that it would be easier to reactivate the effects settings if needed.

damithc: Dead code rarely gets revived. It's best to just delete it. We can always recover from Git if we need to. Having commented out code makes the code looks very unprofessional.

PR #770, opened 2015-07-13
Fix agenda lab sync for grouped items

Ohohcakester:
Not a very good fix, but I don't know a better way.


Ohohcakester:
Issue: If you synchronise when there is a grouped item or chart, sync will fail.
This is a workaround that fixes it, I don't know of a better way to fix it.

damithc:
Can do a dev release with this fix? I need to use it.

Ohohcakester:
The fix is present in `2.6.2.0`.

PR #779, opened 2015-08-17
Organize context menu better

### LINKED ISSUE #597: Organize context menu better
Currently the context menu is too long (require scroll) for some kind of shapes, for example a simple rectangle (Auto Shape). I think we could group up some buttons into a secondary menu, or simply put every our button into a new menu "PowerPointLabs".

kai33:
fixes #597


damithc:
This might require lot of changes to documentation.

PR #788, opened 2015-09-30
Fix  'Instruction window should display info or warning icon instead of error #784'

### LINKED ISSUE #784: Instruction window should display info or warning icon instead of error
Use case:
When user tries to install pptlabs, the Installer UI will verify whether the environment is available to install. If it's not available, the Installer UI will ask user to follow some instructions in order to make the environment available to install.

Problem:
The icon shown in the instruction window is for error, which may confuse users.
Instead, the icon should be for info or warning.

Codes at:
https://github.com/PowerPointLabs/powerpointlabs/blob/master/PowerPointLabsInstallerUi/PowerPointLabsInstallerUi/Form1.cs#L82
https://github.com/PowerPointLabs/powerpointlabs/blob/master/PowerPointLabsInstallerUi/PowerPointLabsInstallerUi/Form1.cs#L63

LIUJIAHAOCS:
change error to information


kai33:
any progress?

Issue #782, opened 2015-09-09
let Crop to Shape throw unknown exception stacktrace


kai33:
a user's powerpoint cannot use Crop to Shape (& Zoom to area) at all,
the stacktrace is needed to fix this issue.

what to do:
when an unknown exception happens, display a window with the stacktrace and ask user to send it to us

kai33:
#### related codes:
1. where we find out it's an unknown error
https://github.com/PowerPointLabs/powerpointlabs/blob/master/PowerPointLabs/PowerPointLabs/CropToShape.cs#L354
2. the dialog to display exception stacktrace
https://github.com/PowerPointLabs/powerpointlabs/blob/c7e28a7877710a7f9d81f46b7f39ced96c57800f/PowerPointLabs/PowerPointLabs/Views/ErrorDialogWrapper.cs

PR #789, opened 2015-10-05
let Crop to Shape throw unknown exception stacktrace #782

suheti:
CropToShape.cs now pops up error message window with trace stack, if the error code is not recognized.

0. Previously, GetErrorMessageForErrorCode() was used directly to show a error message box without any trace stack.
1. A new method ProcessErrorMessage() is created. This is the method, within CropToShape.cs, is now used in 3 places where GetErrorMessageForErrorCode() was used directly.
2. There are two more references to GetErrorMessageForErrorCode() outside CropToShape.cs. These are left untouched for now, since I do not yet have a complete understanding of what those references do.

### DIFF
--- a/PowerPointLabs/PowerPointLabs/CropToShape.cs
+++ b/PowerPointLabs/PowerPointLabs/CropToShape.cs
@@ -37,8 +37,6 @@ namespace PowerPointLabs
             try
             {
                 VerifyIsSelectionValid(selection);
-                //FOR DEBUG. TO BE DELETED.
-                ThrowErrorCode(250);
             }
             catch (Exception e)
             {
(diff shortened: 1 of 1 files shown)


suheti:
Here is the link to issue [#782](https://github.com/PowerPointLabs/powerpointlabs/issues/782)

PowerPointLabs/PowerPointLabs/CropToShape.cs
```diff
@@ -377,6 +378,22 @@ public static string GetErrorMessageForErrorCode(string errorCode)
             }
         }
 
+        private static void ProcessErrorMessage(Exception e)
+        {
+            //This method prompts the error message to user. If it has an unrecognised error code,
+            //an alternative message window with erro trace stack pops up and prompts the user to
```
damithc: erro --> error

PR #822, opened 2015-12-01
Fix #592 Text highlight is behind the text

### LINKED ISSUE #592: Text highlight is behind the text
Highlight test does not work if the text box has a fill color. Perhaps in such cases the highlight should be in front of the text?

LIUJIAHAOCS:


PowerPointLabs/PowerPointLabs/HighlightTextFragments.cs
```diff
@@ -41,7 +41,12 @@ public static void AddHighlightedTextFragments()
                 if (selectedText.Length <= 0) return;
                 if (selectedShapes.Count != 1) return;
 
-                List<PowerPoint.Shape> selectionToAnimate = GetShapesFromLinesInText(currentSlide, selectedText, selectedShapes[1]);
+                 if (selectedShapes.Fill.ForeColor.RGB == 16777215)
```
kai33: what is 16777215?

can move all these codes into function GetShapesFromLinesInText?

and about the naming, for a boolean var it should'd better start with is. like
isHasBackground = selectedShapes.Fill.ForeColor.RGB == ...

PowerPointLabs/PowerPointLabs/HighlightTextFragments.cs
```diff
@@ -122,6 +127,7 @@ private static List<PowerPoint.Shape> GetShapesFromLinesInText(PowerPointSlide c
                 highlightShape.Fill.ForeColor.RGB = Utils.Graphics.ConvertColorToRgb(backgroundColor);
                 highlightShape.Fill.Transparency = 0.50f;
                 highlightShape.Line.Visible = Office.MsoTriState.msoFalse;
+                if (hasBackGround)
```
kai33: no indentation?

for if/else conditions, always add the brackets {}

PR #827, opened 2015-12-10
Fix #592 Text highlight is behind the text

### LINKED ISSUE #592: Text highlight is behind the text

LIUJIAHAOCS:
Use Transparency instead of comparing forecolor

### DIFF
--- a/PowerPointLabs/PowerPointLabs/HighlightTextFragments.cs
+++ b/PowerPointLabs/PowerPointLabs/HighlightTextFragments.cs
@@ -108,7 +108,7 @@ namespace PowerPointLabs
         private static List<PowerPoint.Shape> GetShapesFromLinesInText(PowerPointSlide currentSlide, Office.TextRange2 text, PowerPoint.Shape shape)
         {
             List<PowerPoint.Shape> shapesToAnimate = new List<PowerPoint.Shape>();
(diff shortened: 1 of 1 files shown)


LIUJIAHAOCS:
image
The highlight will look like this.

PowerPointLabs/PowerPointLabs/HighlightTextFragments.cs
```diff
@@ -107,8 +107,9 @@ private static List<PowerPoint.Shape> GetShapesToAnimate(PowerPointSlide current
 
         private static List<PowerPoint.Shape> GetShapesFromLinesInText(PowerPointSlide currentSlide, Office.TextRange2 text, PowerPoint.Shape shape)
         {
-            List<PowerPoint.Shape> shapesToAnimate = new List<PowerPoint.Shape>();
-
+            List<PowerPoint.Shape> shapesToAnimate = new List<PowerPoint.Shape>();
+            Boolean isHasBackground = (shape.Fill.Transparency).CompareTo(1.0f) == 0;
```
kai33: the variable name is not good enough.
maybe isTextBoxTransparent

Issue #840, opened 2015-12-23
blurred edge can be seen at the slide edge


kai33:
it doesn't look good, requested by a user

to fix: use ppt internal cropping to crop the resulted spotlight shape (so that shape region outside of slide will be cropped)

kai33:
ref: https://github.com/PowerPointLabs/PowerPointLabs/blob/master/PowerPointLabs/PowerPointLabs/PictureSlidesLab/Service/EffectsDesigner.cs#L459

PR #893, opened 2016-01-21
#840 blurred edge can be seen at the slide edge

oswellchan:
PowerPointLabs/PowerPointLabs/issues/840

Screenshots below:
image

image


kai33:
It looks pretty good :+1: 

although its test case FT_Spotlight can be passed, pls also update its expected slide [here](https://github.com/PowerPointLabs/PowerPointLabs/blob/master/doc/test/Spotlight.pptx).

PR #894, opened 2016-01-22
Add a subtle animation to the 'duplicate' function #863

### LINKED ISSUE #863: Add a subtle animation to the 'duplicate' function
May be duplicated object can be put a few pixels off from the original location and then automatically moved to the original location (animated) so that the user sees the duplicate was created?

vhazali:
Fixes #863

### DIFF
--- a/PowerPointLabs/PowerPointLabs/Models/PowerPointSlide.cs
+++ b/PowerPointLabs/PowerPointLabs/Models/PowerPointSlide.cs
@@ -560,7 +560,7 @@ namespace PowerPointLabs.Models
         /// If you are having difficulty getting a shaperange, use the ToShapeRange method.
         /// TODO: Test this method more thoroughly in more cases other than Graphics.SquashSlides
         /// </summary>
-        public ShapeRange CopyShapesToSlide(ShapeRange shapes)
+        public ShapeRange CopyShapesToSlide(ShapeRange shapes, Boolean needOffset)
         {
             // First Index all the shapes by name, so they can be identified later.
             int index = 0;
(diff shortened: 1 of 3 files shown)


PowerPointLabs/PowerPointLabs/Models/PowerPointSlide.cs
```diff
@@ -560,7 +560,7 @@ public Shape CopyShapeToSlide(Shape shape)
         /// If you are having difficulty getting a shaperange, use the ToShapeRange method.
         /// TODO: Test this method more thoroughly in more cases other than Graphics.SquashSlides
         /// </summary>
-        public ShapeRange CopyShapesToSlide(ShapeRange shapes)
+        public ShapeRange CopyShapesToSlide(ShapeRange shapes, Boolean needOffset)
```
damithc: `needsOffset`?

vhazali: oops. Edited!

PowerPointLabs/PowerPointLabs/Models/PowerPointSlide.cs
```diff
@@ -560,7 +560,7 @@ public Shape CopyShapeToSlide(Shape shape)
         /// If you are having difficulty getting a shaperange, use the ToShapeRange method.
         /// TODO: Test this method more thoroughly in more cases other than Graphics.SquashSlides
         /// </summary>
-        public ShapeRange CopyShapesToSlide(ShapeRange shapes)
+        public ShapeRange CopyShapesToSlide(ShapeRange shapes, Boolean needsOffset)
```
kai33: Turn it into private if possible

kai33:
Pls write unit test to assert your changes. There's a readme under Test project telling how to write unit test.

PowerPointLabs/PowerPointLabs/Models/PowerPointSlide.cs
vhazali: Got it

Ohohcakester:
I am against this.
The whole purpose of the duplicate function is to duplicate without an offset.
This is one of the most difficult things to do in powerpoint.
You have to add the feedback a different way.

damithc:
The intention of the issue is not to offset the duplicate. What I wanted was to create a subtle animation that serves as a visual cue to indicate that the duplicate was created. The animation I suggested was to put the duplicate slightly off from the original position and then automatically move it to the original position in a few steps (0.2 sec per step?).

vhazali:
I see. It seems I have misunderstood the issue. I'll find another way to do the visual cue

Ohohcakester:
It would be good to read what was written in the comments before making changes. It pays to be a little more careful.

Also, avoid using Boolean in C#. The primitive type for boolean variables is "bool". Unlike java, wrappers for primitive objects, like Integer, Float, Boolean, are almost never used. (e.g. var foo = new List<int>() )

vhazali:
Yup, sorry 'bout the confusion. 
Will change the type to bool also. Thanks!

Issue #836, opened 2015-12-18
Auto Animation exception: another slide already has this name


kai33:
a user reported this:
System.Runtime.InteropServices.COMException
Slide (unknown member) : Invalid request.  Another slide already has this name.
Stack Trace:
   at Microsoft.Office.Interop.PowerPoint._Slide.set_Name(String Name)
   at PowerPointLabs.AutoAnimate.PrepareNextSlide(PowerPointSlide nextSlide)
   at PowerPointLabs.AutoAnimate.AddCompleteAnimations(PowerPointSlide currentSlide, PowerPointSlide nextSlide)
   at PowerPointLabs.AutoAnimate.AddAutoAnimation()

kai33:
https://github.com/PowerPointLabs/powerpointlabs/blob/master/PowerPointLabs/PowerPointLabs/AutoAnimate.cs#L144
https://github.com/PowerPointLabs/powerpointlabs/blob/master/PowerPointLabs/PowerPointLabs/AutoAnimate.cs#L146

should also use GUID to make the slide name more unique

PR #911, opened 2016-01-30
Auto Animation exception: another slide already has this name #836

youthinkk:
Refer to PowerPointLabs/PowerPointLabs#836


kai33:
ops forget to add it to the issue description:
pls also fix these two
https://github.com/PowerPointLabs/powerpointlabs/blob/master/PowerPointLabs/PowerPointLabs/AutoAnimate.cs#L152
https://github.com/PowerPointLabs/powerpointlabs/blob/master/PowerPointLabs/PowerPointLabs/AutoAnimate.cs#L154

and another thing:
1. pls add brackets for if-else condition
2. take substring with length 7 for the Guid.

youthinkk:
1. `GetSlideIdentifer()` is created as there are many duplicated codes. 
2. GUID with length 7 is implemented.

PR #915, opened 2016-01-30
Fix exception when removing audio #701

### LINKED ISSUE #701: Remove Audio button causes an exception
Sometimes the add audio button sometimes does not work. I have to resort to using right-click on the comment text to add audio.
After using right-click to add audio, the remove audio button causes an exception.

digawp:
How to trigger the bug:

1. Run PPL from VS in debug mode
2. Create an audio.
3. Remove the audio by clicking the Remove Audio button on the ribbon

I am unable to reproduce the issue where the Add Audio button does not work. So I am not sure if this PR should close #701


kai33:
Hi can u amend the commit msg to contain issue's title?

Issue #965, opened 2016-02-28
Make logger uses date based log file


kai33:
instead of appending everything into one file

https://github.com/PowerPointLabs/PowerPointLabs/blob/master/PowerPointLabs/PowerPointLabs/ThisAddIn.cs#L682

PR #970, opened 2016-03-04
Make logger uses date based log file #965

youthinkk:
Refer to PowerPointLabs/PowerPointLabs#965


kai33:
can provide hyphens between year/month/day?

youthinkk:
@kai33 The hyphens have been added.

PR #1021, opened 2016-05-10
Can rotate the picture in Adjust Dimensions Window #975

### LINKED ISSUE #975: Can rotate the picture in Adjust Dimensions Window

Dawnlord:
Fixes #975 

Add 4 buttons in the Adjust Dimensions Window
1. Rotate to the left 90°
2. Rotate to the right 90°
3. Flip the picture horizontally
4. Flip the picture vertically


damithc:
Post a screenshot?

Dawnlord:
The four buttons in the red box are added.

PowerPointLabs/PowerPointLabs/PictureSlidesLab/View/AdjustImageWindow.xaml.cs
```diff
@@ -256,5 +260,37 @@ private void ZoomOutButton_OnClick(object sender, RoutedEventArgs e)
         {
             _croppingAdorner.ZoomCroppingRect(-AdjustUnit);
         }
+
+        private void FilpOrRotate(RotateFlipType roatateFlipType)
+        {
+            var img = (Bitmap)Bitmap.FromFile(thumbnailImageFile.Text);
```
kai33: following SaveCropButton_OnClick(),
so use `CropResult` instead of `thumbnailImageFile.Text `

PowerPointLabs/PowerPointLabs/PictureSlidesLab/View/AdjustImageWindow.xaml.cs
```diff
@@ -256,5 +260,37 @@ private void ZoomOutButton_OnClick(object sender, RoutedEventArgs e)
...
+            img.RotateFlip(roatateFlipType);
+            String rotatedImg = StoragePath.GetPath("crop-"
```
kai33: naming `rotateFlipImg`

PowerPointLabs/PowerPointLabs/PictureSlidesLab/View/AdjustImageWindow.xaml.cs
```diff
@@ -256,5 +260,37 @@ private void ZoomOutButton_OnClick(object sender, RoutedEventArgs e)
...
+                                    + DateTime.Now.GetHashCode() + "-"
+                                    + Guid.NewGuid().ToString().Substring(0, 7));
+            img.Save(rotatedImg);
+            thumbnailImageFile.Text = rotatedImg;
```
kai33: `CropResultThumbnail` instead of `thumbnailImageFile.Text`

kai33:
does it have better icons for `flip`? it feels like `fit to width/height`

Dawnlord:
I have changed the icons for `flip`.

However, I have tried and think I should not change `thumbnailImageFile.Text`, because I need to rotate or flip the image held by the `ImageHolder`. If I just change `CropResult`, the picture shown won't be changed.

kai33:
Since rotation will change the ratio for pic's height & width, pls ensure the crop region is updated accordingly.

e.g. before rotate, the crop region may have a large width; after rotate, the picture's width can be much smaller, and so the crop region should be updated to match this

kai33:
the thumbnail is used for display only.
`CropResult` is used for applying styles to the ppt

Dawnlord:
Therefore, I should change both the display and the styles to ppt. I think I should change both thumbnail and `CropResult`. BTW, what's the purpose of `CropResultThumbnail `?

PowerPointLabs/PowerPointLabs/PictureSlidesLab/View/AdjustImageWindow.xaml.cs
```diff
@@ -256,5 +260,37 @@ private void ZoomOutButton_OnClick(object sender, RoutedEventArgs e)
         {
             _croppingAdorner.ZoomCroppingRect(-AdjustUnit);
         }
+
+        private void RotateFlipImg(RotateFlipType roatateFlipType)
+        {
+            var img = (Bitmap)Bitmap.FromFile(thumbnailImageFile.Text);
```
kai33: it should rotateFlip `CropResult `

PowerPointLabs/PowerPointLabs/PictureSlidesLab/View/AdjustImageWindow.xaml.cs
```diff
@@ -256,5 +260,37 @@ private void ZoomOutButton_OnClick(object sender, RoutedEventArgs e)
...
+            img.RotateFlip(roatateFlipType);
+            String rotatedImg = StoragePath.GetPath("crop-"
+                                    + DateTime.Now.GetHashCode() + "-"
+                                    + Guid.NewGuid().ToString().Substring(0, 7));
+            img.Save(rotatedImg);
+            thumbnailImageFile.Text = rotatedImg;
```
kai33: then generate new thumbnail to `thumbnailImageFile.Text` (for display in the window) and `CropResultThumbnail` (one of the output)

kai33:
updated comments

Dawnlord:
However, if I just rotate `CropResult`, the display in the window won't change so that the user will not see the Image rotated. Also, I have change the value of `CropResult` at the end of the function. I have tried to rotate `CropResult` in L266, but there is no effect.

kai33:
see all my comments

Dawnlord:
I will try it tomorrow, but why I cannot rotate `thumbnailImageFile.Text` directly?

kai33:
`thumbnailImageFile.Text` is a thumbnail with low resolution and quality for display only

Dawnlord:
I am still thinking about how to make the crop auto fit after every rotating, but there is some problems.

Dawnlord:
What are the differences between `SlideWidth`, `Width` and `ActualWidth` ?

damithc:
@Dawnlord After you've found out the difference, consider adding a comment to the code to clarify it.

PowerPointLabs/PowerPointLabs/PictureSlidesLab/View/AdjustImageWindow.xaml.cs
```diff
@@ -18,6 +18,7 @@ namespace PowerPointLabs.PictureSlidesLab.View
     public partial class AdjustImageWindow
     {
         private ObservableString thumbnailImageFile = new ObservableString();
+        //used for display in the adjust dimensions window only.
```
kai33: comment should be `above` the line, instead of `below`

PowerPointLabs/PowerPointLabs/PictureSlidesLab/View/AdjustImageWindow.xaml.cs
```diff
@@ -256,5 +266,87 @@ private void ZoomOutButton_OnClick(object sender, RoutedEventArgs e)
         {
             _croppingAdorner.ZoomCroppingRect(-AdjustUnit);
         }
+
+
+         private Rect RotateAutoFit()
```
kai33: how about just use `AutoFit()` to get new rect

Dawnlord:
When use `AutoFit()` directly in the `Rotate_OnClick` function, at that time the `element` (ImageHolder?) has not been updated. Therefore, at that time, the `AutoFit()` cannot work correctly. It should be executed after the `Click` action having been executed and updated the size of `ImageHolder`. That's the problem I met.

PowerPointLabs/PowerPointLabs/PictureSlidesLab/View/AdjustImageWindow.xaml
```diff
@@ -39,9 +39,13 @@
...
                 <ColumnDefinition Width="40"></ColumnDefinition>
                 <ColumnDefinition Width="40"></ColumnDefinition>
+                <ColumnDefinition Width="40"></ColumnDefinition>
+                <ColumnDefinition Width="40"></ColumnDefinition>
+                <ColumnDefinition Width="40"></ColumnDefinition>
+                <ColumnDefinition Width="40"></ColumnDefinition>
                 <ColumnDefinition Width="*"></ColumnDefinition>
             </Grid.ColumnDefinitions>
-            <Button Grid.Column="1" Style="{DynamicResource MetroCircleButtonStyle}" x:Name="MoveLeftButton" ToolTip="Move selection left" Click="MoveLeftButton_OnClick" Width="40" Height="40">
+            <Button Grid.Column="1" Style="{DynamicResource MetroCircleButtonStyle}" x:Name="MoveLeftButton" ToolTip="Move selection left" Click="MoveLeftButton_OnClick" Width="40" Height="40" HorizontalAlignment="Left" Margin="0,0,0,1">
```
kai33: is the added part necessary?

PowerPointLabs/PowerPointLabs/PictureSlidesLab/View/AdjustImageWindow.xaml
```diff
@@ -60,9 +64,21 @@
                 <Image Width="24" Height="24" Name="ZoomOutImage"></Image>
             </Button>
             <Button Grid.Column="7" Style="{DynamicResource MetroCircleButtonStyle}" x:Name="AutoFitButton" ToolTip="Auto fit picture dimensions" Click="AutoFitButton_OnClick" Width="40" Height="40">
-                <Image Width="24" Height="24" Name="AutoFitImage"></Image>
+                <Image Width="24" Height="24" x:Name="AutoFitImage_Copy"/>
```
kai33: is it necessary to rename it to `AutoFitImage_Copy `?

PowerPointLabs/PowerPointLabs/PictureSlidesLab/View/AdjustImageWindow.xaml.cs
```diff
@@ -256,5 +264,50 @@ private void ZoomOutButton_OnClick(object sender, RoutedEventArgs e)
...
+        {
+            var img = (Bitmap)Bitmap.FromFile(CropResult);
+            img.RotateFlip(roatateFlipType);
+            String rotatedImg = StoragePath.GetPath("crop-"
+                                    + DateTime.Now.GetHashCode() + "-"
+                                    + Guid.NewGuid().ToString().Substring(0, 7));
+            img.Save(rotatedImg);
+            thumbnailImageFile.Text = rotatedImg;
+            CropResult = rotatedImg;
+            CropResultThumbnail = rotatedImg;
```
kai33: use `ImageUtil.GetThumbnailFromFullSizeImg` to get thumbnail for `CropResultThumbnail` & `thumbnailImageFile.Text`

PowerPointLabs/PowerPointLabs/PictureSlidesLab/View/AdjustImageWindow.xaml.cs
```diff
@@ -256,5 +264,50 @@ private void ZoomOutButton_OnClick(object sender, RoutedEventArgs e)
...
+
+            Dispatcher.Invoke(DispatcherPriority.SystemIdle, new Action(() =>
+            {
+                var rect = AutoFit();
+                _croppingAdorner.ClippingRectangle = rect;
+            }));
+        }
+
+       /* public void RotateFlipAutoFit()
```
kai33: remove unused codes

PowerPointLabs/PowerPointLabs/PictureSlidesLab/View/AdjustImageWindow.xaml.cs
```diff
@@ -256,5 +264,50 @@ private void ZoomOutButton_OnClick(object sender, RoutedEventArgs e)
...
+            img.RotateFlip(roatateFlipType);
+            String rotatedImg = StoragePath.GetPath("crop-"
+                                    + DateTime.Now.GetHashCode() + "-"
+                                    + Guid.NewGuid().ToString().Substring(0, 7));
+            img.Save(rotatedImg);
+            thumbnailImageFile.Text = rotatedImg;
+            CropResult = rotatedImg;
+            CropResultThumbnail = rotatedImg;
+
+            Dispatcher.Invoke(DispatcherPriority.SystemIdle, new Action(() =>
```
kai33: Does `Dispatcher.BeginInvoke` (async) work for this case? Better not use `Invoke` (sync) and with this priority

Dawnlord:
Actually, I didn't notice why the name is renamed and `HorizontalAlignment="Left" Margin="0,0,0,1` appears. Maybe there is something wrong during the copy and paste. It is not necessary and I will delete it.

Also, I will update the other codes.

Dawnlord:
I have tried `Dispatcher.BeginInvoke`, and it cannot work.

Dawnlord:
Still have bugs: 
1. When close PSL and reopen it, the Image loaded is still the original one. 
2. When roate, save and close AIW, the reopened AIW is not the same size  as the old one, then the crop rigon is changed.

Dawnlord:
@kai33 I think this time it can work.

kai33:
It works as expected in general cases now 👍 

one more thing, a strange behaviour can be produced:
1. import a picture
2. rotate it in the adjust dimensions window
3. apply it to a slide
4. delete the imported picture
5. import the same picture again
6. go to the slide with rotated picture applied, which will load the picture back
7. open adjust dimensions window and see the strange result

the cause I believe is that when it tries to load back the rotated picture from the slide, it finds that the picture (not rotated one) is still in the picture collection and so stop loading the rotated picture.

Solution can be add conditions for loading back picture, or when a picture is rotated, change its full size image path also?

Dawnlord:
I am trying to change full size image path of rotated image, but there seems to be something wrong. I will continue to try it tomorrow and find out the way of loading back picture.

Dawnlord:
@kai33 I think this bug is solved.

PowerPointLabs/PowerPointLabs/PictureSlidesLab/View/PictureSlidesLabWindow.xaml.cs
```diff
@@ -966,6 +966,7 @@ private void AdjustImageDimensions(ImageItem source)
                     source.Source = CropWindow.RotateResult;
                     source.ImageFile = ImageUtil.GetThumbnailFromFullSizeImg(CropWindow.RotateResult);
                     source.FullSizeImageFile = CropWindow.RotateResult;
+                    source.ContextLink = source.ContextLink + "-Rotated";
```
kai33: does it reflect the correct ContextLink? usually it's equal to `source.FullSizeImageFile` for local image

Dawnlord:
I just find that after changing the `ContextLink`, the image is thought as different image in the program.

PR #1022, opened 2016-05-11
Support new aspect: text transparency #1014

### LINKED ISSUE #1014: Support new aspect: text transparency
Here is an example:
image

fionachang:
Fixes #1014

Also sort order of aspects and fonts in font aspect


kai33:
looks good, I'll merge it tonight

PR #1033, opened 2016-05-19
Support 'duplicate and rotate' #1002

### LINKED ISSUE #1002: Support 'duplicate and rotate'
Similar to the current 'rotate' feature, but duplicates the object before rotating it.
image

Dawnlord:
Fixes #1002


Dawnlord:
qq 20160519113649

damithc:
@vhazali @oswellchan can review this one?

Dawnlord:
Because the dash line is not clear enough, I change it back to full line.
qq 20160519150429
qq 20160519150752

Dawnlord:
qq 20160519153013

vhazali:
Hello, sorry for the slow reviewing. 

@damithc I understood the functionality as described [here](https://github.com/nus-fboa2016-PL/PowerPointLabs/issues/63), did the requirements change? The comments below are based on the requirements as mentioned in the link above.

vhazali:
The current design works well with one object to rotate and duplicating it once. However, if the user wishes to duplicate again and rotate, the user must repeat the whole process of clicking the reference object and then the object to rotate. It'll be better if there's an easier way for the user to repeat the process.

vhazali:
The current implementation also does not work well if the user wishes to duplicate and rotate more than one objects at a time. (i.e. using the image above, if the user selects the rectangle and two triangles and clicks on the duplicate & rotate, the first selected triangle is duplicated, but both are rotated). I think the optimal behaviour should be all selected shapes (except reference) are duplicated and rotated?

vhazali:
Other than that, I think the changes made are fine. It's working fine with images and shapes and text boxes. Good job!

damithc:
@vhazali we'll try to move in the direction of the original requirement, but will keep two buttons.
@Dawnlord will try to add,
1. support for multiple objects
2. support for multiple drag-to-duplicate

vhazali:
Tried it out again and it's **working for multiple objects**! Good job.

However, I think there's **some bug with the mouse listener for the multiple drag-to-duplicate** function.
The initial stage:
image

Then on clicking duplicate and rotate, the oval gets duplicated, which is correct:
image

But on rotation, when I let go of the left mouse click, another oval gets duplicated, which should not be the case (user ends up with additional set of shapes to be duplicated)
image

(The oval at the bottom is actually two ovals overlapping)

vhazali:
Nice!! I think it's working as intended now!

vhazali:
@kai33 would you like to review first before merging?

kai33:
1. `This branch has conflicts that must be resolved`
2. test?

Dawnlord:
@kai33 @vhazali  Do you know how to simulate a Clicking event of a `ImageButton` in WPF Pane? I have searched it, but did not find a good method.

Dawnlord:
I could not complete the Unit Test because of the simulation of clicking button.

vhazali:
Was this working before merging? I tried using the method but it doesn't seem to work properly

Dawnlord:
@vhazali I think you have tested it before and it works well. Are there any problems? If you think there is something wrong, we can hide its icon first before public release.

PR #1036, opened 2016-05-24
support 'lock direction' feature #1000

### LINKED ISSUE #1000: support 'lock direction' feature
Similar to lock axis, but allows movement away/toward a reference object. 
If there is only one object, it will move in the rotated direction.

image

Dawnlord:
Fixes #1000


vhazali:
Doesn't seem to work with one selected object

> If there is only one object, it will move in the rotated direction.

Dawnlord:
It can work. When you select one object and click `lock direction`, then you can pull it to the left or right to make it rotate.

vhazali:
Isn't it supposed to move in the rotated direction? 
I understood it as given just the arrow, it will still move in the direction shown in the image (without selecting the circle)
image

Dawnlord:
However, in this situation, when we choose an object at the first time, how can we or users judge which object is the rotation center?

vhazali:
Hmm, not really sure what you mean by rotation center. Are you asking how we can determine where the object is "pointing" to? Can take a look at the Snap Horizontal and Snap Vertical functions for that

damithc:
Actually, the feature is meant to be used with a reference shape. However,
a good fallback for single object (i.e. no reference) is to lock the axis
based on its rotated direction.

Dawnlord:
Therefore, when there is only one object, should it move or just rotate basing on its center axis? I did not understand rotated direction well.

damithc:
Move, not rotate. This applies to objects that are rotated by some angle
(like the arrow in the screenshot). Instead of moving
horizontally/vertically relative to the slide, it will move relative to its
rotated direction.

Dawnlord:
Ok, I will try to fix this function.

Dawnlord:
@vhazali  How about now?

vhazali:
Hi, sorry for the very long lack of response.

It's giving a quirky behavior as of now. Can I clarify what's the intended behavior for one object? 

For example, I'm testing it with these shapes
image
If i select the circle first then either or both of the arrows, they work perfectly (the arrows move in the direction they are pointing at, away and towards the circle depending on mouse movement)

However, if I select just the one arrow, the movements of the arrow are perpendicular to where they are pointing. i.e., for the right arrow, it moves as such:
image
and 
image

The same for the left arrow
image

I'm not sure if this is intended behavior. I guess it does lock the direction. If this is intended behavior then there's nothing wrong! :D

vhazali:
Does it only work with arrows? I think another obvious object with direction might be the triangle but it doesn't seem to work with it.
image

Moving mouse to right:
image

vhazali:
The Snap Vertical Icon seems to be distorted also?

Dawnlord:
Actually, when just selecting one object, it is set to move in the rotated direction, but not the direction of the arrow. I will try to go through the code and check whether I can improve it.

Dawnlord:
The bug of the text has been fixed. For the selected single object, it will only be locked by the rotated horizontal direction.
qq 20160823150019
qq 20160823150039
qq 20160823150055
qq 20160823150111

Dawnlord:
@vhazali

vhazali:
Cool! I think it works great now :D

Dawnlord:
Document and Icon have been updated.

vhazali:
@Dawnlord is there unit tests for this?

Dawnlord:
The unit test for 'lock direction' is too complex. Now I still don't know how to write it. I will solve the conflict first.

Dawnlord:
I think it may be similar to the test for 'Lock Axis', but I did not find the test for 'Lock Axis'.

Dawnlord:
There are something wrong during my merging the conflict. Need time to solve it.

Dawnlord:
Because of the error happening during solving the conflicts, I close this and rewrite one.

PR #1039, opened 2016-05-27
Add 'flip' support #1007

### LINKED ISSUE #1007: Add 'flip' support
For completeness sake, we can also support the default flip features that come with PowerPoint

Dawnlord:
Fixes #1007 

Use keyboard `Ctrl` to change the function Icon and functions.


Dawnlord:
@kai33 @vhazali Who can help review this?

vhazali:
I think the functionality wise it's fine, however the Ctrl button is supposed to change the icon also right?

This doesn't happen unless I click on the Position Lab's sidebar

i.e. if I click on the object, then I ctrl+click, it does the vertical flipping, but the icon and tooltip says vertical flipping. But if after clicking on the object, I click on the sidebar first, when I hold ctrl, then the icon changes to the vertical one.

I think this is a tad confusing for users?

Dawnlord:
I have improve the function. This time, if you move your mouse into the position lab pane, the icon will be changed by `ctrl`. I think now it would not confuse the users.

Dawnlord:
@vhazali  May you have a check?

vhazali:
Nice nice, I think the functionality is good.

Just, one thing to consider though. While I think using the `Ctrl` key is nice and all, I think it won't be intuitive for the user initially? And It's kind of not in line with how we implement the rest of PPTLab's features.

@damithc What do you think of the `Ctrl` button? If you're fine with it then I think we can go ahead with the merging

vhazali:
Almost forgot about this but, the other functions in Positions Lab and Resize Lab has a preview function if you hold `Shift` and hover over the icon. 

I think we would eventually want this function too? Not sure if it's better to merge first then add the preview or vice versa

Dawnlord:
There are still some trouble in the `preview` function for `flip`, I am still working on it and try to complete it today.

Dawnlord:
The conflict has been solved and the preview has been completed. However, the `ctrl` may make the users confusing when using preview. It still needs to be improved. The basic function is fine now.

Dawnlord:
@vhazali  I have updated the `flip` function and add `preview` for it.

vhazali:
The icon switching has a bug.
If mouse is at the side panel, pressing `ctrl` will switch the icon (between vertical and horizontal)
If I hold `ctrl + shift`, let go of `shift`, then from now on pressing `shift` will switch the icon (while holding onto `ctrl`)
This does **not** affect the actual functionality, just the icon is wrong.

vhazali:
The keys has a bug.
Mouse is at side panel.
When I preview any of the other function, by pressing `ctrl` will remove the preview. I think this is related / the same bug as the previous one

Dawnlord:
I will check it and try to think out a new way to release the changing of functions.

Dawnlord:
@vhazali  I change the way of `ctrl` working. When you press `ctrl` one time. It is changed to the other icon and function. When you press `ctrl` again, it will be changed to the original icon and function.

PR #1044, opened 2016-06-13
Support 'equalize angle' feature #1001

### LINKED ISSUE #1001: Support 'equalize angle' feature
For example, in the scenario below, equalize the angle between each arrow pair and the center of the circle
image

fionachang:
Fixes #1001

Need icon


fionachang:
May not work properly if there are shapes that overlap the center of the reference shape.
It is not able to accurately determine whether there are shapes at a particular point (shapes that visually overlaps a point (seen from the fill/line) and not the bounding box).
Only certain shapes will show error messages while others continue to execute the feature.

damithc:
What's the status of this?

fionachang:
Waiting for merge

vhazali:
Nice, seems to work perfectly as intended

PR #1045, opened 2016-06-13
Support 'resize proportionally' for area #997

### LINKED ISSUE #997: Support 'resize proportionally' for area
Currently we have support for height and width. We can support it for area as well, at least for some basic shapes such as circles, rectangles, triangles etc.

image

fionachang:
Fixes #997

Need icon
Uses bounding box area instead of shape fill area


yancmo:
Looks good! Not sure if we'll be releasing that soon though.

youthinkk:
Although the error will be prompted if two different objects are selected, the following example that both objects are images do not work properly. The reference is the squared purple image whereas the target image is in rectangle form. After entering the ratio of 1:3, the target image became square and does not maintain its aspect ratio. One possible solution is to have a check if both have same aspect ratio before changing the size of images. :smile: 

## Original Image
area_original
## Resulting Image
area_result

fionachang:
How about using the maintain aspect ratio option? When checked, aspect ratio of selected objects remain unchanged. When unchecked, aspect ratio of selected objects will be the same as the reference object (current).

PR #1047, opened 2016-06-20
Fix unknown office control id in context menu for ppt 2010

fionachang:


kai33:
did u set up the deployment environment?

PR #1055, opened 2016-07-20
effects lab: support 'frosted glass' feature #994

### LINKED ISSUE #994: effects lab: support 'frosted glass' feature
Here is an example:
image
It's already supported inside Picture Slides Lab. Just needs to provide the same feature in Effects Lab.

fionachang:
Fixes #994

- Add blurriness sub menu for Blur Selected, Blur Remainder and Blur All Except Selected
- Has options to determine blurriness level and insert tint
- Does not support hover preview


PowerPointLabs/PowerPointLabs/Ribbon1.cs
```diff
@@ -49,32 +49,56 @@ public class Ribbon1 : Office.IRibbonExtensibility
...
+        private CheckBoxActionHandlerFactory CheckBoxActionHandlerFactory { get; set; }
         #endregion
 
         #region Action Framework entry point
 
         public void OnAction(Office.IRibbonControl control)
         {
-            var actionHandler = ActionHandlerFactory.CreateInstance(control.Id);
-            actionHandler.Execute(control.Id);
+            var actionHandler = ActionHandlerFactory.CreateInstance(control.Id, control.Tag);
```
kai33: usually what's tag's value?

PowerPointLabs/PowerPointLabs/ActionFramework/Content/EffectsLabBlurrinessContentHandler.cs
```diff
@@ -0,0 +1,52 @@
...
+{
+    [ExportContentRibbonId(
+        TextCollection.EffectsLabBlurrinessFeatureSelected + TextCollection.DynamicMenuId,
+        TextCollection.EffectsLabBlurrinessFeatureRemainder + TextCollection.DynamicMenuId,
+        TextCollection.EffectsLabBlurrinessFeatureBackground + TextCollection.DynamicMenuId)]
+    class EffectsLabBlurrinessContentHandler : ContentHandler
+    {
+        protected override string GetContent(string ribbonId, string ribbonTag)
+        {
+            var feature = ribbonId.Replace(TextCollection.DynamicMenuId, "");
```
kai33: can you add comments for the documents/links for `dynamic menu`?

PowerPointLabs/PowerPointLabs/ActionFramework/Content/EffectsLabBlurrinessContentHandler.cs
```diff
@@ -0,0 +1,52 @@
...
+
+            var xmlString = new System.Text.StringBuilder("<menu xmlns=\"http://schemas.microsoft.com/office/2009/07/customui\">");
+
+            for (int i = 0; i < 7; i++)
+            {
+                xmlString.Append("<button id=\"");
```
kai33: can make it use String.Format?
and the tag xml can be extracted into constant string for menu, button, menuSeparator, check box, etc

PowerPointLabs/PowerPointLabs/ActionFramework/Action/EffectsLabBlurSelectedActionHandler.cs
```diff
@@ -1,22 +0,0 @@
-﻿using PowerPointLabs.ActionFramework.Common.Attribute;
-using PowerPointLabs.ActionFramework.Common.Extension;
-using PowerPointLabs.ActionFramework.Common.Interface;
-
-namespace PowerPointLabs.ActionFramework.Action
-{
-    [ExportActionRibbonId("EffectsLabBlurSelectedButton")]
-    class EffectsLabBlurSelectedActionHandler : ActionHandler
```
kai33: why remove

PowerPointLabs/PowerPointLabs/ActionFramework/Action/ColorsLabActionHandler.cs
```diff
@@ -7,7 +7,7 @@ namespace PowerPointLabs.ActionFramework.Action
     [ExportActionRibbonId("ColorsLabButton")]
     class ColorsLabActionHandler : ActionHandler
     {
-        protected override void ExecuteAction(string ribbonId)
```
kai33: what's the reason for changing the interface? (adding param ribbonTag)

PowerPointLabs/PowerPointLabs/ActionFramework/Common/Extension/CheckBoxActionHandlerExtensions.cs
```diff
@@ -0,0 +1,85 @@
...
+{
+    /// <summary>
+    /// Extend action handler with powerpoint context.
+    /// Type `this` in the class/subclass of CheckBoxActionHandler to access the APIs below.
+    /// </summary>
+    [Obsolete("DO NOT use this class in your feature! Used only by Action Framework.")]
+    static class CheckBoxActionHandlerExtensions
+    {
+#pragma warning disable 0618
+        public static Application GetApplication(this CheckBoxActionHandler handler)
```
kai33: Can you also do a refactoring:
all (abstract) handler implements the same interface (empty), and then just need one extensions for that interface, instead of adding extensions for every handler

PowerPointLabs/PowerPointLabs/Ribbon1.cs
fionachang: From Microsoft: If this attribute is omitted, the control's tag value SHOULD default to an empty string.

PowerPointLabs/PowerPointLabs/ActionFramework/Action/EffectsLabBlurSelectedActionHandler.cs
fionachang: I changed Blur Selected to a dynamic menu instead of a button. The dynamic menu is "Blurriness" so that Blur Remainder and Blur All Except Selected can use.

PowerPointLabs/PowerPointLabs/ActionFramework/Action/ColorsLabActionHandler.cs
fionachang: I thought I might need the ribbon tag for some functions but in the end I did not use it. I will remove it.

PowerPointLabs/PowerPointLabs/ActionFramework/Content/EffectsLabBlurrinessContentHandler.cs
fionachang: Do you mean adding comments with the url to resources about dynamic menu? Do I add the comment in this file?

vhazali:
### Missing Tool tip for the three blur options! 

### Blur Remainder
 - Creates a new slide with the intended final effects
 - Old slide has the selected shape edited (Shape fill changed to gray, and transparency set to 70%)

Would it be better to simply delete the initial slide? Is there any reason to keep it?
If we want to keep the initial slide, then I feel that it's better that if we don't change anything? I'm not sure what is the desired behavior but if the current state is what's wanted then just ignore what I just said ;)

### Blur Selected
Is there any reason that only `Blur Selected` function doesn't create a new slide?
I feel it's better if the behavior of the functions are similar?

fionachang:
The tooltip will appear after the sub-menu opens, may have to wait awhile or move the cursor slightly.
Can @damithc review the feature suggestions?

damithc:
@fionachang is this specific behavior is on the dev branch?

fionachang:
dev branch has the blur selected button without options to select blurriness or tint

damithc:
For blur remainder, we assume user wants to show the unblurred version first and then blur all slide except some area to focus attention on that area. This needs two slides.
Blur selected is simply to blur the background of a text box so that the text has a better contrast to the background elements, so there is no need for two slides.

vhazali:
Ahh. I see, cool, okay. 

In that case can I propose that the initial slide is left unchanged? Or is there no need to concern with that?

And yeah the tool tip requires the user to move the mouse after sub-menu opens. Waiting doesn't seem to do it. Mouse has to move (at least on my machine). 

I think other than that it's working great!!

damithc:
I think we modify the initial slide to make sure the shape that is used to mark the area is no longer visible during slideshow (we assume the shape was put there only to mark the area).
Is the tooltip thing different for this feature? If so, can it be fixed?

fionachang:
The tooltip uses the same function as the other features (transparent and recolor).

PR #1060, opened 2016-08-29
support 'lock direction' feature #1000

### LINKED ISSUE #1000: support 'lock direction' feature

Dawnlord:
Fixes #1000


Dawnlord:
The old one meets error when solving the conflicts, so I reopen one. @vhazali 

For the UT:
The unit test for 'lock direction' is too complex. Now I still don't know how to write it. I will solve the conflict first.
I think this test should be similar to the test of `Lock Axis`, but I did not find the test of `Lock Axis`.

vhazali:
Cool ok. But I don't think we'll be able to merge it without having tests for it. For now just by using it and manually testing it seems fine though.

IanTeo:
Closing this stale PR

PR #1079, opened 2017-01-21
Colors Lab: Add scrollbar #1078

### LINKED ISSUE #1078: Colors Lab: Panel missing scroll bar
When the window is shrunk, some of the controls can be rendered off screen and are unclickable:

image

A solution would be to add a scroll bar, similar to the Resize and Positions labs.

image

jamos-tay:
Fixes #1078

### DIFF
--- a/PowerPointLabs/PowerPointLabs/ColorPicker/ColorPane.cs
+++ b/PowerPointLabs/PowerPointLabs/ColorPicker/ColorPane.cs
@@ -67,9 +67,19 @@ namespace PowerPointLabs
             // Default color to CornFlowerBlue
             SetDefaultColor(Color.CornflowerBlue);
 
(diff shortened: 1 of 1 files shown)


kai33:
Hi Jamos,
can u post a screenshot for it?
also make sure ColorsLabTest can pass, or u'll need to update it

jamos-tay:
Here you go:

image

The test passes because PpOperations.MaximizeWindow() is called in the ColorsLabTest so the scrollbar wouldn't have any effect. However, when I took it out and shrunk the screen, it fails when the favorite color panel is offscreen since it tries to click something outside the window. This happens even without my code added.

kai33:
Yes, I added `PpOperations.MaximizeWindow` intentionally

PR #1084, opened 2017-01-22
Message in the sidebar of Shapes Lab is out of alignment #912 

### LINKED ISSUE #912: Message in the sidebar of Shapes Lab is out of alignment
As the image shown below, the message in gray does not display properly.

shapes lab - bug

initialshl:
Fixes #912 

**My claims on the problem and solution (correct me if I am wrong):**
This problem occurs on machines with DPI settings that is not 96 DPI. 
Due to the Labels being added in programmatically, the Location attributes are not scaled to DPI and not anchored properly. 
The parent Panel is also missing the AutoSize=true property, causing the Labels to be clipped.

This is tested on Windows 10, PowerPoint 2016, 192 DPI, Resolution 3200 x 1800.

**Before fix:**
* on 96 DPI
100_beforefix

* on 192 DPI
200_beforefix

**After fix:**
* on 96 DPI *(no change)*
100_afterfix

* on 192 DPI *(message shows properly)*
200_afterfix

### DIFF
--- a/PowerPointLabs/PowerPointLabs/CustomShapePane.Designer.cs
+++ b/PowerPointLabs/PowerPointLabs/CustomShapePane.Designer.cs
@@ -1,5 +1,6 @@
 ﻿using System.Drawing;
 using System.Windows.Forms;
+using PowerPointLabs.Utils;
 
 namespace PowerPointLabs
 {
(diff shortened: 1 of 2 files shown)


kai33:
looks good to me :+1:
ShapesLabTest can pass?

initialshl:
Yes it passes

PR #1085, opened 2017-01-22
Update Acknowledgement slide to new design

### LINKED ISSUE #1072: Clean up Acknowledgement Slide
The acknowledgement slide currently looks like this:
image

However, I believe that the slide could be better designed, maybe to one that is professional looking and makes use of our branding.

Skaty:
Fixes #1072.

image

Followed design that was agreed upon by @damithc.


kai33:
Hi Syed, can u provide a screenshot for it?
I'm very interested in how it looks like

Skaty:
@kai33 edited

kai33:
Looks good to me 👍 
just in case, can u ensure that all functional test can pass?
I rmb it'll try to verify the looking of slides, and changing the design may break it.

Skaty:
@kai33 I've updated the test results to consider the new ack slides

PR #1086, opened 2017-01-24
Animation Lab: Animate does not take into account shape flipping #1075

### LINKED ISSUE #1075: Animation Lab: Animate does not take into account shape flipping
This problem only occurs when shapes are modified using the flip tool or resized such that it flips. It is present on both `Add Animation Slide` and `Animate In Slide`.

**Current Behaviour**

actual

**Expected Behaviour**

expected

jamos-tay:
Fixes #1075.

The 'expected behaviour' wouldn't work because the arrow ends up upside down rather than flipped horizontally. Rather, we should animate the horizontal flip.

New behaviour:

expected

Flipped vertical:

vert

Flipped horizontal:

horiz

Flipped both ways:

both


kai33:
@damithc @Ohohcakester  is this the expected behaviour?

damithc:
@kai33 yes, it is an improvement over the current behavior. 'Flipped bothways' looks a bit odd, but still better than the current behavior.

PowerPointLabs/PowerPointLabs/FrameMotionAnimation.cs
```diff
@@ -30,6 +30,17 @@ public static void AddFrameMotionAnimation(PowerPointSlide animationSlide, Power
             float finalHeight = finalShape.Height;
             float finalFont = 0.0f;
 
+            bool flipHorizontal = initialShape.HorizontalFlip != finalShape.HorizontalFlip;
+            bool flipVertical = initialShape.VerticalFlip != finalShape.VerticalFlip;
```
kai33: bool start with `is`?

PowerPointLabs/PowerPointLabs/FrameMotionAnimation.cs
```diff
@@ -92,10 +103,10 @@ public static void AddZoomToAreaPanFrameMotionAnimation(PowerPointSlide animatio
             float incrementLeft = (finalX - initialX) / numFrames;
             float incrementTop = (finalY - initialY) / numFrames;
 
-            AddFrameAnimationEffects(animationSlide, initialShape, incrementLeft, incrementTop, incrementWidth, incrementHeight, 0.0f, 0.0f, duration, numFrames);
+            AddFrameAnimationEffects(animationSlide, initialShape, incrementLeft, incrementTop, incrementWidth, incrementHeight, false, false, 0.0f, 0.0f, duration, numFrames);
         }
 
-        private static void AddFrameAnimationEffects(PowerPointSlide animationSlide, PowerPoint.Shape initialShape, float incrementLeft, float incrementTop, float incrementWidth, float incrementHeight, float incrementRotation, float incrementFont, float duration, int numFrames)
+        private static void AddFrameAnimationEffects(PowerPointSlide animationSlide, PowerPoint.Shape initialShape, float incrementLeft, float incrementTop, float incrementWidth, float incrementHeight, bool flipHorizontal, bool flipVertical, float incrementRotation, float incrementFont, float duration, int numFrames)
```
kai33: function param, boolean starts with `is`

kai33: like isFlipHorizontally

PowerPointLabs/PowerPointLabs/FrameMotionAnimation.cs
```diff
@@ -92,10 +103,10 @@ public static void AddZoomToAreaPanFrameMotionAnimation(PowerPointSlide animatio
             float incrementLeft = (finalX - initialX) / numFrames;
             float incrementTop = (finalY - initialY) / numFrames;
 
-            AddFrameAnimationEffects(animationSlide, initialShape, incrementLeft, incrementTop, incrementWidth, incrementHeight, 0.0f, 0.0f, duration, numFrames);
+            AddFrameAnimationEffects(animationSlide, initialShape, incrementLeft, incrementTop, incrementWidth, incrementHeight, false, false, 0.0f, 0.0f, duration, numFrames);
```
kai33: add a prefix of param name to indicate the meaning of the boolean value,
e.g. if the param name is isFlipHorizontally, then `isFlipHorizontally: false`

kai33:
Hi Jamos thanks for this fix.
Can you also create a functional test to guard your changes?

Ohohcakester:
Yeah, the behaviour is expected behaviour to me.
I see scaling as a generalisation of flipping. (scale by -1 along an axis)

PowerPointLabs/PowerPointLabs/FrameMotionAnimation.cs
jamos-tay: If I add the prefix I must also add them for the other 4 params, is that okay?

image

jamos-tay:
Okay, will do.

Also, @damithc suggested that for flipping both ways, the scaling should be skewed differently for both axes so it doesn't shrink to a dot. I'll try to implement that too.

PowerPointLabs/PowerPointLabs/FrameMotionAnimation.cs
kai33: it's fine

kai33:
Hi Jamos, just rebase this pr and ensure FT still passes, then I can merge it

jamos-tay:
Hi, should be good to go

PR #1087, opened 2017-01-24
auto-animate doesn't consider line shapes #661 

### LINKED ISSUE #661: auto-animate doesn't consider line shapes
Lines and arrows are not auto-animated

initialshl:
Fixes #661 

This PR implements auto animation for straight line shapes. 
Elbow connector lines and curves are not supported as they are much more difficult to implement.

Every line shape begins at top left corner and ends at bottom right corner of its boundary box. Its rotation is always zero, and uses flip instead. Due to this behavior, some trigonometry is used to calculate the final dimensions and rotation with respect to the initial properties. 

Exisiting FTs: AutoAnimateSuccessfully, AnimateInSlideTest, Flaky_AutoAnimateWithCopyPasteSuccessfully passed.
Added FTs: AutoAnimateStraightLinesSuccessfully, AnimateInSlideStraightLinesTest passed.

ezgif com-gif-maker


kai33:
Hi Heng Le, thanks for this cool fix!
Could you help to add a piece of functional test to guard this change?

PowerPointLabs/PowerPointLabs/DefaultMotionAnimation.cs
```diff
@@ -9,12 +9,15 @@ namespace PowerPointLabs
     class DefaultMotionAnimation
     {
 #pragma warning disable 0618
+
+        private const float ArrowheadLength = 6.0f;
```
kai33: why a magic number 6 here?

initialshl: Oh thanks for pointing out the bad practice. I realized that arrowhead can have different sizes, which causes the rotation pivot to be off from the center. The new commit fixes the pivot issue by grouping a transparent line in the opposite direction.

PowerPointLabs/PowerPointLabs/FrameMotionAnimation.cs
```diff
@@ -157,5 +177,80 @@ private static void AddFrameAnimationEffects(PowerPointSlide animationSlide, Pow
...
+            List<PowerPoint.Shape> toGroup = new List<PowerPoint.Shape> { line, transparentLine };
+            PowerPoint.Shape newLine = animationSlide.ToShapeRange(toGroup).Group();
+
+            animationSlide.TransferAnimation(tempAnimationHolder, newLine);
+            tempAnimationHolder.Delete();
+
+            return newLine;
+        }
+
+        private static double GetLineAngle(PowerPoint.Shape shape, bool acute = false)
```
kai33: the acute is for?

initialshl: Its supposedly a flag on whether the method should return the angle as an acute angle, turns out i can reduce the number of trigo functions and its not needed anymore (removed in latest commit).

initialshl:
> Hi Heng Le, thanks for this cool fix!
> Could you help to add a piece of functional test to guard this change?

@kai33 Hi, I have added the functional tests for AutoAnimate and AnimateInSlide for straight lines, thanks!

Issue #1088, opened 2017-01-25
Sync Agenda: Highlighted agenda does not follow alignment of original text


Kenneth-LJS:
![image](https://cloud.githubusercontent.com/assets/21071217/22277101/e2e70be8-e2f3-11e6-930c-634ac116b8c9.png)

![image](https://cloud.githubusercontent.com/assets/21071217/22277110/e979ecfa-e2f3-11e6-8e06-07ede2b82962.png)

Skaty:
Hmm, technically it is expected behavior.

I mean, if we give users the liberty to modify the style of the highlighted agenda, wouldn't it be correct for us to ensure that the highlighted agenda follows the master slide?

kai33:
keep me updated

Kenneth-LJS:
The rationale is from a user's point of view. If they have this as an agenda:
![image](https://cloud.githubusercontent.com/assets/21071217/22314858/62d48196-e39e-11e6-9658-4db8cd0c7645.png)
The highlighting should not affect the arrangement of words
![image](https://cloud.githubusercontent.com/assets/21071217/22314883/75d56a12-e39e-11e6-903a-bbc0e396d41a.png)
![image](https://cloud.githubusercontent.com/assets/21071217/22314886/7b53c092-e39e-11e6-8c17-02bcfbd9a44b.png)
Although this might not be useful if the font or font size is different.

damithc:
@Kenneth-LJS You can explain this to me and the team during today's session and we can decide together.

damithc:
Sorry we didn't have time to discuss this Y'day.

> Hmm, technically it is expected behavior.
> I mean, if we give users the liberty to modify the style of the highlighted agenda, wouldn't it be correct for us to ensure that the highlighted agenda follows the master slide?

I feel the same way. Typically, I would use the same alignment for all agenda items and the highlighted item. If I want the alignment of the highlighted item to be different, I would change the alignment of the highlighted item in the template slide.

PR #1089, opened 2017-01-25
Sync Agenda: Highlighted agenda does not follow alignment of original text #1089

Kenneth-LJS:
#1088 Highlighted agenda will now follow alignment of the original text.


kai33:
FT passed?

PR #1092, opened 2017-01-30
Provide a way to prevent revealing future agenda items #695

### LINKED ISSUE #695: Provide a way to prevent revealing future agenda items
Currently, all agenda items are shown in the agenda slide. Some users might not want to reveal items not yet discussed.

jamos-tay:
Fixes #695.

When the user sets the text of a bullet point in the template slide to be a space, the text is removed, only displaying the bullet point.

Here's what it looks like:

image


damithc:
UX-wise, looks fine. Perhaps you can allow 'empty text' (i.e. 0 or more white space) as the indicator rather than 'exactly one space'. 
@Ohohcakester any input?

kai33:
Is it intuitive? If change the ref text to be transparent, will it work?

damithc:
> If change the ref text to be transparent, will it work?

It already works that way. The only problem is when the entire text is transparent, the bullet point becomes transparent too.

jamos-tay:
@damithc I think this code already supports any number of white spaces: `refTextRange.Text.Trim().Equals("")`

@kai33 Yeah, it doesn't change the old way. It's intuitive, but we should probably document it because users probably won't know it's there.

@Kenneth-LJS  When the bullet text is set to empty, the bullet point disappears, so we have to use a space.

Skaty:
@Kenneth-LJS I believe empty space should be preferred. As some might want to change the bullet point design for the highlighted agenda, there must be a way for us to allow this (so that the unvisited agenda has the correct bullet design).

However, do we want another kind of "hiding" where the bullet points are not shown?

damithc:
> However, do we want another kind of "hiding" where the bullet points are not shown?

Probably not in this PR. Also, I think in most cases showing how many more points remain is the right thing to do so hiding bullet point may not even be worth supporting.

jamos-tay:
Hi, is there anything else I need to do for this PR?

damithc:
> Hi, is there anything else I need to do for this PR?

Next step is to get approval from at least one peer reviewer.

damithc:
@jamos-tay also, test if the user can restore the 'future points' by adding some text (not necessarily the original text) and synchronizing again.

damithc:
@youthinkk @yancmo can one of you review this one?

jamos-tay:
> also, test if the user can restore the 'future points' by adding some text (not necessarily the original text) and synchronizing again.

Just tested this, it works fine.

damithc:
> Just tested this, it works fine.

Is it covered in automated tests? Can it be?

jamos-tay:
> Is it covered in automated tests? Can it be?

Sure, added FT

yancmo:
In my opinion, I think it's weird to see empty bullet points without any content following it. Maybe it'll be better to remove the unused points?

PowerPointLabs/PowerPointLabs/AgendaLab/AgendaLabMain.cs
```diff
@@ -644,7 +644,7 @@ private static void AdjustBulletReferenceSlideContent(PowerPointSlide refSlide)
             {
                 textRange.InsertAfter("\r ");
             }
-
+
```
yancmo: Remove this line change since it does nothing?

jamos-tay:
Whitespace removed

yancmo:
@kai33

PR #1097, opened 2017-01-31
Sync background producing many extra undo entries #754

### LINKED ISSUE #754: Sync background producing many extra undo entries
Due to the change made [here](https://github.com/PowerPointLabs/powerpointlabs/commit/2e66084659e66a784cbf9d14bc69cdd355fb80f7#diff-d89800e0309a9526cb840e137f20455eR160) to fix issue #744, a lot of unnecessary undo entries are being created.

Apparently, setting the slide background will automatically create a new undo entry. As the slide background is set for each agenda slide, a whole lot of undo entries are created when the agenda is generated/synchronised.

This is making undoing agendas a very painful process.

I can't find any way to prevent this line of code from generating undo entries.
`myFill.ForeColor.RGB = refFill.ForeColor.RGB;`

jamos-tay:
Fixes #754.

We can get around setting the background fill by saving the template slide's background to a design, then setting all agenda slides to follow the design. This doesn't trigger any extra undo entries.

### DIFF
--- a/PowerPointLabs/PowerPointLabs/Utils/Graphics.cs
+++ b/PowerPointLabs/PowerPointLabs/Utils/Graphics.cs
@@ -788,9 +788,41 @@ namespace PowerPointLabs.Utils
             Native.SendMessage(control.Handle, (uint) Native.Message.WM_SETREDRAW, new IntPtr(1), IntPtr.Zero);
             control.Refresh();
         }
-        # endregion
+        #endregion
 
-        # region Color
+        #region Design
+
+        public static Design CreateDesign(string designName)
+        {
+            return Globals.ThisAddIn.Application.ActivePresentation.Designs.Add(designName);
+        }
+
+        public static Design GetDesign(string designName)
+        {
+            foreach (Design design in Globals.ThisAddIn.Application.ActivePresentation.Designs)
+            {
+                if (design.Name.Equals(designName))
+                {
+                    return design;
+                }
+            }
+            return null;
+        }
+
+        public static void CopyToDesign(string designName, PowerPointSlide refSlide)
+        {
+            var design = GetDesign(designName);
+            if (design == null)
+            {
+                design = CreateDesign(designName);
+            }
+            design.SlideMaster.Background.Fill.ForeColor = refSlide.GetNativeSlide().Background.Fill.ForeColor;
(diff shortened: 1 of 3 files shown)


jamos-tay:
Hi @AngShiYa, thanks for the feedback.

1. I tested this on the dev-release branch, and can confirm that the same thing happens, basically it undoes the added slides first, then removes the template slides. This might be triggered by the same problem below. (Also a workaround is available since there is a Remove Agenda function)

2. Some lines of code still trigger an undo entry (Adding the ack slide, syncing bullet points, etc.), so it still requires about 5-6 undos to undo the sync. I could try to isolate the lines, but I don't think a fix is possible. However, I still think this update helps since it prevents the undo entry that is created for every single slide. This allows the sync to be undone in O(1) undos instead of O(n).

damithc:
@Ohohcakester any input?

damithc:
Looks like @Ohohcakester is not around. Can go ahead with this PR.

PowerPointLabs/PowerPointLabs/Utils/Graphics.cs
```diff
@@ -790,6 +790,38 @@ public static void ResumeDrawing(Control control)
         }
         # endregion
 
+        # region Design
+
+        public static Design CreateDesign(string designName)
+        {
+            return Globals.ThisAddIn.Application.ActivePresentation.Designs.Add(designName);
```
kai33: use PowerPointPresentation.Current

kai33: although it's deprecated FYI

PowerPointLabs/PowerPointLabs/Utils/Graphics.cs
```diff
@@ -790,6 +790,38 @@ public static void ResumeDrawing(Control control)
...
+        # region Design
+
+        public static Design CreateDesign(string designName)
+        {
+            return Globals.ThisAddIn.Application.ActivePresentation.Designs.Add(designName);
+        }
+
+        public static Design GetDesign(string designName)
+        {
+            foreach (Design design in Globals.ThisAddIn.Application.ActivePresentation.Designs)
```
kai33: use PowerPointPresentation.Current

jamos-tay:
Okay, changes made

Issue #1099, opened 2017-02-01
Narrations Lab: Remove magic numbers


Kenneth-LJS:
There are a lot of magic numbers in the narrations lab code that can be made clearer with the use of named constants.

PR #1100, opened 2017-02-01
Narrations Lab: Remove magic numbers #1099

Kenneth-LJS:
#1099 Refactor magic numbers and constants to AudioMisc.Audio


PowerPointLabs/PowerPointLabs/AudioMisc/Audio.cs
```diff
@@ -17,7 +17,11 @@ public enum AudioType
         public const int GeneratedSamplingRate = 22050;
         public const int RecordedSamplingRate = 11025;
         public const int GeneratedBitRate = 16;
-        public const int RecordedBitRate = 8;
+        public const int RecordedBitRate = 16;
+        public const int GeneratedChannels = 1;
+        public const int RecordedChannels = 1;
+        public const string RecordedFormatName = "WAVE";
+        public const string RecordedFormatExtension = "wav";
```
damithc: @AngShiYa Is this naming compliant with the coding standard? The Java coding standard requires these to be in ALL_CAPS but I'm not sure about the C# equivalent.

Skaty: @damithc According to the Naming Conventions, Constant fields use Pascal casing (first letter of every word is capitalized)

Issue #1098, opened 2017-01-31
Graphics: Create smaller version of existing lab icons


Skaty:
As discussed in #1093, we need a smaller set of icons for the various labs, so that it can be easily displayed in the sidebar of the documentation.

PR #1113, opened 2017-02-05
Design new minimized icons #1120

Kenneth-LJS:
#1098 Designed compact icons in the style of existing icons so that they can be identifiable even in the minimized window

Detailed view:
image

Drawn to scale (20x20 px):
image


damithc:
Good work @Kenneth-LJS 
Is it possible that we use a circular crop (copy icon, paste as picture, crop using PowerPoint crop feature, add outline) of the regular icon to create the smaller icon? The advantage is that there is an unmistakable link between the small and normal icons of the same lab and also no extra design work involved. Here's an example, but I'm sure you can do better.

image

Kenneth-LJS:
@damithc Here's the results
image

I don't think they look as well because the layout of the portions was designed for large rectangles, whereas if I cropped them as circles they look a little weird.

Before:
image
After:
image

Drawn to scale:
image

damithc:
Yes, that's what I meant. You can slightly adjust the placement of the circle and size of the logo to get the best 'small icon' representation of each logo. 
At the moment we have two different styles for the main logo of a lab (some have orange background while others have transparent background). Things will look even better when we migrate to one style.

Kenneth-LJS:
My icons were designed like that. For instance for narrations lab:

No nice way to get the best representation with the arrow
image

So I removed the object, and create the orange background fill
image

Then I added some line designs to make the icon more apparent
image

Similarly for highlights lab, the orange list items blended into the background:
image
image

Drawn to scale, the huge amount of details causes it to be unidentifiable
image

It's for these reasons that I didn't simply stick with the original design, but instead reworked them to be more identifiable in its scaled form. I really don't suggest using the original design unless absolutely necessary because there is too much detail to fit into a small 15x15-20x20 icon.

damithc:
For the case of highlights lab, you can enlarge the logo before cropping. e.g.
image

If you must redesign, I prefer if you only deleted stuff rather than add. Don't delete too much stuff so that the icon doesn't look like the original logo.

Kenneth-LJS:
Alright will do

Kenneth-LJS:
Icons updated

Original icons:
image


Minimized icons:
image

Drawn to scale:
image

Skaty:
The previous Narrations Lab icon looked better than the current one. The current icon does not look like it conveys the meaning of sound, rather it feels like a silent speaker icon.

damithc:
@Kenneth-LJS can go ahead

PR #1115, opened 2017-02-05
Highlight Point doesn't work with last bullet #1114

### LINKED ISSUE #1114: Highlight Lab: Highlight Point doesn't work with last bullet
To reproduce:

Create a text box with some bullet points. Ensure the cursor is at the last bullet point and that there are no empty lines following:
image

Click Highlight Points.

When run using VS, this throws an exception here:

image

When run using standalone PowerPoint, nothing happens since the exception is caught. However, this still isn't expected behavior because doing this on a paragraph in the center highlights the point:

Expected:
image     expectedhighlight

jamos-tay:
Fixes #1114.

The exception is thrown by this line:

`Globals.ThisAddIn.Application.CommandBars.ExecuteMso("AnimationPreview");`

Which causes an error when there are no animations in the slide. The last paragraph isn't creating an animation because unlike the middle paragraphs, it doesn't end with `\r` and is one character shorter so the cursor position is detected as out of the paragraph. The solution is to add 1 to the paragraph length when the paragraph is missing a `\r`.


PowerPointLabs/PowerPointLabs/HighlightBulletsText.cs
```diff
@@ -202,8 +205,11 @@ private static List<int> GetUnselectedParagraphs(Office.TextRange2 textRange, Of
                 {
                     continue;
                 }
+
+                bool hasCarriageReturn = textRange.Paragraphs[i].Text.EndsWith("\r");
+
                 if (selectedText.Start + selectedText.Length < paragraph.Start ||
-                    selectedText.Start > paragraph.Start + paragraph.Length - 1)
+                    selectedText.Start > paragraph.Start + paragraph.Length - 1 + (hasCarriageReturn ? 0 : 1))
```
Skaty: Wouldn't it be better to assign the 0/1 to `hasCarriageReturn` (maybe use a different name)  instead?

jamos-tay: Yeah I was thinking of that too but I couldn't come up with a name (`lengthToAdd`? `carriageReturnOffset`? the name needs to capture the `if (ends with \r) value = 1` relationship). Any ideas?

Skaty: Why not something like `actualParagraphLength`?

So we have something like `paragraph.Start + actualParagraphLength`?

Skaty: Why not something like `actualParagraphLength`?

So we have something like `paragraph.Start + actualParagraphLength`?

(Ignore the request for changes, GitHub thought I wanted to do a review.

Kenneth-LJS:
Works on 2013

PowerPointLabs/PowerPointLabs/HighlightBulletsText.cs
jamos-tay: Looks good, I'll make the change

PowerPointLabs/PowerPointLabs/HighlightBulletsText.cs
```diff
@@ -99,7 +99,10 @@ public static void AddHighlightBulletsText()
                     isFirstShape = false;
                 }
 
-                Globals.ThisAddIn.Application.CommandBars.ExecuteMso("AnimationPreview");
+                if (currentSlide.HasAnimationForClick(1))
```
kai33: can u help refactor the param of `Slide.HasAnimationForClick` from `click` to `clickNumber`,
and here add a prefix `clickNumber: ` for the magic number

kai33: also, try to add test to guard this PR (if possible?)

jamos-tay: Sure, I've made the changes

Issue #511, opened 2014-07-18
[story] user can see a moving magnifier near the tip of the eyedropper


damithc:
This can help the choosing of the right pixel to pick the color form

initialshl:
Can i try to work on this?

damithc:
Sure @initialshl

PR #1118, opened 2017-02-07
[story] user can see a moving magnifier near the tip of the eyedropper #511 

initialshl:
Implements #511 

This fix uses Windows Magnification API as many attempted solutions do not seem plausible.

> Attempted solution 1:
Loop: (capture screen portion -> magnify)
This will magnify the captured portion to infinity.
There is no proper way to capture portion under the magnifier control. The only (hackish) way is to traverse and paint every window handle on the system, and the taskbar area will still be black.

> Attempted solution 2:
Loop: (hide magnifier -> capture screen portion -> show magnifier) using Visible/Opacity/Position.
This causes an unbearable flicker for the magnifier.

> Attempted solution 3:
(Capture entire screen to bitmap once -> magnify portion)
The color slider on the panel will not work properly since the magnified area is frozen.

Based on these, I recommend using Windows Magnification API,
where the only downside is that Windows XP is not supported.

Windows Magnification API supports Windows Vista and above. 
No magnifier will be shown in Windows XP.

Tests: FT_ColorLabTest passed.


Kenneth-LJS:
Trying it on 2013. The preview doesn't seem to be there anymore. Previously when hovering over a certain color we can see it reflected on the shape, but now I only see the color on mouse up. Is this expected behaviour?

initialshl:
> Trying it on 2013. The preview doesn't seem to be there anymore. Previously when hovering over a certain color we can see it reflected on the shape, but now I only see the color on mouse up. Is this expected behaviour?

Hmm must be a bug. Which shape are you referring to? The square box with the hovered color?

PowerPointLabs/PowerPointLabs/ColorPicker/MagnifierForm.cs
```diff
@@ -0,0 +1,212 @@
...
+            Visible = false;
+
+            try
+            {
+                if (isMagInitialized = Native.MagInitialize())
+                {
+                    SetupMagnifier();
+                }
+            }
+            catch
```
kai33: what can be the exact exception?
must log it

PowerPointLabs/PowerPointLabs/ColorPicker/MagnifierForm.cs
```diff
@@ -0,0 +1,212 @@
...
+                Native.WC_MAGNIFIER, "MagnifierWindow", 
+                (int)Native.WindowStyles.WS_CHILD |
+                (int)Native.WindowStyles.WS_VISIBLE |
+                (int)Native.MagnifierStyle.MS_CLIPAROUNDCURSOR,
+                0, 0, Width, Height, 
+                Handle, IntPtr.Zero, hInst, IntPtr.Zero);
+
+            if (hwndMag == IntPtr.Zero)
+            {
+                throw new Exception();
```
kai33: must put the error msg

kai33: and try to use the correct exception that describes the problem instead of the generic one

PowerPointLabs/PowerPointLabs/ColorPicker/MagnifierForm.cs
```diff
@@ -0,0 +1,212 @@
...
+using System.Collections.Generic;
+using System.Drawing;
+using System.Drawing.Drawing2D;
+using System.Windows.Forms;
+
+using PPExtraEventHelper;
+
+namespace PowerPointLabs.ColorPicker
+{
+    public partial class MagnifierForm : Form
```
kai33: Can use WPF instead of WinForm for the new UI?

kai33: ref here https://github.com/PowerPointLabs/PowerPointLabs/blob/master/doc/SoftwareDesign.md#software-design
see `UI` part

initialshl: Ok thanks, I will make the changes soon

initialshl: Regarding changing from WinForm to WPF, can I do it in a separate PR? It lives in the backend since the form rendered content is replaced by Magnification API.

I tried using WPF but somehow Native.CreateWindowEx/Magnification API doesn't work nicely with WPF. Not sure if its due to window handle or what.

kai33: okie

kai33:
@Kenneth-LJS Hi could u help to verify this PR again

PR #1119, opened 2017-02-07
Change a way of identifying "recorded" and "generated" #350

### LINKED ISSUE #350: Change a way of identifying "recorded" and "generated"
Currently the two are distinguished based on bit rate: recorded = 11025, generated = 22050. This may not be a unique identifier now.

Kenneth-LJS:
Fixes #350. The type of audio will be tagged as an enum in the shape's tag under a specified key.


PowerPointLabs/PowerPointLabs/AudioMisc/Audio.cs
```diff
@@ -41,6 +45,28 @@ public Audio(string name, string saveName, int matchScriptID)
             Type = AudioHelper.GetAudioType(saveName);
         }
 
+        public Audio(Shape shape, string saveName)
+        {
+            // detect audio type from shape tag
+            AudioType audioType;
+            MessageBox.Show(shape.Tags[AudioTypeTagName]);
```
jamos-tay: Is this supposed to be here?

Kenneth-LJS: My bad :P Fixed

PowerPointLabs/PowerPointLabs/AudioMisc/Audio.cs
```diff
@@ -41,6 +45,27 @@ public Audio(string name, string saveName, int matchScriptID)
...
+            AudioType audioType;
+            if (!Enum.TryParse<AudioType>(shape.Tags[AudioTypeTagName], out audioType))
+            {
+                audioType = AudioType.Unrecognized;
+            }
+            this.Type = audioType;
+
+            // derive matched id from shape name
+            var temp = shape.Name.Split(new[] { ' ' });
+            this.MatchScriptID = Int32.Parse(temp[2]);
```
kai33: pls check the array here.
if doesn't have index 2, or if not a number, must log it and throw exception or show a friendly msg box

PowerPointLabs/PowerPointLabs/AudioMisc/Audio.cs
```diff
@@ -41,6 +45,27 @@ public Audio(string name, string saveName, int matchScriptID)
             Type = AudioHelper.GetAudioType(saveName);
         }
 
+        public Audio(Shape shape, string saveName)
+        {
+            // detect audio type from shape tag
+            AudioType audioType;
+            if (!Enum.TryParse<AudioType>(shape.Tags[AudioTypeTagName], out audioType))
```
kai33: the audio shape created previously may not have the tag, and then it will be treated as `unrecognized`. pls make sure it's backward compatible

PowerPointLabs/PowerPointLabs/AudioMisc/Audio.cs
```diff
@@ -74,6 +99,9 @@ public void EmbedOnSlide(PowerPointSlide slide, int clickNumber)
                     slide.DeleteShapesWithPrefixTimelineInvariant(Name);
 
                     audioShape.Name = shapeName;
+                    
+                    // tag item with type
+                    audioShape.Tags.Add(AudioTypeTagName, Type.ToString());
```
kai33: where did you check the `shape.MediaFormat.AudioSamplingRate`?

PowerPointLabs/PowerPointLabs/AudioMisc/Audio.cs
Kenneth-LJS: Done, added an extra check for sampling rates

PowerPointLabs/PowerPointLabs/AudioMisc/Audio.cs
Kenneth-LJS: This was from RecorderTaskPane.cs, line 436. I can make a check and indicate the format is unrecognized.

PowerPointLabs/PowerPointLabs/AudioMisc/Audio.cs
Kenneth-LJS: The type is already defined in the constructor when it's checked from the shape, line 52 of Audio.cs.

And I'm avoiding the use of sampling rate as part of the issue I'm solving.

PowerPointLabs/PowerPointLabs/AudioMisc/Audio.cs
kai33: quick question,
for a new audio shape, it won't have the audio type.
so it's still using the sampling rate to distinguish `recorded` & `generated`.
does it fix the issue?

Kenneth-LJS: When audio is recorded/generated, it creates an audio object, sets its type, then calls the function to add to slide.

So (wav file -> audio object -> shape) has no problem

The issue is that when we take a shape and create an audio object from it, it checks for the sampling rate to determine the type. So (shape -> audio object) is an issue. By setting a tag to indicate the type from audio object -> shape, we can get shape -> audio object without using the sampling rate.

kai33: can u add comments to both constructors?

PowerPointLabs/PowerPointLabs/RecorderTaskPane.cs
```diff
@@ -415,31 +415,23 @@ private void MapShapesWithAudio(PowerPointSlide slide)
...
-                        audio.Type = Audio.AudioType.Record;
-                        break;
-                    default:
-                        MessageBox.Show(TextCollection.RecorderUnrecognizeAudio);
-                        break;
+                    audio = new Audio(shape, saveName);
+                }
+                catch (FormatException ex)
+                {
+                    MessageBox.Show(ex.Message);
```
kai33: use logger to log it

kai33:
Can you also add a new UT for the class Audio?
just verify the backward compatibility (can parse Type from legacy audio shape) and whether the tag's there for new audio shape

Kenneth-LJS:
@kai33 Updated with requested changes

PR #1137, opened 2017-02-13
PasteLab: Core + Paste to Fill Slide

Skaty:
Paste to Fill Slide

Includes a common `PasteLabMain` class.


Skaty:
@initialshl I am aware of that. However, there's no consistent way of filling the slide if the shape is rotated.

initialshl:
> @initialshl I am aware of that. However, there's no consistent way of filling the slide if the shape is rotated.
I see, so its intended to work this way. Would be better if its at least positioned in the center of the slide though.

Skaty:
@IanTeo Ok, will check that one out.

Skaty:
@IanTeo @initialshl I've changed the routine such that it takes into account the visual dimensions. I found out that ResizeLab used PPShape, which is a neat abstraction.

damithc:
> As for not being able to see it, I made it such that it only works if you select an empty item. It does make sense, as the action does not need an object. It can be added, but since PowerPoint has different context menus for different selections, it would involve duplicating the menu button.

We can go with this option in the initial version. We'll revisit the decision after dogfooding the feature for a while.

PR #1142, opened 2017-02-14
PasteLab: Add core functionality and Paste to Fill Slide #1138

### LINKED ISSUE #1138: PasteLab: Add core functionality and Paste to Fill Slide
This will officially introduce PasteLab to the development releases.

Skaty:
Fixes #1138

### DIFF
--- a/PowerPointLabs/PowerPointLabs/ActionFramework/Enabled/PasteLab/PasteLabFillEnabledHandler.cs
+++ b/PowerPointLabs/PowerPointLabs/ActionFramework/Enabled/PasteLab/PasteLabFillEnabledHandler.cs
@@ -1,0 +1,16 @@
+﻿using System.Windows;
+
+using PowerPointLabs.ActionFramework.Common.Attribute;
+using PowerPointLabs.ActionFramework.Common.Interface;
+
+namespace PowerPointLabs.ActionFramework.Enabled.PasteLab
+{
+    [ExportEnabledRibbonId("pasteToFillSlide")]
+    class PasteLabFillEnabledHandler : EnabledHandler
+    {
+        protected override bool GetEnabled(string ribbonId)
+        {
+            return !(Clipboard.GetDataObject() == null);
+        }
+    }
+}
--- a/PowerPointLabs/PowerPointLabs/PasteLab/PasteLabMain.cs
+++ b/PowerPointLabs/PowerPointLabs/PasteLab/PasteLabMain.cs
@@ -1,0 +1,31 @@
+﻿using PowerPointLabs.ActionFramework.Common.Log;
+using PowerPointLabs.Utils;
+using PowerPoint = Microsoft.Office.Interop.PowerPoint;
+
+namespace PowerPointLabs.PasteLab
+{
+    public class PasteLabMain
+    {
+        public static void PasteToFillSlide(Models.PowerPointSlide slide, bool clipboardIsEmpty, float width, float height)
+        {
+            if (clipboardIsEmpty)
+            {
+                Logger.Log("PasteToFillSlide encountered empty clipboard");
+                return;
+            }
+
+            PowerPoint.ShapeRange pastedObject = slide.Shapes.Paste();
+
+            Logger.Log(string.Format("PasteToFillSlide: {0} objects pasted", pastedObject.Count));
+
+            for (int i = 1; i <= pastedObject.Count; i++)
+            {
+                var shape = new PPShape(pastedObject[i]);
+                shape.AbsoluteHeight = height;
+                shape.AbsoluteWidth = width;
+                shape.VisualTop = 0;
+                shape.VisualLeft = 0;
+            }
+        }
+    }
+}
--- a/PowerPointLabs/PowerPointLabs/ActionFramework/Action/PasteLab/PasteLabFillActionHandler.cs
+++ b/PowerPointLabs/PowerPointLabs/ActionFramework/Action/PasteLab/PasteLabFillActionHandler.cs
@@ -1,0 +1,19 @@
+﻿using System.Windows;
+
+using PowerPointLabs.ActionFramework.Common.Attribute;
+using PowerPointLabs.ActionFramework.Common.Extension;
+using PowerPointLabs.ActionFramework.Common.Interface;
+
+namespace PowerPointLabs.ActionFramework.Action.PasteLab
+{
+    [ExportActionRibbonId("pasteToFillSlide")]
+    class PasteLabFillActionHandler : ActionHandler
+    {
+        protected override void ExecuteAction(string ribbonId)
+        {
+            var curPresentation = this.GetCurrentPresentation();
+            bool clipboardIsEmpty = (Clipboard.GetDataObject() == null);
+            PowerPointLabs.PasteLab.PasteLabMain.PasteToFillSlide(this.GetCurrentSlide(), clipboardIsEmpty, curPresentation.SlideWidth, curPresentation.SlideHeight);
+        }
+    }
+}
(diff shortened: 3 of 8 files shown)


PowerPointLabs/PowerPointLabs/PasteLab/PasteLabMain.cs
```diff
@@ -0,0 +1,34 @@
...
+                shape.AbsoluteHeight = height;
+                shape.AbsoluteWidth = width;
+                shape.VisualTop = 0;
+                shape.VisualLeft = 0;
+            }
+        }
+
+        internal static bool IsClipboardEmpty()
+        {
+            return Clipboard.GetDataObject() == null;
```
kai33: is this line testable in UT?

kai33: if not, simply move this check to the handler layer to make it testable in UT

PowerPointLabs/PowerPointLabs/PasteLab/PasteLabMain.cs
```diff
@@ -0,0 +1,34 @@
+﻿using System.Windows;
+
+using PowerPointLabs.Utils;
+
+using PowerPoint = Microsoft.Office.Interop.PowerPoint;
+
+namespace PowerPointLabs.PasteLab
+{
+    public class PasteLabMain
```
kai33: any logging?
important lifecycle, check point, conditions ...

Skaty: For now, I don't see where logging will help.

PowerPointLabs/PowerPointLabs/PasteLab/PasteLabMain.cs
Skaty: It should be testable, as `Clipboard` is not a VSTO class, but in the standard C# libraries.

PowerPointLabs/PowerPointLabs/PasteLab/PasteLabMain.cs
damithc: The idea is the log might be useful if a user reports an error and we want to know what happened. The user might say something like 'PasteLab stopped working after pasting a few items' and then we want to know exactly what sequence of actions led to the error.

PowerPointLabs/PowerPointLabs/PasteLab/PasteLabMain.cs
kai33: e.g. can you mock its return value? null, or the desired object. does it make the test stable? because if test is to run in parallel, other features can use `copy()` and this may fail your lab's logic.

to me, `Clipboard.GetDataObject()` is just another input, then why not make it as a parameter explicitly.

Skaty: I see where you are coming from. Currently this method is used in the `PasteToFillSlide` method and in the Enabled handler.

There are several ways to approach this:
1. `PasteToFillSlide` takes in the status of the Clipboard (status may not reflect the actual status of the clipboard, as pasting is done here)
2. `PasteToFillSlide` assumes that the Clipboard has been checked (riskier, as we need to guard the callers)

Kenneth-LJS:
Works for most items copied to clipboard, but when I paste with a folder copied in the clipboard, I get the following error.

image

Skaty:
@Kenneth-LJS It's an edge case, though I have to agree that a meaningful error should be displayed.

PowerPointLabs/PowerPointLabs/PasteLab/PasteLabMain.cs
Skaty: So what kind of data we need to log? How intensive should it be?

Skaty:
@kai33 May I get your input on the questions asked in the review section?

PowerPointLabs/PowerPointLabs/PasteLab/PasteLabMain.cs
kai33: both looks testable to me. the 2nd one can make the interface cleaner, but why riskier?

PowerPointLabs/PowerPointLabs/PasteLab/PasteLabMain.cs
kai33: some principles i follow:
logs events in a way that, when u read the log file
1. you can understand the steps to reproduce what a user has done
2. when ur feature contains states, better record them in the lifecycle (begin, end, any state update, encounter error etc) so that to also reproduce the states
3. you know what's the error and exception (error msg, stacktrace etc) being thrown

PowerPointLabs/PowerPointLabs/PasteLab/PasteLabMain.cs
Skaty: @kai33 Okay, will implement either one. However, I believe for `alpha-release`, we won't need to write extensive tests.

PowerPointLabs/PowerPointLabs/PasteLab/PasteLabMain.cs
Skaty: Which logging function to use?

kai33: https://github.com/PowerPointLabs/PowerPointLabs/blob/master/PowerPointLabs/PowerPointLabs/ActionFramework/Common/Log/Logger.cs

Skaty: And where can we inspect the logs?

kai33: here: https://github.com/PowerPointLabs/PowerPointLabs/blob/master/PowerPointLabs/PowerPointLabs/ThisAddIn.cs#L717

Skaty:
@kai33 Logging and separation of clipboard checking has been done.

PowerPointLabs/PowerPointLabs/PasteLab/PasteLabMain.cs
```diff
@@ -0,0 +1,38 @@
...
+            {
+                var shape = new PPShape(pastedObject[i]);
+                shape.AbsoluteHeight = height;
+                shape.AbsoluteWidth = width;
+                shape.VisualTop = 0;
+                shape.VisualLeft = 0;
+            }
+        }
+        
+        public static bool IsClipboardEmpty()
```
kai33: remove this generic method and directly use it in the handler

Skaty: @kai33 How to go about ensuring that this piece of code can be reused? There's multiple places where this check is done (for instance, to determine if the button is enabled)

kai33: Im fine with just 1 line.
Or u can put it into a utility class. It's too generic and shld not be put in a specific class like PasteLabs

Skaty:
@kai33 Done! :)

PowerPointLabs/PowerPointLabs/ActionFramework/Enabled/PasteLab/PasteLabFillEnabledHandler.cs
```diff
@@ -0,0 +1,16 @@
...
+using PowerPointLabs.ActionFramework.Common.Interface;
+
+namespace PowerPointLabs.ActionFramework.Enabled.PasteLab
+{
+    [ExportEnabledRibbonId("pasteToFillSlide")]
+    class PasteLabFillEnabledHandler : EnabledHandler
+    {
+        protected override bool GetEnabled(string ribbonId)
+        {
+            return !(Clipboard.GetDataObject() == null);
```
kai33: ha, != null?

PR #1152, opened 2017-02-18
Double clicking 'Create Spotlight' causing crash #1151

### LINKED ISSUE #1151: Double clicking 'Create Spotlight' causing crash
When creating a spotlight, if the user selects a shape and double clicks 'Create Spotlight' (instead of single click), it throws an error here:

image

This is possibly due to the function being run twice.

jamos-tay:
Fixes #1151.

The error occurs because the shape is deselected by the first click, then when the second click runs with no selected shapes, it throws an error.

We can fix this by adding a check.


kai33:
ft passed?

Issue #607, opened 2014-11-21
optimize auto-zoom for pictures


damithc:
If the zoom area is covered entirely by a picture, we can place an enlarged version of the picture in the end slide (to reduce pixelation)

PR #1159, opened 2017-02-22
Optimize Magnify feature for picture #631

### LINKED ISSUE #631: Optimize Magnify feature for picture
As mentioned in #607: If the zoom area is covered entirely by a picture, we can place an enlarged version of the picture in the end slide (to reduce pixelation).

initialshl:
Fixes #631 and #607 

### Zoom To Area:

Zoom To Area (before fix):

Zoom To Area (after fix):

### Magnify Effect:

Magnify Effect (before fix):

Magnify Effect (after fix):

Limitations faced:
* Exporting pictures to original dimensions: 
    * Imported pictures are downscaled if larger than slide dimensions
    * No API to get original picture dimensions (used in "Save as picture..." context menu)
* Scaling every objects in PPT to mimic a zoom-in effect:
    * Complicated to scale all objects from an anchor point (because of text fonts, etc.)

Tests passed: FT_AutoZoomTest, FT_ZoomToAreaTest, FT_EffectsLabTest

### DIFF
--- a/PowerPointLabs/PowerPointLabs/CropToShape.cs
+++ b/PowerPointLabs/PowerPointLabs/CropToShape.cs
@@ -26,6 +26,8 @@ namespace PowerPointLabs
 
         private const string MessageBoxTitle = "Unable to crop";
 
+        private static float currentMagnifyRatio = 1.0f;
+        private const float MinMagnifyRatio = 0.1f;
         private const float MaxMagnifyRatio = 2.0f; // we don't want to export too large resolution and load for too long
 
         private static readonly string SlidePicture = Path.GetTempPath() + @"\slide.png";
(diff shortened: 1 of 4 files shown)


initialshl:
It's to make it consistent across colors lab, since float is used for the magnification factor in Magnification API (although its a different component).

initialshl:
@kai33 Hi Xie Kai, do you mind reviewing this for merging? (when u are free)

PowerPointLabs/PowerPointLabs/CropToShape.cs
```diff
@@ -243,10 +243,10 @@ private static void CreateRotatedFillInBackground(PowerPoint.Shape shape, Bitmap
             }
         }
 
-        private static void TakeScreenshotProxy(PowerPoint.ShapeRange shapeRange)
```
kai33: this magnifyRatio is passed in every method, what do you think if turn it into a state/field/member of class CropToShape

kai33:
and does it work well under different dpi?

initialshl:
Changes made. It has the same effect under different DPIs (tested 96/192/288 DPI).

PR #1160, opened 2017-02-22
Provide a way to have hierarchical agendas

### LINKED ISSUE #696: Provide a way to have hierarchical agendas
Currently, all agenda items appear at the same level. Some users might have agenda levels at multiple levels.

Skaty:
Fixes #696


damithc:
Probably need to update documentation as well (and tests).

Skaty:
@damithc Tests have already been added.

Skaty:
@IanTeo 

> Indented points do not follow the format

Explain further.

> Making font size 54 and above causes a crash

I need to investigate this. Doesn't occur in 2016, might be an issue in 10/13. Can you run `AgendaSlidesTextHierarchicalAfterSync.pptx` and check if increasing font size on the bullet point named MoreThan9 causes the bullet point to be less indented?

Skaty:
@damithc

> Indented points do not follow the format

I can't seem to fix this problem, as the PowerPoint API doesn't allow styling of background for certain part of the text (like how you can do it in the actual PowerPoint).

See: [StackOverflow Answer](https://stackoverflow.com/questions/33700487/how-do-you-set-the-fill-color-of-a-textrange)

As such, the only way to resolve this is to have the user define the style for every level on the master template.

The only styling that can be copied over is the font-related styles.

damithc:
> As such, the only way to resolve this is to have the user define the style for every level on the master template, which isn't user friendly.

In that case, how does the master template will look like?

Skaty:
@damithc 

Off the top of my head, something like this?

image

damithc:
1. You need to do that for each of the three main bullet points right?
2. Can you include lower level bullet points in the template _only_ if there are lower level headings in the actual agenda?. For example, if the agenda has two levels of headings, the template shows only two levels of bullet points. That way, it doesn't complicate the life of those who don't have lower level headings.

Skaty:
@damithc

To be fair, that is a last resort. I found a peculiarity in PowerPoint that I'm trying to exploit.

damithc:
> Anyways, that is a good idea, as it allows styling of sub points.

Exactly. It's actually a good thing.

IanTeo:
@Skaty increasing the font size does not decrease the indent

Skaty:
@IanTeo @damithc

I've updated the code such that it allows sublevel styling.

Skaty:
@IanTeo 
Please explain 2 further. What do you mean? I enlarged the size of Unvisited Bullet Point, but it only affects the unvisited ones.

IanTeo:
See the pictures I made.
Picture 1 shows Visited Bullet Format of indent level 3 to be **very large** and green and Unvisited Bullet Format of indent level 3 to be small and **black**.
But Picture 2 shows that Unvisited Bullet Format of indent level 3 is **very large and black**

Skaty:
@IanTeo 

I can't seem to replicate the issue. :| It works fine for me.

IanTeo:
So the steps to recreate the problem are as follows:

1. Open a new blank powerpoint presentation
1. Create 3 addition empty slides (You should have 1 empty slide at the beginning)
1. Create a section for each slide: Slide 2 should be in section **1.1**. Slide 3 should be in section **-1.1**. Slide 4 should be in section **-1.1.1**
1. Now create Text Agenda
1. Now rename the 3rd section (The section with the name **-1.1.1**) to **--1.1.1**.
1. Synchronize agenda. (At this point you should see the first error)
1. Change the font size of the 3rd bullet point in Visited bullet format to 54
1. Synchronize agenda. (At this point you should see the second error in slide 3)

IanTeo:
[696_test.pptx](https://github.com/PowerPointLabs/PowerPointLabs/files/896165/696_test.pptx)
@Skaty the slide I used, for your reference

IanTeo:
image
Following the steps I mentioned above, `Point 6` will cause this error message to appear now.

image
But the error where they are highlighted in the wrong color still exists

The second error seems to be resolved.

IanTeo:
@Skaty are you intending to fix this? Or is it too difficult to fix as of now?

Skaty:
@IanTeo It is doable, though I have to find time to do it, as it's not a trivial bug.

IanTeo:
@skaty ok, will keep this issue open as you work on it

IanTeo:
@Skaty I will close this issue for now. When you can around to doing it, you can reopen or create a new issue

PR #1161, opened 2017-02-23
Convert ColorPicker Magnifier to WPF

initialshl:
As previously requested in #1118 

Some improvements made from converting to WPF:
* Magnifier Border is now drawn in WPF itself rather than a separate window (MagnifierOverlay removed)
    * Now draws 1 window instead of 2 (less work)
    * Magnification API does not need to filter MagnifierOverlay anymore (less work)
* Easier to edit the dimensions directly from WPF xaml
* Magnification API win32 window is now wrapped as a WPF control (using HwndHost)

FT_ColorLabTest passed.


PowerPointLabs/PowerPointLabs/ColorPicker/Magnifier.xaml
```diff
@@ -4,22 +4,24 @@
...
-        Width="250" Height="250"
+        Name="MagnifierMainWindow"
+        AllowsTransparency="True"
         Loaded="Window_Loaded"
         Unloaded="Window_Unloaded"
-        WindowStyle="None"
-        WindowStartupLocation="Manual"
         ShowInTaskbar="False"
-        AllowsTransparency="True">
+        Topmost="True"
```
AngShiYa: Shouldn't this be TopMost?

PowerPointLabs/PowerPointLabs/ColorPicker/Magnifier.xaml.cs
```diff
@@ -12,59 +14,127 @@ namespace PowerPointLabs.ColorPicker
...
             Visibility = Visibility.Visible;
         }
 
+        #region Magnification HwndHost Win32 Interop
+
+        private class MagnificationControlHost : HwndHost
+        {
+            IntPtr hwndControl;
+            IntPtr hwndHost;
+            int hostWidth, hostHeight;
```
AngShiYa: C# default is internal, but I think it should be specified for clarity.

PowerPointLabs/PowerPointLabs/ColorPicker/Magnifier.xaml.cs
```diff
@@ -12,59 +14,127 @@ namespace PowerPointLabs.ColorPicker
...
+
+            internal const int HOST_ID = 0x00000002;
+            internal const string WC_MAGNIFIER = "Magnifier";
```
AngShiYa: Constant field should come before non-constant field and should use Pascal casing.

PowerPointLabs/PowerPointLabs/NativeMethods.cs
```diff
@@ -344,6 +366,17 @@ internal enum MagnifierFilterMode : int
...
 
+        internal enum WindowLong : int
+        {
+            GWL_WNDPROC = -4,
+            GWL_HINSTANCE = -6,
+            GWL_HWNDPARENT = -8,
+            GWL_ID = -12,
+            GWL_STYLE = -16,
+            GWL_EXSTYLE = -20,
+            GWL_USERDATA = -21
```
AngShiYa: Enum should be in Pascal casing

PowerPointLabs/PowerPointLabs/ColorPicker/Magnifier.xaml
initialshl: WPF uses Topmost instead

PowerPointLabs/PowerPointLabs/NativeMethods.cs
initialshl: I would suggest keeping it in uppercase (same as other enum in Native class) to keep it the same as the ones defined in Winuser.h
https://msdn.microsoft.com/en-us/library/windows/desktop/ms633591(v=vs.85).aspx

initialshl:
Thanks for the review

initialshl:
Ok will do, thanks

IanTeo:
If you can't fix the problem, maybe you can disable this feature for 2010. As of now, I use 2010 anymore once I select ColorPicker

initialshl:
Ok, disabled for PPT 2010 for now (until i fix the bug).

AngShiYa:
Tried testing on 2013 but ColorsLabTest failed. I think there is some bug with the scaling that displace the positioning in the magnifier.

ezgif-3-71c297722f

initialshl:
The position bug should be fixed now, i will test it in the lab's pc again.

IanTeo:
@initialshl have you tested on the lab PC? I tested it on the lab PC and found that the bug only occurs when no shape is selected

initialshl:
> @initialshl have you tested on the lab PC? I tested it on the lab PC and found that the bug only occurs when no shape is selected

Do u mean the winform version where magnifier cannot be closed? The WPF should have it fixed now. But I am currently having trouble with the WPF on PPT 2010, it still doesnt appear

initialshl:
@IanTeo @AngShiYa All the existing bugs should be fixed now, please help me review again when u have the time, thanks!

PR #1167, opened 2017-03-02
Help button redirecting to old docs site #1165

### LINKED ISSUE #1165: Help button redirecting to old docs site
The help button in the Ribbon currently redirects to the old docs html:

http://www.comp.nus.edu.sg/~pptlabs//docs.html

This should be the correct one:

http://www.comp.nus.edu.sg/~pptlabs/docs/

jamos-tay:
Fixes #1165.

Corrected docs link.
Also, 'Contact Us' is redirecting to
http://www.comp.nus.edu.sg/~pptlabs//contact.html
instead of
http://www.comp.nus.edu.sg/~pptlabs/contact.html
so changed to direct url.

### DIFF
--- a/PowerPointLabs/PowerPointLabs/TextCollection.cs
+++ b/PowerPointLabs/PowerPointLabs/TextCollection.cs
@@ -9,8 +9,8 @@
         # endregion
 
         # region URLs
-        public const string FeedbackUrl = "http://powerpointlabs.info/contact.html";
-        public const string HelpDocumentUrl = "http://powerpointlabs.info/docs.html";
+        public const string FeedbackUrl = "http://www.comp.nus.edu.sg/~pptlabs/contact.html";
+        public const string HelpDocumentUrl = "http://www.comp.nus.edu.sg/~pptlabs/docs/";
         public const string PowerPointLabsWebsiteUrl = "http://PowerPointLabs.info";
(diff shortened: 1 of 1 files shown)


damithc:
> Why use www.comp.nus.edu.sg/~pptlabs instead of powerpointlabs.info?

Originally we were not sure where to host the website, so we gave the generic link. I guess better to use `www.comp.nus.edu.sg/~pptlabs` (faster)

Skaty:
@damithc Hmm, sure.

jamos-tay:
@Skaty Ah, the powerpointlabs.info link was adding an extra slash (e.g. ~pptlabs//contact.html instead of ~pptlabs/contact.html), probably due to the way the redirecting works.

damithc:
> Ah, the powerpointlabs.info link was adding an extra slash (e.g. ~pptlabs//contact.html instead of ~pptlabs/contact.html), probably due to the way the redirecting works.

I removed the extra slash from the forwarding settings. Should take effect after a while.

Skaty:
@jamos-tay It wasn't really affecting it that much though.

IanTeo:
@Skaty I don't see a problem with this PR. After all, changing the link to use a direct URL is faster, and it does not affect anything else. Maybe you could approve the review, unless you have other reasons why we should use the old URL?

Issue #713, opened 2015-05-24
'Sync agenda' does not respect height of text box


damithc:
In this case, I've tweaked the height of some items (because the text is longer) like this:
![image](https://cloud.githubusercontent.com/assets/1673303/7787003/717c674c-0225-11e5-9f2d-28d2cd2c133b.png)

After the sync, it ended up like this:
![image](https://cloud.githubusercontent.com/assets/1673303/7786998/3724b9e6-0225-11e5-8123-7706d9d0064e.png)

damithc:
Now the text box size is respected. But line indentation is not. 
![image](https://cloud.githubusercontent.com/assets/1673303/8568350/7ad5cc78-259f-11e5-8485-b6d0c7953a3f.png)

damithc:
Check if this issue is still relevant. Some parts of AgendaLab was rewritten since the issue was posted.

Kenneth-LJS:
![image](https://cloud.githubusercontent.com/assets/21071217/22234645/9f39f172-e234-11e6-81ed-604769626e75.png)

![image](https://cloud.githubusercontent.com/assets/21071217/22234651/a7a46d4c-e234-11e6-9faf-ce9f529d99a3.png)

Alignment isn't followed for highlighted agenda.

damithc:
I see. Go ahead and try to fix it.

damithc:
Seems the original issue is fixed. Closing.

PR #1168, opened 2017-03-03
Corrected unaligned highlighted agenda titles on sync

Dartteon:
Previously, highlighting the agenda title on each respective page caused it to lose its alignment settings as the properties of the Highlight textbox is copied instead of the title's textbox itself, in `Graphics.SyncTextRange`.

I saved the Alignment property of the title textBox, and reassigned it after passing it through `Graphics.SyncTextRange`.

image


Dartteon:
Oh brought up a valid point, that the user may actually be expecting the highlighting to affect the alignment on the respective pages. Reviewer may want to look if this is a UX case or a bug.

Kenneth-LJS:
I've tried to fix this before XP #1088

Dartteon:
Oh no.. alright then.

There wasn't any indication in the thread that this issue was being dealt with in another thread, so I presumed it was still not fixed. My bad for not looking through all the issues beforehand.

Should #713 be closed, then?

Dartteon:
Since issue is confirmed as not a bug and has already found a solution in another PR, I will be closing this one.

PR #1173, opened 2017-03-08
Sync Lab GUI

Kenneth-LJS:
For dogfooding


Kenneth-LJS:
Those were changes from dev release :P Branched from dev-release, but the alpha branch started from before that.

PR #1174, opened 2017-03-08
Sync Lab GUI

Kenneth-LJS:
For dogfooding


IanTeo:
Sorry @Kenneth-LJS, can you check PR #1175 which I just made? Then if it OK, you can close this, I think I added all your parts in already.

PR #1176, opened 2017-03-08
Add Blocks Timer Prototype

AngShiYa:
Add prototype for Blocks type timer. This is a MVP that only supports 1 minute timers. I plan to make the timer configurable in subsequent updates


Kenneth-LJS:
Works on 2013

PR #1185, opened 2017-03-12
Crop to same dimension: dimensions don't match #1180

### LINKED ISSUE #1180: Crop to same dimension: dimensions don't match
image

1. After applying (result shown above), the lower image dimension did not match the top one.
2. If I click the menu item again, that resulted in different output (this feature should give the same output if applied multiple times).

jamos-tay:
Fixes #1180

Should've been += instead of +.

Fixes the following:
Dimensions not matching when applying crop to same on an image that is already cropped
Added test for the above case
Pressing crop to same multiple times giving different results


Skaty:
@jamos-tay Resolve the conflicts (through rebase).

jamos-tay:
Conflicts resolved

PowerPointLabs/PowerPointLabs/CropLab/CropToSame.cs
```diff
@@ -15,25 +15,25 @@ public static void CropSelection(PowerPoint.ShapeRange shapeRange)
             float refScaleWidth = Graphics.GetScaleWidth(refShape);
             float refScaleHeight = Graphics.GetScaleHeight(refShape);
 
+            float cropTop = Math.Max(shapeRange[1].PictureFormat.CropTop, Epsilon);
```
initialshl: Maybe could use the refShape var instead of shapeRange[1]?

jamos-tay: Good idea, done

PR #1186, opened 2017-03-12
Crop to slide: support even cropping of non-picture objects #1179

### LINKED ISSUE #1179: Crop to slide: support even cropping of non-picture objects
I suppose we can convert the object to a picture first and then do the cropping?

jamos-tay:
Fixes #1179.

Cropping shapes now automatically casts them to picture.


jamos-tay:
Closing and making new one

PR #1189, opened 2017-03-13
Redirect in README could be improved #1188

### LINKED ISSUE #1188: Redirect in README could be improved
This is with reference to this link in readme.md under the Setup section:

image

When clicked, it redirects to [this link](https://github.com/PowerPointLabs/PowerPointLabs-Website/blob/master/vsto-redirect.html), which is a html file within github where the user is asked to extract a URL from:

image

This seems pretty unelegant. A better idea would be to link to the same html file hosted on the server, which redirects them to the correct URL directly:

http://powerpointlabs.info/vsto-redirect.html

jamos-tay:
Fixes #1188.

Changed link in readme from [this](https://github.com/PowerPointLabs/PowerPointLabs-Website/blob/master/vsto-redirect.html) to [this](http://powerpointlabs.info/vsto-redirect.html).

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ Interested to contribute? Please take a moment to review the [guidelines for con
 1. Turn off *Office Version Upgrading*: Open the solution `PowerPointLabs\PowerPointLabs.sln` >> open Tools (menu) >> Options >> Office Tools >> Project Migration >> uncheck ‘Always upgrade to installed version of Office’.
 2. Set up *External Office Program*: Open ‘PowerPointLabs’ Properties >> Debug >> select ‘Start external program’ and choose `POWERPNT.exe` in the Office folder. Ensure that both **Debug** and **Release** configurations have set up this.
 3. Run the solution by pressing F5 and then PowerPointLabs tab will appear in the PowerPoint ribbon. If you have installed PowerPointLabs add-in, you may have to uninstall it first and rebuild the solution.
-4. If failed to build PowerPointLabs solution, try to install `VSTO for PowerPoint` from the link inside [this file](https://github.com/PowerPointLabs/PowerPointLabs-Website/blob/master/vsto-redirect.html).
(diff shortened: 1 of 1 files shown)


Kenneth-LJS:
LGTM

acjh:
1. The sentence no longer fits the new link.
1. Is there a purpose for clicking on a link to a file to be immediately redirected? Why not link directly?

jamos-tay:
@acjh 
1. Thanks for pointing this out, I have reworded the sentence to make it more applicable.
2. I think this is safer since the vsto-redirect may be accessible from elsewhere, and also we might want to change the link to download in future if Microsoft's URL changes. In this case, we only need to change the link in the vsto-redirect file instead of both the vsto-redirect and the readme.

Skaty:
I think linking to our website from this repository is not really a good choice. It would be better if this repository does not depend on the website.

Also, @jamos-tay, your second point doesn't really hold, as we can update the README when it is needed.

jamos-tay:
@Skaty Fair enough, you could update both, but I think this violates DRY. It makes sense that a string const such as a URL should only be stored in one location, and have everything reference it rather than having it stored in multiple places. Sure, you and I will know to update both, but future programmers who need to change the link might just change it in one place and not realize that another link needs to be changed.

Skaty:
@jamos-tay Makes sense, in a way.

kai33:
the history of vsto-redirect:
few years ago, MS has a convention to update VSTO for PowerPoint quarterly or at least a few times yearly. every time when it updates, the previous link won't work anymore (404). that in fact results in: we'll need to modify the link in powerpoint installer and re-publish it; the installer downloaded previously by users will have a broken link, which may be misleading to them. so i added this vsto-redirect thing, and let the CI guard it. then no more new release required, and its link should never be broken

jamos-tay:
@kai33 Ah, so is there anything else I should do for this PR?

PR #1191, opened 2017-03-14
Crop out padding: over cropping #1182 

### LINKED ISSUE #1182: Crop out padding: over cropping
Seems to be cropping more than necessary
image
Probably because it is under-cropping on the other side:
image
I was using the PowerPointLabs logo that comes with the tutorial file by the way.

initialshl:
Fixes #1182 
Problem caused by width/height index offset as array.


initialshl:
Will open another PR for this since the alpha-release build changed

PR #1192, opened 2017-03-14
Crop out padding: over cropping #1182 

### LINKED ISSUE #1182: Crop out padding: over cropping

initialshl:
Fixes #1182 
Problem caused by width/height index offset as array


initialshl:
Hmm i'm not sure about the shift, but i suspect that it will happen using PPT's Crop feature too,
since this method only changes the crop dimensions using the PPT API (i.e. no positioning/scaling/etc.).

PR #1201, opened 2017-03-15
PowerPointPresentation: Presentation name not defined #1198

### LINKED ISSUE #1198: PowerPointPresentation: Slide name not defined
When a new PowerPointPresentation object is created from an existing opened presentation, the name of the presentation is not defined in the constructor.

As such, any functions/properties that rely on the name, for instance Opened, will not work.

Skaty:
Fixes #1198

As a Presentation name does not necessarily have an extension (for instance, a new presentation), this commit allows setting of Name without the extension.

### DIFF
--- a/PowerPointLabs/PowerPointLabs/Models/PowerPointPresentation.cs
+++ b/PowerPointLabs/PowerPointLabs/Models/PowerPointPresentation.cs
@@ -14,6 +14,7 @@ namespace PowerPointLabs.Models
 #pragma warning disable 0618
         #region Properties
         private string _name;
+        private const string extensionRegex = "\\.pptx?$";
 
         public static Application Application { get; set; }
 
@@ -30,7 +31,14 @@ namespace PowerPointLabs.Models
         {
             get
             {
-                return Path + @"\" + Name;
+                if (Regex.IsMatch(Name, extensionRegex, RegexOptions.IgnoreCase))
+                {
+                    return Path + @"\" + Name;
+                }
+                else
+                {
+                    return Path + @"\" + Name + ".pptx";
+                }
             }
         }
 
@@ -79,8 +87,16 @@ namespace PowerPointLabs.Models
             get { return _name; }
             set
             {
-                NameNoExtension = value;
-                _name = value + ".pptx";
+                if (Regex.IsMatch(value, extensionRegex, RegexOptions.IgnoreCase))
+                {
+                    NameNoExtension = Regex.Replace(value, extensionRegex, "", RegexOptions.IgnoreCase);
+                }
+                else
+                {
+                    NameNoExtension = value;
+                }
+
+                _name = value;
             }
         }
 
@@ -222,12 +238,13 @@ namespace PowerPointLabs.Models
         public PowerPointPresentation(string path, string name)
         {
             Path = path;
-            Name = name;
+            Name = name + ".pptx";
         }
 
         public PowerPointPresentation(Presentation presentation)
         {
             Presentation = presentation;
+            Name = presentation.Name;
         }
         # endregion
 
--- a/PowerPointLabs/PowerPointLabs/PictureSlidesLab/Service/StylesDesigner.cs
+++ b/PowerPointLabs/PowerPointLabs/PictureSlidesLab/Service/StylesDesigner.cs
@@ -41,7 +41,7 @@ namespace PowerPointLabs.PictureSlidesLab.Service
         public StylesDesigner(Application app = null)
         {
             Path = TempPath.TempFolder;
-            Name = "PictureSlidesLabPreview" + Guid.NewGuid().ToString().Substring(0, 7);
+            Name = "PictureSlidesLabPreview" + Guid.NewGuid().ToString().Substring(0, 7) + ".pptx";
             Option = new StyleOption();
             Application = app;
             OpenInBackground();


PowerPointLabs/PowerPointLabs/Models/PowerPointPresentation.cs
```diff
@@ -72,8 +79,16 @@ public string Name
             get { return _name; }
             set
             {
-                NameNoExtension = value;
-                _name = value + ".pptx";
+                if (Regex.IsMatch(value, "\\.pptx?$"))
+                {
+                    NameNoExtension = Regex.Replace(value, "\\.pptx?$", "");
```
Kenneth-LJS: Store the regex as a constant

PowerPointLabs/PowerPointLabs/Models/PowerPointPresentation.cs
```diff
@@ -30,7 +30,14 @@ public string FullName
...
             {
-                return Path + @"\" + Name;
+                if (Regex.IsMatch(Name, "\\.pptx?$"))
+                {
+                    return Path + @"\" + Name;
+                }
+                else
+                {
+                    return Path + @"\" + Name + ".pptx";
+                }
```
Kenneth-LJS: Can you just use 

> return Path + @"\\" + NameNoExtension + ".pptx";

 to reduce the need for the regex check?

Skaty: I don't think so, as some files may have a ppt extension.

PowerPointLabs/PowerPointLabs/PictureSlidesLab/Service/StylesDesigner.cs
```diff
@@ -41,7 +41,7 @@ public sealed class StylesDesigner : PowerPointPresentation, IStylesDesigner
         public StylesDesigner(Application app = null)
         {
             Path = TempPath.TempFolder;
-            Name = "PictureSlidesLabPreview" + Guid.NewGuid().ToString().Substring(0, 7);
```
kai33: I intentionally made it w/o extension, in case users see it as a ppt file and open it. any error u encountered when w/o extension?

Skaty: "see it as a ppt file" meaning?

damithc: Not sure if this is the right answer. In some cases we store internal data as pptx files but we omit `.pptx` in the file name because we don't want users to open those files in PowerPoint and edit them.

Skaty: @damithc

I see. The change was done as the old setter for `Name` appended the `.pptx` extension whenever the Name field is set.

[Relevant line](https://github.com/PowerPointLabs/PowerPointLabs/pull/1201/files#diff-87b0ea049b75914d91a33f9493b97058L76)

PowerPointLabs/PowerPointLabs/Models/PowerPointPresentation.cs
```diff
@@ -14,6 +14,7 @@ public class PowerPointPresentation
 #pragma warning disable 0618
         #region Properties
         private string _name;
+        private const string extensionRegex = "\\.pptx?$";
```
kai33: does it work with `PPTX` in upper case?

PowerPointLabs/PowerPointLabs/Models/PowerPointPresentation.cs
```diff
@@ -30,7 +31,14 @@ public string FullName
         {
             get
             {
-                return Path + @"\" + Name;
+                if (Regex.IsMatch(Name, extensionRegex))
```
kai33: here u assume that fullname must have .pptx extension, which may not hold?

kai33: ok the existing one assumes that name has extension

kai33:
found this. does shapes lab work?
https://github.com/PowerPointLabs/PowerPointLabs/blob/dev-release/PowerPointLabs/PowerPointLabs/CustomShapePane.cs#L959

the existing assumption is that,
for property `name`, dev need to give a name w/o extension, and its logic will add extension for you.

bsoc u have modified its behavior (`name` may be with or w/o extension), please also verify the existing use cases are correct

Skaty:
@kai33

I have verified it. `PowerPointShapeGalleryPresentation` constructor does not use the name field, for some reason.

PowerPointLabs/PowerPointLabs/Models/PowerPointPresentation.cs
Skaty: @kai33 

Yes it works, I just had to add `RegexOptions.IgnoreCase`.

Skaty:
@kai33 changes made.

kai33:
I'm thinking that 
for `public PowerPointPresentation(string path, string name)` and property `Name` access, we can keep the old way. so that no need to add/change `.pptx` here and there.
and just for `public PowerPointPresentation(Presentation presentation)`, you can set the name to suit your feature, e.g. set directly to `_name`

Skaty:
That wouldn't really be a good way, as setting `Name` sets the `NameNoExtension` variable also.

kai33:
also set it?

Skaty:
Hmm, I think that's more complicated than this solution, as I only changed `Name' and not both of them.

kai33:
just a setter for them? why complicated

Skaty:
I see. Actually, another argument is that filename extensions should be explicitly defined, so that we don't assume that a file without an extension defaults to `.pptx' (you can open a file without an extension in PowerPoint).

kai33:
If the name must have an extension, the setter should guard this behavior. But now, it can be with or w/o extension, and its interface doesn't tell what to expect explicitly: it assumes when to use it with extension and when not to. To me it's confusing

Skaty:
@kai33

I only did the conditional because of the `NameNoExtension` requirement. The Presentation object in VSTO follows the name that is displayed on the title bar (if it's New Presentation, it will be New Presentation).

Thus, `Name` property can expect either one, since this `Name` property is a reflection of the Name property in VSTO.

kai33:
Can u help verify this:
When it's a new ppt file (unsaved) it doesnt carry extension. When it's saved (manually or API) then it has extension

kai33:
And mapping to VSTO doesn't mean it's a right behavior. U'r changing many usage to include extension that means the assumption is there, but the setter doesnt tell that, which is confusing

kai33:
And does Opened still work properly when:
1 pass a new presentation instance, its name doesnt have extension
2 that presentation is saved and its name now has extension

Skaty:
What do you mean by the "setter doesnt tell that"?

The Name in VSTO follows the name as what is shown on the titlebar.

In those cases, it doesn't work because we don't hook onto the Save event (if there is even an event like so). There's no good way to know if something is opened, as there is no unique ID. It's due to the way the original implementer of Opened did it.

The only reason why I did this is to allow me to use the AddSlide method.

kai33:
It has an assumption that, for saved file it has extension and for unsaved one it's without extension, but both use cases share the interface (constructor or name's setter) which doesn't explicitly tell and guard when with/without extension.

Why need the event?
For Opened, we have original presentation instance, why not use its name, instead of using its copy

Skaty:
>It has an assumption that, for saved file it has extension and for unsaved one it's without extension, but both use cases share the interface (constructor or name's setter) which doesn't explicitly tell and guard when with/without extension.

I still don't get what you're trying to say. I'm not assuming anything, instead, this implementation makes it clear that filename that is supplied MUST include the extension, if there's ANY. It is not acceptable that the current form assumes a PPTX extension for ALL presentations.

Let me give you an example. Try saving a pptx file, close PowerPoint, change/remove extension and then finally, re-opening the file in PowerPoint (by selecting "All Files" in the filter). The result, a file that has no extension that is opened in PowerPoint.

Example:
image

See the ".txt" extension in that Presentation? It WILL break the logic in the old Name method, as the VSTO object will supply Name as `DoNotAssumeThatFileIsAlwaysPPTX.txt` and that setter will define Name as `DoNotAssumeThatFileIsAlwaysPPTX.txt.pptx`, which is WRONG. The logic for `Opened` compares the Presentation object's `Name` with the VSTO `Name`. Although I do concede that my check for NameNoExtension is wrong, as it should consider this possibility too (it assumes that file extension is ppt[x] and not others).

> For Opened, we have original presentation instance, why not use its name, instead of using its copy

Well, if it's closed, you get a COMException when accessing the `Name` object. This object does not sync with the VSTO object, so the way `Opened` is implemented by the original developer is that it is safe from exceptions raised by the COM object. Though I do have to agree that it is a BAD implementation and a better one would use `FullName` instead, since PowerPoint does not allow user to open the same presentation twice (except through `View -> New Window`, which does not create a separate Presentation object in `Applications.Presentations`).

kai33:
1. I get the example. Here we have 3 use cases:
a. constructor `PowerPointPresentation(string path, string name)`
b. constructor `PowerPointPresentation(Presentation presentation)`
c. setter of `Name`

for b, I agree it should take presentation's name with or w/o extension, not limited to .pptx
for a and c, the intention is to open or create a pptx file that we own, for this case we can maintain the extension. but a and c don't guard and enforce this behavior, e.g. throw exception when w/o extension.

2. just catch it and return false (not opened)?

Skaty:
1. The setter of `Name` should just set the `Name` as it is. But if you want it as such, then I'm okay with assuming `.pptx` like previously.

2. Hmm, a generic try catch COMException would be fine? Don't have to check if the COMException is due to other things?

kai33:
1 im fine with it
2 MS uses COMException for many kinds of errors. If a COM object throws this error, likely means the object is corrupted or released. So should be ok, or u can check for the specific error code inside it if any

Skaty:
1. So we should revert to original? Because the change is not really huge (just move appending the "pptx" to constructor (a)).
2. That would be solved in a different PR, because this PR is explicitly about the Name field not being set if constructor (b) is used.

kai33:
go ahead

Skaty:
I've made changes.

PR #1204, opened 2017-03-16
Crop to slide: support even cropping of non-picture objects #1179

### LINKED ISSUE #1179: Crop to slide: support even cropping of non-picture objects

jamos-tay:
Fixes #1179 

Cropping shapes now automatically casts them to picture.

Copy of #1186, rebased on alpha-release to clean branch history.


damithc:
> It is possible for the cropped shape to remain as a shape if merge shapes is used.

IIRC, that api is not available in ppt 2010

jamos-tay:
Ah, also I recall one of the requirements of CropToSlide was to set the picture's crop window rather than actually cropping the image. This allows the crop to be undone in future should the user want to.

AngShiYa:
@damithc Yes, the api is not available but there is a workaround to execute merge shapes using the command bar. E.g. `this.ExecuteOfficeCommand("ShapesIntersect");`

@jamos-tay I see.. didn't know that requirement exist.

damithc:
> Yes, the api is not available but there is a workaround to execute merge shapes using the command bar. E.g. this.ExecuteOfficeCommand("ShapesIntersect");

Nice. Are we using that already? I remember some features were done in a sub-optimal way simply because  of the 2010 API limitation. e.g. spotlight

AngShiYa:
If I'm not wrong, we are not using merge shapes for any feature.

jamos-tay:
Oh yeah, one thing to note is also that merge shapes uses the intersect of every selected shape, i.e. (A n B n C ...) so it wouldn't work for cases like this:

image

Since the intersection is an empty set.

damithc:
> one thing to note is also that merge shapes uses the intersect of every selected shape, i.e. (A n B n C ...) so it wouldn't work for cases like this

On the UI, there are different options:

image

jamos-tay:
> On the UI, there are different options:

I'm not sure how any of those help...

My point is that merge shapes wouldn't work for Crop to Slide when the user is cropping multiple shapes, for example a case like this:

image

We can't just create a shape of the slide and use Merge Shape because it uses the intersect of every single selected shape (A n B n C n ... n Slide), when actually what we want is the intersect of each shape with the slide ((A n Slide) u (B n Slide) u (C n Slide) u ...). We can get around this by creating a slide shape for each selected shape, but that seems rather inefficient.

PowerPointLabs/PowerPointLabs/CropLab/CropToSlide.cs
```diff
@@ -81,5 +81,12 @@ private static RectangleF GetCropArea(PowerPoint.Shape shape, float slideWidth,
 
             return new RectangleF(cropLeft, cropTop, cropWidth, cropHeight);
         }
+
+        private static bool IsShape(PowerPoint.Shape shape)
```
initialshl: Can move this to a util class?

PowerPointLabs/PowerPointLabs/CropLab/CropLabErrorHandler.cs
```diff
@@ -55,6 +57,7 @@ public void ProcessErrorCode(int errorCode, string featureName, string validSele
                     }
                     ShowErrorMessage(errorCode, featureName, validSelectionMinCount.ToString(), validSelectionType);
                     break;
+                case ErrorCodeSelectionMustBeShapeOrPicture:
```
initialshl: This switch case seems to be duplicated at line 60 and 112

initialshl: This switch case seems to be duplicated at line 60 and 112

PowerPointLabs/PowerPointLabs/CropLab/CropToSlide.cs
jamos-tay: Sure, I moved it to graphics

damithc:
> We can't just create a shape of the slide and use Merge Shape because it uses the intersect of every single selected shape (A n B n C n ... n Slide), when actually what we want is the intersect of each shape with the slide ((A n Slide) u (B n Slide) u (C n Slide) u ...). We can get around this by creating a slide shape for each selected shape, but that seems rather inefficient.

Yes, no one option can give you the result you want. How about these 2 steps? 1. union all shapes 2. intersect with a shape that matches the slide

PR #1227, opened 2017-03-21
Support naming copied attributes #1224

### LINKED ISSUE #1224: Support naming copied attributes
When copying some attributes from an object, allow the user to specify a name for it (add a text box to the 'copy' dialog?). If the user does not specify a name, we can create a name based on the name of the object and the attributes copied.

Kenneth-LJS:
Fixes #1224 

A textbox has been added in the format dialog for users to specify the display name.

image

image

When copying the shape, the default name will be the shape name. Users can change the display name through the textbox on the dialog.

When double-clicking the format to open the options dialog, the textbox will show the current display name, and allow users to change it if required.

### DIFF
--- a/PowerPointLabs/PowerPointLabs/SyncLab/View/SyncFormatPaneItem.cs
+++ b/PowerPointLabs/PowerPointLabs/SyncLab/View/SyncFormatPaneItem.cs
@@ -77,12 +77,14 @@ namespace PowerPointLabs.SyncLab.View
         private void OnMouseDoubleClick(object sender, MouseButtonEventArgs e)
         {
             SyncFormatDialog dialog = new SyncFormatDialog(shape, this.formats);
+            dialog.ObjectName = this.Text;
             bool? result = dialog.ShowDialog();
             if (!result.HasValue || !(bool)result)
             {
                 return;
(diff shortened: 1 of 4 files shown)


PowerPointLabs/PowerPointLabs/SyncLab/View/SyncFormatPaneItem.cs
```diff
@@ -77,12 +77,14 @@ private void RadioBoxChecked(object sender, RoutedEventArgs e)
...
         {
             SyncFormatDialog dialog = new SyncFormatDialog(shape, this.formats);
+            dialog.ObjectName = this.Text;
             bool? result = dialog.ShowDialog();
             if (!result.HasValue || !(bool)result)
             {
                 return;
             }
             this.formats = dialog.Formats;
+            this.Text = dialog.ObjectName;
```
jamos-tay: Suggestion: How about using the default name if the user leaves the dialog field blank?

Kenneth-LJS: Can you make a new issue?

PR #1229, opened 2017-03-21
Improve text display in SyncLab

### LINKED ISSUE #1228: Standardize SyncLab dialog font size
The size of the text scales to fill the box, so that overly-large text doesn't fill the whole box. However the disadvantage is that the display text now have inconsistent sizes.

image

I suggest using a constant font size first, but making it smaller if it cannot fit in the box.

Kenneth-LJS:
Fixes #1228 

image

Font size is now standardized (See Width/Height/X/Y). For line weight I displayed it with a placeholder to show that the font shrinks to fit longer text.

### DIFF
--- a/PowerPointLabs/PowerPointLabs/SyncLab/SyncFormatUtil.cs
+++ b/PowerPointLabs/PowerPointLabs/SyncLab/SyncFormatUtil.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Drawing;
+using System.Drawing.Text;
 
 using Microsoft.Office.Interop.PowerPoint;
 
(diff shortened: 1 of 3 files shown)


Kenneth-LJS:
This is actually sharper than before. I turned on anti-aliasing when drawing the string. I can look into making it clearer in future.

PowerPointLabs/PowerPointLabs/SyncLab/SyncFormatUtil.cs
```diff
@@ -21,22 +22,23 @@ public static Shapes GetTemplateShapes()
...
-            double scale = Math.Min(size.Width / textSize.Width, size.Height / textSize.Height);
-            double newWidth = textSize.Width * scale;
-            double newHeight = textSize.Height * scale;
-            double newX = (size.Width - newWidth) / 2;
-            double newY = (size.Height - newHeight) / 2;
-            g.DrawImage(textImage, Convert.ToSingle(newX), Convert.ToSingle(newY),
-                Convert.ToSingle(newWidth), Convert.ToSingle(newHeight));
+            if (textSize.Width > imageSize.Width || textSize.Height > imageSize.Height)
+            {
+                double scale = Math.Min(imageSize.Width / textSize.Width, imageSize.Height / textSize.Height);
```
jamos-tay: Is there potential for a div by 0 error if the text is nothing?

PowerPointLabs/PowerPointLabs/SyncLab/SyncFormatUtil.cs
```diff
@@ -21,22 +22,23 @@ public static Shapes GetTemplateShapes()
...
+                System.Drawing.Font newFont = new System.Drawing.Font(font.FontFamily, Convert.ToSingle(font.Size * scale),
+                                                            font.Style, font.Unit, font.GdiCharSet, font.GdiVerticalFont);
+                return GetTextDisplay(text, newFont, imageSize);
```
jamos-tay: Is it better to just assign font and textSize rather than rerunning the function? It seems like only the font and textSize changes.

PR #1231, opened 2017-03-21
Timer Lab: Provide images for rough indication of timer size #1230

### LINKED ISSUE #1230: Timer Lab: Provide images for rough indication of timer size

AngShiYa:
Fixes #1230 

Before:
before

After:
taskpane


damithc:
Um, this feature is not useful unless you show the real size in the preview.

AngShiYa:
@damithc i don't think it is possible to provide a preview.

I did it this way because this will give the user a better idea of the size of the timer with respect to the slide dimension as compared to just a text box.

And in consideration of customizing the timer, this UI can also facilitate real time changing of size for the timer if it is already created. Although this is only possible if we restrict each slide to have only one timer. Which I think is a reasonable trade off considering that there are not many cases that a user would want to have multiple timer.

damithc:
OK, I'll dogfood this version and give you feedback after that.

IanTeo:
I just tested on 2010. It works too

PR #1233, opened 2017-03-22
Crop Lab: Convert some functional tests to unit tests

initialshl:
Converted CropOutPadding and CropToAspectRatio FTs to UTs.


PowerPointLabs/PowerPointLabs/Utils/Graphics.cs
```diff
@@ -53,9 +53,19 @@ public static Shape CorruptionCorrection(Shape shape, PowerPointSlide ownerSlide
...
+            int slideWidth = 0;
+            int slideHeight = 0;
+            try
+            {
+                slideWidth = (int)PowerPointPresentation.Current.SlideWidth;
+                slideHeight = (int)PowerPointPresentation.Current.SlideHeight;
+            }
+            catch (NullReferenceException)
+            {
+                // Presentation may be null during Unit Tests
```
initialshl: The `NullReferenceException` was thrown from the getter body of `PowerPointPresentation.Current` when it tries to reference some other objects, so I think catching the exception is the better design for this ExportShape function.
(My comment there about Presentation being null isn't very accurate, i should change it.)

jamos-tay: Ah okay then

initialshl:
> Tests passing on 2013 and 2010. Is there any reason why the other FTs cannot be converted to UTs?

I'm not as familiar about the FTs of other crop functions, so i'd better leave it to @jamos-tay

Issue #1220, opened 2017-03-19
Crop to same dimensions: give options for anchor points


damithc:
When cropping based on another model object, the result can vary on where we consider the 'anchor point'.
In this example, the crop window in anchored to the center.
![image](https://cloud.githubusercontent.com/assets/1673303/24078844/f0c84e9a-0cb3-11e7-8d06-0b7d2fe9812c.png)
Similarly, it can be anchored to 8 other points.
![image](https://cloud.githubusercontent.com/assets/1673303/24078847/0fa2dfa6-0cb4-11e7-923d-345bbb6d11a1.png)
Perhaps we can provide 9 sub-menu options for each of these anchor points?

initialshl:
Closing issue: fix has been merged into dev-release

PR #1244, opened 2017-03-24
Crop Lab: Implement Settings for anchor points

initialshl:
Related to #1220 

Created Crop Lab Settings dialog.
Anchor point is implemented for Crop To Aspect Ratio.

croplab-settings

### DIFF
--- a/PowerPointLabs/PowerPointLabs/CropLab/CropLabSettings.cs
+++ b/PowerPointLabs/PowerPointLabs/CropLab/CropLabSettings.cs
@@ -1,0 +1,26 @@
+﻿namespace PowerPointLabs.CropLab
+{
+    public enum AnchorPosition
+    {
+        TopLeft,
+        Top,
+        TopRight,
+        MiddleLeft,
+        Middle,
+        MiddleRight,
+        BottomLeft,
+        Bottom,
+        BottomRight
+    }
+
+    internal static class CropLabSettings
+    {
+        private static AnchorPosition anchorPosition = AnchorPosition.Middle;
+
+        public static AnchorPosition AnchorPosition
+        {
+            get { return anchorPosition; }
+            set { anchorPosition = value; }
+        }
+    }
+}
--- a/PowerPointLabs/PowerPointLabs/CropLab/CropLabSettings.cs
+++ b/PowerPointLabs/PowerPointLabs/CropLab/CropLabSettings.cs
@@ -15,12 +15,6 @@
 
     internal static class CropLabSettings
     {
-        private static AnchorPosition anchorPosition = AnchorPosition.Middle;
-
-        public static AnchorPosition AnchorPosition
-        {
-            get { return anchorPosition; }
-            set { anchorPosition = value; }
-        }
+        public static AnchorPosition AnchorPosition = AnchorPosition.Middle;
     }
 }
--- a/PowerPointLabs/PowerPointLabs/CropLab/CropLabSettingsDialog.xaml
+++ b/PowerPointLabs/PowerPointLabs/CropLab/CropLabSettingsDialog.xaml
@@ -1,0 +1,59 @@
+﻿<controls:MetroWindow x:Class="PowerPointLabs.CropLab.CropLabSettingsDialog"
+                      xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+                      xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+                      xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006" 
+                      xmlns:d="http://schemas.microsoft.com/expression/blend/2008" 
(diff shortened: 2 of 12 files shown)


initialshl:
> An option is to put crop lab as a group on its own in the ribbon so that you have a bottom right arrow to crop lab

I'm not sure if users use crop lab often enough to warrant it to be expanded on the ribbon, this may require more dogfooding and feedback.
Personally, I think that we should make crop lab and its settings open in a side pane (not sure if we have enough time to), so that the user can crop all he wants using one mouse click each time.

damithc:
> I'm not sure if users use crop lab often enough to warrant it to be expanded on the ribbon, this may require more dogfooding and feedback.

Most likely we'll be collapsing each lab to just one button on the ribbon, to save ribbon space. Users can use the 'add to quick access bar' feature to add frequently used features to the quick access bar.

> Personally, I think that we should make crop lab and its settings open in a side pane (not sure if we have enough time to), so that the user can crop all he wants using one mouse click each time.

If we do that, it should be an option rather than the only way to use crop lab.  Otherwise those who need to crop something just once will have to do too many things i.e. open panel, click, close panel.

PowerPointLabs/PowerPointLabs/CropLab/CropLabSettings.cs
```diff
@@ -0,0 +1,26 @@
...
+        Middle,
+        MiddleRight,
+        BottomLeft,
+        Bottom,
+        BottomRight
+    }
+
+    internal static class CropLabSettings
+    {
+        private static AnchorPosition anchorPosition = AnchorPosition.Middle;
```
jamos-tay: Why are there 2 vars anchorPostition and AnchorPosition? Try `public static AnchorPosition AnchorPosition { get; set; } = AnchorPosition.Middle;`

PowerPointLabs/PowerPointLabs/CropLab/CropLabSettingsDialog.xaml.cs
```diff
@@ -0,0 +1,35 @@
...
+    {
+        private AnchorPosition selectedAnchor;
+
+        public CropLabSettingsDialog()
+        {
+            selectedAnchor = CropLabSettings.AnchorPosition;
+            InitializeComponent();
+        }
+
+        public AnchorPosition SelectedAnchor
```
jamos-tay: Same as above

PowerPointLabs/PowerPointLabs/CropLab/CropLabSettings.cs
initialshl: Thanks, fixed

PowerPointLabs/PowerPointLabs/CropLab/CropLabSettingsDialog.xaml
```diff
@@ -0,0 +1,59 @@
...
+            <RadioButton x:Name="bottomAnchorRadioButton" Content="" HorizontalAlignment="Left" Margin="10,6,0,0" VerticalAlignment="Top"
+                         IsChecked="{Binding Path=SelectedAnchor, Converter={StaticResource EnumToBooleanConverter}, ConverterParameter={x:Static local:AnchorPosition.Bottom}}" />
+            <RadioButton x:Name="bottomRightAnchorRadioButton" Content="" HorizontalAlignment="Left" Margin="10,6,0,0" VerticalAlignment="Top"
+                         IsChecked="{Binding Path=SelectedAnchor, Converter={StaticResource EnumToBooleanConverter}, ConverterParameter={x:Static local:AnchorPosition.BottomRight}}" />
+        </WrapPanel>
+        <Button x:Name="okButton" Content="OK" HorizontalAlignment="Right" Margin="0,0,100,10" VerticalAlignment="Bottom" Width="80" IsDefault="True" Click="OkButton_Click"/>
+        <Button x:Name="cancelButton" Content="Cancel" HorizontalAlignment="Right" Margin="0,0,10,10" VerticalAlignment="Bottom" Width="80" IsCancel="True" Click="CancelButton_Click"/>
```
Kenneth-LJS: Apply the property
IsDefault="True" on okButton, and
IsCancel="True" on cancelButton

PowerPointLabs/PowerPointLabs/CropLab/CropLabSettingsDialog.xaml.cs
```diff
@@ -0,0 +1,29 @@
...
+        private void OkButton_Click(object sender, RoutedEventArgs e)
+        {
+            CropLabSettings.AnchorPosition = SelectedAnchor;
+            Close();
+        }
+
+        private void CancelButton_Click(object sender, RoutedEventArgs e)
+        {
+            Close();
+        }
```
Kenneth-LJS: CancelButton_Click can be removed after setting IsCancel="True" on the cancel button.

PR #1248, opened 2017-03-25
Timer duration not syncing automatically #1241

### LINKED ISSUE #1241: Timer duration not syncing automatically
Currently, the width and height are able to change when the user uses the slider. However, the animation time will not change. We need to provide a way for users to change the duration of the timer

AngShiYa:
Fixes #1241 

Markers and slider on the timer will now change to reflect changes in the duration.


AngShiYa:
@Skaty Oh good catch. This happens because of the control. I can fix it by changing the UI to take in both minutes and seconds as integers instead. I will create another issue for this.

PR #1250, opened 2017-03-28
Sync Lab: Default to shape name if name text box is empty #1239

### LINKED ISSUE #1239: Sync Lab: Default to shape name if name text box is empty
Follow up from #1227 

When the user adds a shape to sync labs but leaves the name empty, it creates a shape with a blank name.

image

image

This is a little unsightly, I suggest setting the name to the default shape name if the name text is empty.

Kenneth-LJS:
Fixes #1239

Behaviour:
1) Text box will show name of shape when copy button is pressed
2) If text box is left blank, the shape's name will be used as the list name
3) On double-clicking to modify the format, leaving a blank text box will not result in a name change

### DIFF
--- a/PowerPointLabs/PowerPointLabs/SyncLab/View/SyncFormatDialog.xaml.cs
+++ b/PowerPointLabs/PowerPointLabs/SyncLab/View/SyncFormatDialog.xaml.cs
@@ -16,15 +16,26 @@ namespace PowerPointLabs.SyncLab.View
     public partial class SyncFormatDialog : Window
     {
 
+        string originalName;
         FormatTreeNode[] formats = null;
 
-        public SyncFormatDialog(Shape shape) : this(shape, SyncFormatConstants.FormatCategories)
+        public SyncFormatDialog(Shape shape) : this(shape, shape.Name, SyncFormatConstants.FormatCategories)
         {
         }
 
-        public SyncFormatDialog(Shape shape, FormatTreeNode[] formats)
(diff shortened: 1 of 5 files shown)


PowerPointLabs/PowerPointLabs/SyncLab/View/SyncFormatDialog.xaml.cs
```diff
@@ -16,15 +16,26 @@ namespace PowerPointLabs.SyncLab.View
...
+        public SyncFormatDialog(Shape shape) : this(shape, shape.Name, SyncFormatConstants.FormatCategories)
         {
         }
 
-        public SyncFormatDialog(Shape shape, FormatTreeNode[] formats)
+        public SyncFormatDialog(Shape shape, string formatName, FormatTreeNode[] formats)
         {
             InitializeComponent();
+            formatName = formatName.Trim();
+            if (SyncFormatUtil.IsValidFormatName(formatName))
```
jamos-tay: This if is used in two places, why not handle it in SyncUtil?
For example changing `bool IsValidFormatName(string name)` to `string GetValidFormatName(string name)`

Kenneth-LJS: If the name is not a valid format, I will need to revert it to
1) The original text before it was changed
2) The original shape's text

This will differ based on where it is called from.

jamos-tay: Fair enough

PR #1253, opened 2017-03-28
Sync Lab UI Revamp

Kenneth-LJS:
The UI looks like this now:

image

Changes made:
1) The paste button has been removed
2) Pasting can be done with 1 click by pressing on the paste button (represented by the "P" button placeholder)
3) The ability to change the copied formats is now visually obvious with a new edit button on each item (represented by the "E" button placeholder)


PowerPointLabs/PowerPointLabs/SyncLab/View/SyncFormatPaneItem.xaml.cs
```diff
@@ -0,0 +1,126 @@
...
+namespace PowerPointLabs.SyncLab.View
+{
+    /// <summary>
+    /// Interaction logic for SyncFormatPaneItem.xaml
+    /// </summary>
+    public partial class SyncFormatPaneItem : UserControl
+    {
+
+        
+        Bitmap image;
```
jamos-tay: Put access modifiers?

Kenneth-LJS:
@IanTeo We'll need someone to redesign the GUI. But I want to finalize the layout before we start putting work into the current layout.

damithc:
The current one can go ahead. For future reference, I wonder if putting the P/E buttons on the other end makes more ergonomic sense (lowers the travel distance for the mouse pointer). Not sure.

IanTeo:
Ok. Understood. Just want to make a comment that the UI requires a delete button too. If the user wants to remove that sync

Issue #1207, opened 2017-03-16
Sync Lab: Improve Sync Fill


IanTeo:
- [ ] Add support for Gradient Fill
- [ ] Add support for Pattern Fill
- [ ] Add support for Picture Fill
- [ ] Add support for Texture Fill
- [ ] Add support for Slide Background Fill
- [ ] Add support for No Fill

IanTeo:
"Gradient Fill" and "Picture/Texture Fill" are quite tough and will require some more research
I am unable to find a way to change the shape fill to "No Fill"

YuPeiHenry:
Tested syncing "no fill" for PPT 2010 and PPT 2016, seems to work if either fill>color or fill>transparency is selected. Checking the "no fill" box.

PR #1254, opened 2017-03-28
Sync Lab: Add Sync Gradient Fill

IanTeo:
Fixes part of #1207 

CanCopy now makes a duplicate of the shape because gradient needs to be synced part by part.

Currently, most well behaved gradients can be synced. There are some gradients that still have weird behaviors though.

* Preset gradients cannot be synced
* Radial Type gradients somehow will be synced as rectangular type
* Gradients which do not have an angle multiple of 45 cannot be synced

Example of some of the gradients that can be synced
image

Despite the shortcomings, I think that we should release this version of gradient sync first as we attempt to solve the remaining issues. That way, we can dogfood it to find any other issues.


damithc:
Yes, we can release with a disclaimer that only simple gradients can be synced at the moment.

IanTeo:
Yes, we can. If there are any other features that can sync gradient in a similar way

PR #1255, opened 2017-03-29
User can still create multiple timers #1242

### LINKED ISSUE #1242: User can still create multiple timers
If the user deletes the main timer body, he can create multiple timers. We should either:

* Disallow creating as long as any timer object exists
* Edit the existing timer

AngShiYa:
Fixes #1242 

If any components of the timer still exist, clicking the create button will trigger the error message that only one timer is allowed. Any components of the timer that is missing will also be generated based on the values on the pane.


AngShiYa:
@Kenneth-LJS I've updated it to follow the main block if it exist and re-position to the center if the main block does not exist.

PR #1265, opened 2017-04-01
Crop to same dimensions: give options for anchor points #1220

### LINKED ISSUE #1220: Crop to same dimensions: give options for anchor points

jamos-tay:
Partially fixes #1220.

Users can use Crop Lab -> Settings to choose an anchor and it will crop based on position. Alternatively, the user can select `Use reference` which uses an anchor point based on how the reference image is cropped.

image


damithc:
`Reference anchor` is a bit too cryptic for the use. `Follow the reference image`?

jamos-tay:
@damithc, Okay, changed to `Follow Reference Image`
@Skaty My bad, the file didn't delete properly

Issue #1276, opened 2017-04-02
Add SyncLab icons


Kenneth-LJS:
SyncLab does not have any icons designed for the following buttons:

1) Copy (currently "borrowing" icon from ColorsLab)
2) Paste
3) Edit formats
4) Delete

Issue #1275, opened 2017-04-02
Error message for shapes that cannot be copied


Kenneth-LJS:
Certain shapes cannot be pasted. When SyncLab tries to copy them, it throws an uncaught exception.

To reproduce:
1) Create a new powerpoint presentation.
2) Select the "Title", which is a textbox created by default on new presentations.
3) Open sync labs and click "Copy"
4) Click "Ok" on the format dialog.

The exception should be caught, and a meaningful error message should be shown.

damithc:
I think "Title" is a special type of a shape called 'placeholders' with special behaviors. For example, they cannot be grouped either.

Kenneth-LJS:
@damithc Makes sense. I'm now catching it so that users will not see the raw exception.

Issue #1274, opened 2017-04-02
Save SyncLab shapes in different file


Kenneth-LJS:
SyncLab currently saves format shapes into the design template, which can be deleted when actions are undone (ctrl-z).

Save the shapes into a different powerpoint presentation so that they are unaffected by undos.

Issue #1273, opened 2017-04-02
Allow deletion of formats in SyncLab


Kenneth-LJS:
The current GUI in SyncLab does not allow for users to delete the formats they do not want. Add functionality to remove the formats from the list.

PR #1277, opened 2017-04-02
Sync Lab Updates

Kenneth-LJS:
Changes:

image

1. Copy, apply format, set formats and delete format icons have been added to the UI. #1276
image
From left to right: Paste, Edit, Delete
1. Hovering over the icons will display a tooltip indicating what the button does. #1278
image
image
image
image

1. Copy button now looks responsive when clicking
image
1. Items can be deleted by clicking the delete button with the trash icon. #1273
1. New formats added to the list will be selected when copying. #1268
1. Formats are saved into a different powerpoint file, and are hence unaffected by undo. #1274
1. Un-copyable items display an error message when trying to copy them #1275
1. Display images will now use a separate template to save, hence hiding the SyncLab design template away from the user (we don't want them to touch this). #1288
1. Templates will be saved on attempting to exit. #1287
1. Fixed undo entries (broke when 2 separate PRs were merged)

Icons used are from Google's open source Material Icons set.


damithc:
Be careful when using icons that already has meaning associated with them. The first one looks like `download` and the second one looks like `settings`

Kenneth-LJS:
This looks the most appropriate icon so far. The typical copy/cut/paste icons can be used if you like.

image

I avoided using them because the clipboard by itself might not look intuitive.

PowerPointLabs/PowerPointLabs/SyncLab/SyncLabShapeStorage.cs
```diff
@@ -0,0 +1,95 @@
...
+
+using PowerPointLabs.Models;
+
+namespace PowerPointLabs.SyncLab
+{
+    public class SyncLabShapeStorage : PowerPointPresentation
+    {
+#pragma warning disable 0618
+
+        int nextKey = 0;
```
jamos-tay: Access modifier

damithc:
Yes, clipboard is not very intuitive for our case. May be the one you have is OK for paste.

PR #1292, opened 2017-04-12
Remove unused codes

IanTeo:
Remove unused codes in Sync Lab for dev-release


IanTeo:
Some other unused codes are already removed in PR #1291

PR #1318, opened 2017-05-09
Merge Into One Group should not replace current clipboard item  #1312 

### LINKED ISSUE #1311: Paste Lab should have multiple undo entries
When pasting multiple objects using paste lab, no new undo entries are created.
This means that all paste actions are removed in a single undo. One Paste action should require one undo.

The fix could be to create a new undo entry before every paste operation

### LINKED ISSUE #1312: Merge Into One Group should not replace current clipboard item
When using "Merge Into One Group" from the context menu, the current clipboard item will be replaced by the first item selected. It should not replace the current clipboard item.

I have 2 groups of items (**The first item is the item that is in the clipboard**):
1) image
2) image

Now when I use "Merge Into One Group" (Select the second item first):
image

Now when I paste, **the second item is pasted**. This behavior is weird, because the user did not copy the second item at all.
image

initialshl:
Fixes #1312 
Also fixes #1311  (closed #1316 in favor of this PR)

### DIFF
--- a/PowerPointLabs/PowerPointLabs/ActionFramework/Action/PasteLab/PasteLabActionHandler.cs
+++ b/PowerPointLabs/PowerPointLabs/ActionFramework/Action/PasteLab/PasteLabActionHandler.cs
@@ -1,0 +1,29 @@
+﻿using System.Windows;
+
+using Microsoft.Office.Interop.PowerPoint;
+
+using PowerPointLabs.ActionFramework.Common.Extension;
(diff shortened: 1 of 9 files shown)


IanTeo:
Just wondering, what is the advantage to using this feature over the normal merge that PowerPoint already has?

Skaty:
@IanTeo Normal merge? You mean the Group functionality? Well, that one doesn't preserve animations. All of the features in PasteLab should preserve animations.

Issue #1293, opened 2017-04-15
Add FT for Sync Lab


IanTeo:
We should add FT for Sync Lab as an end-to-end integration test

PR #1324, opened 2017-05-11
Add FT for Sync Lab #1293

IanTeo:
The base for sync lab FT
Currently, the implementation is hackish because I am unable to get the pop up pane with the controller (Using FunctionalTestExtensions)

### DIFF
--- a/PowerPointLabs/Test/FunctionalTest/SyncLabTest.cs
+++ b/PowerPointLabs/Test/FunctionalTest/SyncLabTest.cs
@@ -1,0 +1,101 @@
+﻿using System;
+using System.Drawing;
+using System.Windows.Forms;
+using TestInterface;
+using Microsoft.Office.Interop.PowerPoint;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Test.Util;
+using Point = System.Drawing.Point;
+using System.Collections.Generic;
+
+namespace Test.FunctionalTest
+{
+    [TestClass]
+    public class SyncLabTest : BaseFunctionalTest
+    {
+        private const int OriginalShapesSlideNo = 4;
+        private const string UnrotatedRectangle = "Rectangle 3";
+        private const string Oval = "Oval 4";
+        private const string RotatedArrow = "Right Arrow 5";
+        private const string CopyFromShape = "CopyFrom";
+
+
+        protected override string GetTestingSlideName()
+        {
+            return "SyncLab.pptx";
+        }
+
+        [TestMethod]
+        [TestCategory("FT")]
+        public void FT_SyncLabTest()
+        {
+            var syncLab = PplFeatures.SyncLab;
+            syncLab.OpenPane();
+
+            TestCopyDialog(syncLab);
+        }
+
+        private void TestCopyDialog(ISyncLabController syncLab)
+        {
+            var actualSlide = PpOperations.SelectSlide(4);
+
+            // No item selected
+            MessageBoxUtil.ExpectMessageBoxWillPopUp("", "Please select one item to copy.", syncLab.OpenCopyDialog, "Ok");
+
+            // 2 item selected
+            List<String> shapes = new List<string> { Oval, RotatedArrow };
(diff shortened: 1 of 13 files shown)


PowerPointLabs/Test/FunctionalTest/SyncLabTest.cs
```diff
@@ -0,0 +1,76 @@
...
+            MessageBoxUtil.ExpectMessageBoxWillPopUp("", "Please select one item to copy.", syncLab.Copy, "Ok");
+
+            // copy successful
+            PpOperations.SelectShape(CopyFromShape);
+            new Task(() =>
+            {
+                ThreadUtil.WaitFor(1000);
+                SendKeys.SendWait("{ENTER}");
+            }).Start();
+            syncLab.Copy();
```
IanTeo: @kai33 Do you have any ideas on a better way to do this? I have tried using the MessageBoxUtil and FunctionalTestExtensions class but both are unable to detect the WPF pop up window

kai33: any available release version I can try the sync lab?

IanTeo: There is a release for dev-release

damithc: I don't think the dev release worked. @Skaty is going to redo the dev release. 
@Skaty we need to do more frequent releases during summer. You may want to teach @IanTeo or @initialshl how to do releases to share the load with you.

Skaty: @damithc Ok noted! :+1:

kai33: keep me updated when it's released

IanTeo: @kai33 I believe @Skaty made the dev-release already. You can check it out

kai33: Ok, did you mean this window (screenshot)? I thought it's a msg box, but in fact a customized window. so that's why msgboxutil doesn't work (only works for msgbox as its name indicates)
capture
so in this case, u'll need to write a controller for it. similar to what u've done for synclab's controller, so like a controller's controller.

kai33: you can also delay the enhancement for a later PR, up to u

IanTeo: Hm.. I tried that and previously and encountered 2 problems

1. The window blocks all execution.
2. The FunctionalTestExtensions class, which is used to get the "_pane" does not work for the pop up. I suspect it is because FunctionalTestExtensions is only able to retrieve a Windows Form object. Does this mean I will have to wrap the pop up in a Windows Form? Is there anyway to just retrieve a WPF object?

kai33: 1. the window is a dialog, which is blocking by nature. i'd suggest make it non-dialog during FT.
2. return value of `FunctionalTestExtensions.GetTaskPane` is the winform instance. u can expose the WPF instance by providing an API form the winform one to access

IanTeo: @kai33 so are you saying that all our WPF should be wrapped in a winform? Is there any reason why it is done like that?

kai33: ?? i didn't say that. that's for task pane only.

IanTeo: But my customized window pop up is not a task pane

kai33: sorry i didn't get your question.
To me, the problem can be resolved as long as
* the window is non-blocking
* write a controller to access the window for FT

Issue #1310, opened 2017-05-08
Too many copies of Sync Lab storage file


IanTeo:
Each time the user interacts with Sync Lab, a new file is created and does not get deleted.
Over time, many files will appear and clutter the hard disk.

Maybe we can do it similar to how Shapes Lab is doing the storage.

PR #1330, opened 2017-05-12
PowerPoint still open in background after exiting #1309

### LINKED ISSUE #1309: PowerPoint still open in background after exiting
Sync Lab uses an extra PowerPoint in the background to "save" the users formats.

But when the user closes PowerPoint, the extra PowerPoint does not get closed and remains in the background. It should close when the user exits PowerPoint.

IanTeo:
Fixes #1309 and #1310 (This fix solves both problems)

- Changed SyncLabShapeStorage to become a singleton class. This is to allow other classes to access the same ppt file instead of creating many "temp" copies
- Deletes temp images when done with them, to avoid clogging up user's hard disk

### DIFF
--- a/PowerPointLabs/PowerPointLabs/SyncLab/SyncLabShapeStorage.cs
+++ b/PowerPointLabs/PowerPointLabs/SyncLab/SyncLabShapeStorage.cs
@@ -9,13 +9,25 @@ namespace PowerPointLabs.SyncLab
     public class SyncLabShapeStorage : PowerPointPresentation
     {
 
+        public const int FormatStorageSlide = 0;
+
         private int nextKey = 0;
 
-        public SyncLabShapeStorage() : base()
+        private static SyncLabShapeStorage storage;
+
+        public static SyncLabShapeStorage GetInstance()
+        {
+            if (storage == null)
+            {
+                storage = new SyncLabShapeStorage();
+            }
+            return storage;
+        }
+
+        private SyncLabShapeStorage() : base()
         {
             Path = System.IO.Path.GetTempPath();
-            Name = string.Format(TextCollection.SyncLabStorageFileName,
-                                 DateTime.Now.ToString("yyyyMMddHHmmssffff"));
+            Name = TextCollection.SyncLabStorageFileName;
             OpenInBackground();
             ClearShapes();
         }
@@ -87,7 +99,7 @@ namespace PowerPointLabs.SyncLab
                 GetSlide(1).Delete();
             }
             AddSlide();
-            Slides[0].DeleteAllShapes();
+            Slides[FormatStorageSlide].DeleteAllShapes();
         }
     }
 }
--- a/PowerPointLabs/PowerPointLabs/SyncLab/SyncLabShapeStorage.cs
+++ b/PowerPointLabs/PowerPointLabs/SyncLab/SyncLabShapeStorage.cs
@@ -6,22 +6,19 @@ using PowerPointLabs.Models;
 
 namespace PowerPointLabs.SyncLab
 {
-    public class SyncLabShapeStorage : PowerPointPresentation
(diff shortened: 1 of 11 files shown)


PowerPointLabs/PowerPointLabs/Utils/Graphics.cs
```diff
@@ -81,13 +81,14 @@ public static void ExportShape(ShapeRange shapeRange, string exportPath)
 
         public static Bitmap ShapeToImage(Shape shape)
         {
-            string fileName = "synclabtemp.png";
+            string fileName = TextCollection.SyncLabImageStorageFileName;
```
initialshl: This method resides in Graphics which may be used by various labs, so it might be better to no specify synclab in the name.

PowerPointLabs/PowerPointLabs/SyncLab/SyncLabShapeStorage.cs
```diff
@@ -9,13 +9,25 @@ namespace PowerPointLabs.SyncLab
...
+        public const int FormatStorageSlide = 0;
+
         private int nextKey = 0;
 
-        public SyncLabShapeStorage() : base()
+        private static SyncLabShapeStorage storage;
+
+        public static SyncLabShapeStorage GetInstance()
+        {
+            if (storage == null)
```
kai33: it's rare the there can be race condition here, do u want to lock/sync this method?

PowerPointLabs/PowerPointLabs/Utils/Graphics.cs
```diff
@@ -79,15 +79,23 @@ public static void ExportShape(ShapeRange shapeRange, string exportPath)
                               slideHeight, PpExportMode.ppScaleToFit);
         }
 
-        public static Image ShapeToImage(Shape shape)
+        public static Bitmap ShapeToImage(Shape shape)
```
kai33: it's rare the there can be race condition here, do u want to lock/sync this method?

IanTeo: Is there any other place in pptlabs that implements lock/sync that I should follow? Or should I just implement it using c# Lock keyword?

kai33: yes lock is enough

IanTeo:
I have made the changes as follows:

- Lazy Type is used for the singleton class, as it is preferable to using double-checked locking (In terms of run time speed)
- For the ShapeToImage method, a lock is used over the entire method. (Need to ensure that the file is freed and deleted before allowing another thread to access)

PowerPointLabs/PowerPointLabs/SyncLab/SyncLabShapeStorage.cs
```diff
@@ -6,22 +6,19 @@
...
-    public class SyncLabShapeStorage : PowerPointPresentation
+    public sealed class SyncLabShapeStorage : PowerPointPresentation
     {
 
         public const int FormatStorageSlide = 0;
 
         private int nextKey = 0;
 
-        private static SyncLabShapeStorage storage;
+        private static readonly Lazy<SyncLabShapeStorage> StorageInstance =
```
initialshl: Is this instance created even when SyncLab is not opened/used? Possible to put it into the getter?

IanTeo: Lazy Type will only create the instance once when it is needed. Subsequently, it will not be created. In a sense, it is a built in singleton with thread safe.

IanTeo: I actually used this article as a guide: http://csharpindepth.com/Articles/General/Singleton.aspx

Issue #1341, opened 2017-05-17
Improve FT for sync labs


IanTeo:
Currently, the FT for sync labs is quite hackish.
It should be improved to use a controller to control the pop up window

PR #1344, opened 2017-05-17
Improve FT for sync labs #1341

IanTeo:
Improved tests for Sync Lab FT.

- Get the controller for Dialog by saving the current open dialog in SyncPaneWPF and exposing it to the FT Controller
- Change SyncFormatDialog to use EventListeners so that it does not block execution. If user attempts to open multiple dialogs, the dialogs will replace each other. This way, only 1 dialog maximum can be open at any time
- FT tests for the error messages and tests sync once. This is because we want to test if users can use all the buttons. Each individual sync test is done in UT

### DIFF
--- a/PowerPointLabs/PowerPointLabs/SyncLab/View/SyncPaneWPF.xaml.cs
+++ b/PowerPointLabs/PowerPointLabs/SyncLab/View/SyncPaneWPF.xaml.cs
@@ -6,6 +6,7 @@ using System.Windows.Media.Imaging;
 using Microsoft.Office.Interop.PowerPoint;
 
 using static PowerPointLabs.ActionFramework.Common.Extension.ContentControlExtensions;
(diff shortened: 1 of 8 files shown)


PowerPointLabs/PowerPointLabs/SyncLab/View/SyncPaneWPF.xaml.cs
```diff
@@ -14,7 +15,7 @@ namespace PowerPointLabs.SyncLab.View
     /// </summary>
     public partial class SyncPaneWPF : UserControl
     {
-
+        public SyncFormatDialog dialog;
```
PPTLabsBot: MINOR Make this field 'private' and encapsulate it in a 'public' property. [rule](https://sonarqube.com/coding_rules#rule_key=csharpsquid%3AS1104)

Issue #1319, opened 2017-05-09
Progress for public release


initialshl:
Documentation (in Google Drive):
- [x] Positions Lab - Last edited: 30 May 2017
- [x] Resize Lab (Requesting for reviews) - Last edited: 11 May 2017
- [x] Sync Lab  - Last edited: 30 May 2017
- [x] Crop Lab - Last edited: 29 May 2017
- [x] Paste Lab - Last edited: 23 May 2017
- [x] Timer Lab - Last edited: 29 May 2017
- [x] Shortcuts Lab (context menu) - Last edited: 26 May 2017

Tutorials (in Google Drive):
- [x] Positions Lab - Last edited: 25 May 2017
- [x] Resize Lab - Last edited: 25 May 2017
- [x] Sync Lab - Last edited: 25 May 2017
- [x] Crop Lab - Last edited: 25 May 2017
- [x] Paste Lab - Last edited: 25 May 2017
- [x] Timer Lab - Last edited: 25 May 2017

Unit/Functional Tests:
- [x] Positions Lab
- [x] Resize Lab
- [x] Sync Lab - #1344 
- [x] Crop Lab
- [x] Paste Lab - #1361 
- [x] Timer Lab - #1347 

Icons:
- [x] Positions Lab
- [x] Resize Lab
- [x] Sync Lab - #1376 
- [x] Crop Lab
- [x] Paste Lab - #1369 
- [x] Timer Lab

IanTeo:
@AngShiYa Could you do a review for the Timer Lab documentation?
@Skaty Could you do a review for the Paste Lab documentation?
@jamos-tay Could you do a review for the Crop Lab documentation?

jamos-tay:
@IanTeo Okay have looked through and made corrections. Just a minor detail which I added as a comment.

AngShiYa:
@IanTeo Sure. It looks good. Added a comment for the timer's position.

IanTeo:
@Kenneth-LJS Could you do a review for the Sync Lab documentation?

IanTeo:
Thanks for the review @jamos-tay @AngShiYa

initialshl:
@jamos-tay Thanks! Do you still have the slides for the documentation?

PR #1347, opened 2017-05-18
Timer lab FT

IanTeo:
Part of #1319 

FT for Timer Lab. Timer Lab will not have any UT for now, as I don't think there are any that are needed.
Also improved the error messages for Asserts


PowerPointLabs/Test/Util/SlideUtil.cs
```diff
@@ -20,25 +20,88 @@ public static void IsSameShape(Shape expShape, Shape actualShape)
...
         }
 
         public static void IsSameZOrderPosition(Shape expShape, Shape actualShape)
         {
             Assert.IsTrue(expShape.ZOrderPosition == actualShape.ZOrderPosition,
-                "different shape z order. exp:{0}, actual:{1}", expShape.ZOrderPosition, actualShape.ZOrderPosition);
+                "different shape z order for {0}. exp:{1}, actual:{2}", expShape.Name, expShape.ZOrderPosition, actualShape.ZOrderPosition);
+        }
+
+        public static void IsSameShapes(Slide expSlide, Slide actualSlide)
```
kai33: does this method work for this case:
slide1 has shape A, B, C;
slide2 has shape A, B, D?

kai33: got it should be working

PowerPointLabs/Test/Util/SlideUtil.cs
```diff
@@ -20,25 +20,88 @@ public static void IsSameShape(Shape expShape, Shape actualShape)
...
+                        {
+                            Assert.IsTrue(expShape.Type == actualShape.Type,
+                                "different shape type for {0}. exp:{1}, actual:{2}", expShape.Name, expShape.Type, actualShape.Type);
+                            Assert.IsTrue(expShape.GroupItems.Count == actualShape.GroupItems.Count,
+                                "different number of shapes in group for {0}. exp:{1}, actual:{2}",
+                                expShape.Name, expShape.GroupItems.Count, actualShape.GroupItems.Count);
+
+                            for (int k = 1; k <= expShape.GroupItems.Count; k++)
+                            {
+                                IsSameShape(expShape.GroupItems[k], actualShape.GroupItems[k]);
```
kai33: how about `group of group` case?

PowerPointLabs/PowerPointLabs/FunctionalTestInterface.Impl/Controller/TimerLabController.cs
```diff
@@ -0,0 +1,107 @@
...
+            });
+        }
+
+        public void ClickCreateButton()
+        {
+            if (_pane != null)
+            {
+                UIThreadExecutor.Execute(() =>
+                {
+                    _pane.TimerPaneWPF.createButton.RaiseEvent(new RoutedEventArgs(ButtonBase.ClickEvent));
```
IanTeo: Currently, all the other controllers are using the if (_pane != null) thing. To keep the code consistent, I think we should keep it as if (_pane != null) for now.

On another note, while I was looking through how to implement the null-conditional operators, it does not seem possible to implement it for stuff like line 49, where I did this:

_pane.TimerPaneWPF.DurationTextBox.Value = value;

Here I would have to check if pane is null before doing the assignment, right?

PowerPointLabs/Test/Util/SlideUtil.cs
IanTeo: Oh. I did not think know that you could have a group within a group in PowerPoint. I will make the changes

IanTeo:
Updated function to handle multi groups using recursion

PowerPointLabs/PowerPointLabs/FunctionalTestInterface.Impl/Controller/TimerLabController.cs
kai33: Yes it doesnt support assignment. But still it can be done something like this `_pane?.setDurationTextBox(...)`.
Up to u, it just makes the codes cleaner. Other controller were written in older c#, so didnt use it

IanTeo: I think I will do it in a separate issue for all the controllers (When refactoring)

PowerPointLabs/Test/FunctionalTest/TimerLabTest.cs
```diff
@@ -0,0 +1,156 @@
...
+            AssertIsSame(OriginalSlideNo, DurationInvalidSlideNo);
+        }
+
+        private void AssertIsSame(int actualSlideNo, int expectedSlideNo)
+        {
+            var actualSlide = PpOperations.SelectSlide(actualSlideNo);
+            var expectedSlide = PpOperations.SelectSlide(expectedSlideNo);
+
+            SlideUtil.IsSameShapes(expectedSlide, actualSlide);
+            SlideUtil.IsSameAnimations(expectedSlide, actualSlide);
```
IanTeo: There is an IsSameAnimations check here, which is called by all of the test methods

initialshl: Oh ok i missed that, the method needs to be updated to check for the motion path and duration though.

IanTeo: I tried updating the method to check for the motion path and duration, but came up with a weird security exception. Maybe we can merge this first as we figure out how to fix that.

PR #1357, opened 2017-05-22
Paste Lab: Group Name should remain the same #1355 

### LINKED ISSUE #1355: Paste Lab: Group Name should remain the same
After actions that edit the group, the group name changes.
The group name should not change, in case the user has assigned a name to their group.

Features that change group name:

- Paste Into Group
- Replace With Clipboard (For groups)
- Merge Into One Group (I guess we should follow the name of the group that was first selected)

initialshl:
Fixes #1355 

Related fix: Paste Into Group no longer creates nested group if only a single group is selected.


IanTeo:
I was also thinking about the behavior of merge into one group. But don't you think that the user expects it to become one group rather than a group of groups? Because to get a group of groups I could just use the PowerPoint group feature right? @damithc what do you think?

damithc:
My feeling is `Merge into one group` should retain name and animations of the first group. If more groups are selected, they should be inserted (as groups) to the first group. i.e. we select g1, o1, g2 and use this feature, not g1 should contain o1 and g2 (treat both o1 and g1 the same way)

IanTeo:
I supposed that this is what you mean?

If we select **First Group** followed by **My Group** and `Merge Into One Group`, we should get this result:
image

damithc:
> I supposed that this is what you mean?

Yes. 👍

initialshl:
Okay, the changes have been made.

PR #1368, opened 2017-05-24
Allow users to use Sync Lab when text is selected #1367

### LINKED ISSUE #1367: Allow users to use Sync Lab when text is selected
Currently, when the user selects the text in a shape, sync lab will prompt an error message asking them to select a shape. We should allow the user to sync, even when select the text in the shape.

IanTeo:
Fixes #1367


initialshl:
Tested and working in PPT 2016.

Issue #1348, opened 2017-05-19
Icons for Sync Lab


IanTeo:
Some Icons, like the copy and sync icons are not suitable.
We should create better, more suitable icons before releasing the lab

PR #1376, opened 2017-05-25
Icons for Sync Lab #1348

IanTeo:
Fixes #1376 

@damithc @AngShiYa How do you find these Icons? I think the icon for Copy looks kind of weird here. Any suggestions?

image

EDIT: I also increased the size of the sync lab panel item so that I could put bigger icons and let the user see a slightly clearer image of their shape


damithc:
Perhaps we stick the blue color for small icons?

AngShiYa:
The copy button looks like add. I'm not sure what the brush and palette represents. I think we should just stick with the conventional copy/paste and edit icon. For example,
sample

IanTeo:
I don't think blue would be very suitable in this case. I think blue would look kind of weird.
image


image

PR #1388, opened 2017-05-29
Crop to same dimensions: support normal shapes for reference object #1383 

### LINKED ISSUE #1383: Crop to same dimensions: support normal shapes for reference object
As the reference shape is only used to determine size, it need not be a picture.
image

initialshl:
Fixes #1383 

UT added for this new case.
Tested on PPT 2016.


PowerPointLabs/PowerPointLabs/CropLab/CropToSame.cs
```diff
@@ -12,22 +16,25 @@ public class CropToSame
...
+            Shape refObj = shapeRange[1];
+            bool isRefObjShape = Graphics.IsShape(refObj);
+
+            float refScaleWidth = Graphics.GetScaleWidth(refObj);
+            float refScaleHeight = Graphics.GetScaleHeight(refObj);
+
+            float cropTop = isRefObjShape ? Epsilon : Math.Max(refObj.PictureFormat.CropTop, Epsilon);
+            float cropBottom = isRefObjShape ? Epsilon : Math.Max(refObj.PictureFormat.CropBottom, Epsilon);
+            float cropLeft = isRefObjShape ? Epsilon : Math.Max(refObj.PictureFormat.CropLeft, Epsilon);
+            float cropRight = isRefObjShape ? Epsilon : Math.Max(refObj.PictureFormat.CropRight, Epsilon);
```
IanTeo: You should use an if statement. Doing it like this will end up checking the same thing 4 times.

PR #1389, opened 2017-05-30
Sync Lab: Deleting shape during copy causes exception #1387

### LINKED ISSUE #1387: Sync Lab: Deleting shape during copy causes exception
1. Open Sync Lab pane
1. Select a shape
1. Press `Copy` which opens a popup sync lab window
1. Delete the shape
1. Press `OK` on the popup sync lab window
1. Unhandled exception occurs

IanTeo:
Fixes #1387 

To fix the problem, I switched the dialog back to a modal dialog (as it makes more sense, the user should not edit the slide before finishing his selection to copy). This also prevents other problems such as the user editing the style of the shape while the copy dialog is open.

But this means that the tests are blocked by the dialog when it opens. To solve this problem, we run the method in a separate thread similar to how MessageBoxUtil's ExpectMessageBoxWillPopUp() handles it.


kai33:
does it make the testing more difficult?
i'm ok to have some hacks like [FT mode](https://github.com/PowerPointLabs/PowerPointLabs/blob/master/PowerPointLabs/PowerPointLabs/FunctionalTestInterface.Impl/PowerPointOperations.cs#L34) to turn it non-dialog if it can make testing easier. up to u

IanTeo:
I think it is fine, I am handling it like a message box, where I had to run another thread before clicking the button. This way, the hack is on the test files rather than the codes that are executing, which I think would be a better way

PR #1394, opened 2017-05-31
Timer has uncaught exception when pasting invalid properties #1392 

### LINKED ISSUE #1392: Timer has uncaught exception when pasting invalid properties
1. Copy the alphabet "a"
1. Open Timer Lab
1. Click `Create`
1. Paste into the Height input field
1. Click outside the input field (lose focus)

Unhandled exception will occur

initialshl:
Fixes #1392

### DIFF
--- a/PowerPointLabs/PowerPointLabs/TimerLab/TimerLabPaneWPF.xaml.cs
+++ b/PowerPointLabs/PowerPointLabs/TimerLab/TimerLabPaneWPF.xaml.cs
@@ -439,7 +439,23 @@ namespace PowerPointLabs.TimerLab
 
         private void WidthTextBox_PreviewTextInput(object sender, System.Windows.Input.TextCompositionEventArgs e)
         {
-            e.Handled = IsNumbersOnly(e.Text);
+            e.Handled = !IsNumbersOnly(e.Text);
+        }
+
+        private void WidthTextBox_TextBoxPasting(object sender, DataObjectPastingEventArgs e)
+        {
+            if (e.DataObject.GetDataPresent(typeof(String)))
+            {
+                String text = (String)e.DataObject.GetData(typeof(String));
+                if (!IsNumbersOnly(text))
+                {
+                    e.CancelCommand();
+                }
+            }
+            else
+            {
+                e.CancelCommand();
+            }
         }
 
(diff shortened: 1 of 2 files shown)


PowerPointLabs/PowerPointLabs/TimerLab/TimerLabPaneWPF.xaml.cs
```diff
@@ -439,7 +439,23 @@ private void WidthTextBox_Loaded(object sender, RoutedEventArgs e)
 
         private void WidthTextBox_PreviewTextInput(object sender, System.Windows.Input.TextCompositionEventArgs e)
```
IanTeo: Could you change this to an import, since we use it multiple times

PowerPointLabs/PowerPointLabs/TimerLab/TimerLabPaneWPF.xaml.cs
```diff
@@ -439,7 +439,23 @@ private void WidthTextBox_Loaded(object sender, RoutedEventArgs e)
...
+        }
+
+        private void WidthTextBox_TextBoxPasting(object sender, DataObjectPastingEventArgs e)
+        {
+            if (e.DataObject.GetDataPresent(typeof(String)))
+            {
+                String text = (String)e.DataObject.GetData(typeof(String));
+                if (!IsNumbersOnly(text))
+                {
+                    e.CancelCommand();
```
IanTeo: This action and the one on line 442 do practically the same thing. We should have some consistency and use either one method or the other, not both.

PowerPointLabs/PowerPointLabs/TimerLab/TimerLabPaneWPF.xaml.cs
initialshl: Ok done

PowerPointLabs/PowerPointLabs/TimerLab/TimerLabPaneWPF.xaml.cs
initialshl: I think these should not be changed.
Line 442 does not have a CancelCommand method, and while this routed event has Handled property, it is better to use CancelCommand method to be more readable.

PR #1428, opened 2017-06-07
Position Swap may be stuck in infinite loop #1423

### LINKED ISSUE #1423: Position Swap may be stuck in infinite loop
Selecting these and pressing Swap under Positions Lab will cause PPT to become not responsive.

Selecting the two groups will be unable to swap.

Slide here: [Icons-PasteLab.pptx](https://github.com/PowerPointLabs/PowerPointLabs/files/1054240/Icons-PasteLab.pptx)

IanTeo:
Fixes #1423


IanTeo:
This solution will not work. Created another PR to fix the problem #1429

PR #1429, opened 2017-06-08
Position Swap may be stuck in infinite loop #1423

### LINKED ISSUE #1423: Position Swap may be stuck in infinite loop

IanTeo:
Fixes #1423 

Remove ZOrdering from PPShape. Use Graphics class instead
Change Positions Lab and Paste Lab to use Graphics class

### DIFF
--- a/PowerPointLabs/PowerPointLabs/Utils/Graphics.cs
+++ b/PowerPointLabs/PowerPointLabs/Utils/Graphics.cs
@@ -503,12 +503,12 @@ namespace PowerPointLabs.Utils
             // This is to account for the zOrder of the items in the group
             if (IsAGroup(lowerZOrderShape))
             {
-                higherZOrder -= originalShape.GroupItems.Count;
+                higherZOrder -= lowerZOrderShape.GroupItems.Count;
             }
 
             if (IsAGroup(higherZOrderShape))
             {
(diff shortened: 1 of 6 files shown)


PowerPointLabs/PowerPointLabs/Utils/Graphics.cs
```diff
@@ -457,7 +466,16 @@ public static void MoveZUntilInFront(Shape shiftShape, Shape destinationShape)
...
+            MoveZUntilBehind(shiftShape, destinationShape.ZOrderPosition);
+        }
+
+        /// <summary>
+        /// Moves shiftShape backward until it is behind destination zOrder index.
+        /// (does nothing if already behind)
+        /// </summary>
+        public static void MoveZUntilBehind(Shape shiftShape, int destination)
+        {
+            while (shiftShape.ZOrderPosition > destination)
```
initialshl: Seems like this is not working as intended, since the parameter is a constant instead of a reference when it is changing.

PR #1454, opened 2017-06-15
Pasting placeholder objects causes exceptions #1449 

### LINKED ISSUE #1449: Pasting placeholder objects causes exceptions
The paste lab features should be disabled, instead of doing nothing or showing an error message.

initialshl:
Fixes #1449

Since we are unable to detect whether a placeholder is in the clipboard, the solution is to catch and handle the COMExceptions that occur when pasting placeholders. 

This behavior (of doing nothing when pasting placeholders) is the same as PowerPoint's.


IanTeo:
I further tested paste with placeholder on `ReplaceWithClipboard` and `PasteIntoGroup`. 

- `ReplaceWithClipboard`: Nothing happened. I think that it is expected to replace the shape if we can paste the placeholder. (I think this is quite difficult to do, so we should keep the current behavior for now)
- `PasteIntoGroup`: Nothing happened, which is expected.

initialshl:
I have fixed the exceptions when pasting placeholders when they do not already exist on the slide, such that `PasteAtCursorPosition` pastes at cursor position, `PasteAtOriginalPosition` and `PasteToFillSlide` pastes them at their original positions (at the same size), and other pastes features do nothing.

PR #1509, opened 2017-07-10
Hide Shape and Add To Shapes Lab unable to select item in group #1464

### LINKED ISSUE #1464: Hide Shape and Add To Shapes Lab unable to select item in group
These features on the context menu are unable to get the items within groups. When we try to use these features for an item in a group, we end up doing the operation on the entire group.

The expected behavior should be that the action is applied to the item in the group, similar to Edit Name

IanTeo:
Fixes #1464


PowerPointLabs/PowerPointLabs/ActionFramework/ShortcutsLab/Hide/HideShapeActionHandler.cs
```diff
@@ -9,7 +9,13 @@ class HideShapeActionHandler : ActionHandler
     {
         protected override void ExecuteAction(string ribbonId)
```
initialshl: HideShapeHandlers are in the folder "Hide", should the folder be renamed to "HideShape"? Similarly to "TextCollection.HideSelectedShapeTag"

PowerPointLabs/PowerPointLabs/Models/PowerPointShapeGalleryPresentation.cs
```diff
@@ -4,10 +4,11 @@
 using System.Linq;
 using System.Text.RegularExpressions;
 using System.Windows.Forms;
-using Microsoft.Office.Core;
+
 using Microsoft.Office.Interop.PowerPoint;
 using PowerPointLabs.Utils;
```
initialshl: Can put a line break here too

IanTeo:
What should I name the folder? I changed it to ShapesLabPane because it is the codes to open the pane from the ribbon. The other labs like resize and positions lab are placed in a folder called "Resize Lab" and "Positions lab" respectively. But I felt it might be kind of weird for the folders to be ShapesLab -> ShapesLab

initialshl:
Maybe ShapesLab -> RibbonButton?

IanTeo:
How about if all the handlers for the ribbon buttons are in the folder itself?

initialshl:
Yup ok sure

PR #1516, opened 2017-07-17
Create LICENSE file

Skaty:
According to README.md, PowerPointLabs is licensed under GPLv2. We should have a LICENSE file, as this would make it clear the license PowerPointLabs is released under.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,339 @@
+                    GNU GENERAL PUBLIC LICENSE
+                       Version 2, June 1991
+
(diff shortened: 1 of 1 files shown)


IanTeo:
@skaty there is some weird error with SonarQube for your PR. Could you check it out?

Skaty:
@IanTeo Oh, it's because I am merging from another branch that doesn't have SonarQube set up. You can ignore that, since I would have to set up for a branch that I created using the GitHub UI.

PR #1541, opened 2017-08-01
Cannot open PPTLabs on PPT 2010 #1539

### LINKED ISSUE #1539: Cannot open PPTLabs on PPT 2010
A user was unable to use PPTLabs, as he encounters an error upon opening PowerPoint 2010.

```
The error occurs 100% of the times upon opening PowerPoint.
After this error, the menu of Labs is shown, but it does nothing (seems that it affects hooking to your code?)
The error is caused after installing all the latest packs, via the “Microsoft Office Configuration analyser” tool… which basically gets al updates released.
```

image001
image002

IanTeo:
Fixes #1539 

Since this is a quick fix, should we do a public release soon?


IanTeo:
Wow, I did not notice. But having to remove so many concerns me, as we might lose some functionality, especially from ContextMenuNotesEditText, which allows the user to use the "Speak Selected" functionality

initialshl:
We may have to disable that for the time being, and do a proper fix in dev-release to check for ppt 2010. The website doc will have to tell users that 'Speak Selected' may/does not work on ppt 2010.

Issue #1300, opened 2017-04-24
Typo in PSL


kai33:
`Delete all picture` -> `Delete all pictures`
![image_19](https://cloud.githubusercontent.com/assets/3591355/25347234/c10f83f2-294c-11e7-9ad4-9e7727826ebe.png)

PR #1548, opened 2018-01-08
Grammar fix for PictureSlidesLabWindow

AlexFJW:
- Fixes Typo in PSL #1300


AlexFJW:
Oops, didn't notice that. Thanks, will be keeping that in mind.

Issue #1576, opened 2018-01-25
Positions Lab: Should not require at least 2-3 shapes to be selected 


leeyh20:
Suggestion: If only one shape is selected, should it be expected to work like the built-in powerpoint "Align" button, so that users don't need to go to Format--> Align to align a single shape?

junkiattan:
Do you want the default align operation if only one shape is selected? Is this included in the desired behavior for the button?

leeyh20:
Yeah was wondering if we should put the default align operation inside the feature if only one shape is selected so that it feels complete.

PR #1579, opened 2018-01-26
#1576 Positions Lab: Should not require at least 2-3 shapes to be selected

junkiattan:
#1576 

I have included default align functionality for when one shape is selected for the following

1. Align-Left
2. Align-Right
3. Align-Top
4. Align-Bottom
5. Align-Center
6. Align-Center (Horizontal)
7. Align-Center (Vertical)

FTs and UTs returned no error

### DIFF
--- a/PowerPointLabs/PowerPointLabs/PositionsLab/PositionsLabMain.cs
+++ b/PowerPointLabs/PowerPointLabs/PositionsLab/PositionsLabMain.cs
@@ -142,7 +142,8 @@ namespace PowerPointLabs.PositionsLab
                 case PositionsLabSettings.AlignReferenceObject.SelectedShape:
                     if (toAlign.Count < 2)
                     {
-                        throw new Exception(PositionsLabText.ErrorFewerThanTwoSelection);
+                        // Insert default align right behaviour here
+                        toAlign.Align(MsoAlignCmd.msoAlignRights, MsoTriState.msoTrue);
                     }
 
                     selectedShapes = ConvertShapeRangeToPPShapeList(toAlign);
--- a/PowerPointLabs/PowerPointLabs/PositionsLab/PositionsLabMain.cs
+++ b/PowerPointLabs/PowerPointLabs/PositionsLab/PositionsLabMain.cs
@@ -232,7 +232,8 @@ namespace PowerPointLabs.PositionsLab
                 case PositionsLabSettings.AlignReferenceObject.SelectedShape:
                     if (toAlign.Count < 2)
                     {
-                        throw new Exception(PositionsLabText.ErrorFewerThanTwoSelection);
+                        // Insert default align bottom behaviour here
+                        toAlign.Align(MsoAlignCmd.msoAlignBottoms, MsoTriState.msoTrue);
                     }
 
                     selectedShapes = ConvertShapeRangeToPPShapeList(toAlign);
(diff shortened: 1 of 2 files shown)


PowerPointLabs/PowerPointLabs/PositionsLab/PositionsLabMain.cs
```diff
@@ -96,7 +96,8 @@ public static void AlignLeft(ShapeRange toAlign)
                 case PositionsLabSettings.AlignReferenceObject.SelectedShape:
                     if (toAlign.Count < 2)
                     {
-                        throw new Exception(PositionsLabText.ErrorFewerThanTwoSelection);
+                        // Insert default align left behaviour here
```
Skaty: There's reason to believe that the comments are quite useless, as we know that the `Align` method is the built-in PowerPoint align.

PowerPointLabs/PowerPointLabs/PositionsLab/PositionsLabMain.cs
```diff
@@ -361,7 +367,9 @@ public static void AlignCenter(ShapeRange toAlign, float slideHeight, float slid
                 case PositionsLabSettings.AlignReferenceObject.SelectedShape:
                     if (toAlign.Count < 2)
                     {
-                        throw new Exception(PositionsLabText.ErrorFewerThanTwoSelection);
+                        // Insert default align center behaviour here
+                        toAlign.Align(MsoAlignCmd.msoAlignMiddles, MsoTriState.msoTrue);
+                        toAlign.Align(MsoAlignCmd.msoAlignCenters, MsoTriState.msoTrue);
```
Skaty: Question: why the need to align twice?

PowerPointLabs/PowerPointLabs/PositionsLab/PositionsLabMain.cs
junkiattan: @Skaty 

Agreed. It was for the purposes of identifying the point of adding the code while I was working on this. Ill remove them

PowerPointLabs/PowerPointLabs/PositionsLab/PositionsLabMain.cs
junkiattan: In order to align center, you need to first align horizontally and then vertically

junkiattan:
@Skaty 

Thanx for the highlighting a potential problem with the removal of exception! I will look into the issue;)

junkiattan:
@Skaty 

Added relevant UTs to check for aligning functionality for one shape. Ran them and they all pass.

PowerPointLabs/Test/UnitTest/PositionsLab/PositionsLabAlignTest.cs
```diff
@@ -433,6 +433,117 @@ public void TestAlignOneCenterDefault()
             CheckShapes(expectedShapes, actualShapes);
         }
 
+        [TestMethod]
+        [TestCategory("UT")]
+        public void TestAlignOneLeftSingleShape()
```
Skaty: What's the difference between this and the method `TestAlignOneLeftDefault`? Same with the other methods.

junkiattan: The reference point is different. For the new tests, the reference points are the respective selected shapes.

Skaty: Thanks for the clarification, didn't spot that bit! :)

Skaty:
@leeyh20 Retest required, since new UTs are added.

leeyh20:
Retested on PPT 2010 with the new UTs. The tests all pass.

PR #1607, opened 2018-02-15
1595-SyncLab-The-Fill-Color-of-textboxes-are-inconsistent-when-using-SyncLab

### LINKED ISSUE #1375: Sync Lab: Line should not provide Shape Fill format
Shape Fill for a line is greyed out in PowerPoint, so it should not be available for Sync Lab copy format.

Also does nothing when such a fill is pasted on another shape.

### LINKED ISSUE #1595: SyncLab: The Fill Color of textboxes are inconsistent when using SyncLab
Occurs in PPT 2010
// Edit: occurs in PPT 2016. (probably 2013 as well)

Notice that for this textbox, the fill color is black when trying to sync with SyncLab.
image

When I create a new textbox, the fill color mysteriously becomes the other blue color.
image

@alexfjw What do you think?

alexfjw:
fixes #1595 
fixes #1375

### DIFF
--- a/PowerPointLabs/PowerPointLabs/SyncLab/ObjectFormats/FillFormat.cs
+++ b/PowerPointLabs/PowerPointLabs/SyncLab/ObjectFormats/FillFormat.cs
@@ -44,10 +44,14 @@ namespace PowerPointLabs.SyncLab.ObjectFormats
             try
             {
                 // force msoFillMixed to msoFillSolid
-                // freshly created textboxes have the msoFillMixed type for unknown reasons
+                // freshly created textboxes have the msoFillMixed type 
                 // otherwise, msoFillMixed only appears when multiple shapes are selected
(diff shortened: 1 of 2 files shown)


leeyh20:
Tested on PPT 2010. Manual testing works but when I run the FT_SyncLabTest it seems to be stuck at 
image
Nothing happens and the test does not proceed until I click something.

alexfjw:
Oops, thanks for the check. I'll get that fixed

alexfjw:
Looks like lines are also msoMixedFill. 
The code written coincidentally forces to make lines have a solid fill.
This throws an exception, and the Sync dialog notes that lines cannot sync fills. 
This is desired behavior which fixes #1375

leeyh20:
LGTM! Works on PPT 2010.

Issue #1608, opened 2018-02-15
Error while opening pptx file


damithc:
I have a pptx file that gives this error. I've been using that file for many years. Error started after I modified the file today using PowerPointLabs.
The file still opens and is usable though.
```
Application (unknown member) : Invalid request.  There is no currently active document window.

System.Runtime.InteropServices.COMException
Application (unknown member) : Invalid request.  There is no currently active document window.
Stack Trace:
   at Microsoft.Office.Interop.PowerPoint._Application.get_ActiveWindow()
   at PowerPointLabs.Models.PowerPointCurrentPresentationInfo.get_CurrentSelection()
   at PowerPointLabs.ActionFramework.Common.Extension.ActionFrameworkExtensions.GetCurrentSelection()
   at PowerPointLabs.ActionFramework.Common.Extension.BaseHandlerExtensions.GetCurrentSelection(BaseHandler handler)
   at PowerPointLabs.ActionFramework.AnimationLab.DrillDownEnabledHandler.GetEnabled(String ribbonId)
   at PowerPointLabs.ActionFramework.Common.Interface.EnabledHandler.Get(String ribbonId)
```

damithc:
Env: Win 10, PPT 2013

Skaty:
- What kind of modifications were made to the file?
- PPTLabs version?

damithc:
> What kind of modifications were made to the file?
> PPTLabs version?

I don't think it's related to the modification. I get the error even when I open blank pptx files. Looks like some meta data expected by PowerPointLabs is corrupted. There was a 'PowerPoint auto-saved a file; do you want to keep it?' message when I opened PPT today. The file name had SyncLab or ShapesLab in it, but I deleted it without thinking too much. Probably that's what caused the problem. 

Latest public release.

leeyh20:
Seems similar to #1543, but I thought it was only for empty ppts. I'll try investigating this.

alexfjw:
I also encountered a similar problem recently when creating new powerpoints. A blank powerpoint's fine but selecting one of those preset templates seems to give the same error on 2016. 
`Invalid request.  There is no currently active document window.` 
I'll check that out again in a bit.

damithc:
I can open the same file in my other comp without an error. However, I noticed these warnings. Looks like something went wrong with our meta-data files 

[...]

Issue #1543, opened 2017-08-10
Spotlight Enabled causing exception on presentation open


initialshl:
SpotlightEnabled function causes exception when opening a PowerPoint document.

```
System.Runtime.InteropServices.COMExceptionApplication (unknown member) : Invalid request.  
There is no currently active document window.
Stack Trace:   
at Microsoft.Office.Interop.PowerPoint._Application.get_ActiveWindow()   
at PowerPointLabs.Models.PowerPointCurrentPresentationInfo.get_CurrentSelection()   
at PowerPointLabs.ActionFramework.Common.Extension.ActionFrameworkExtensions.GetCurrentSelection()   
at PowerPointLabs.ActionFramework.Common.Extension.BaseHandlerExtensions.GetCurrentSelection(BaseHandler handler)   
at PowerPointLabs.ActionFramework.EffectsLab.AddSpotlightEnabledHandler.GetEnabled(String ribbonId)   
at PowerPointLabs.ActionFramework.Common.Interface.EnabledHandler.Get(String ribbonId)
```
Occurs at https://github.com/PowerPointLabs/PowerPointLabs/blob/master/PowerPointLabs/PowerPointLabs/ActionFramework/EffectsLab/Spotlight/AddSpotlight/AddSpotlightEnabledHandler.cs#L16, we may need to debug whether this happens for all enabledhandlers, and handle exceptions in a helper method when nothing is selected (for all enabled handlers).

https://github.com/PowerPointLabs/PowerPointLabs/blob/master/PowerPointLabs/PowerPointLabs/ThisAddIn.cs#L955 also seems to have duplicate Ribbon.RefreshRibbonControl("AddSpotlight") and Ribbon.RefreshRibbonControl("AddSpotlightButton"), the unused one should be removed.

junkiattan:
hey @initialshl 

I have removed duplicate Ribbon.RefreshRibbonControl("AddSpotlight").

However with regards to the other problem, I cannot replicate the exception you are reporting. I have tried to open a powerpoint document

initialshl:
Not sure how or on what environment/OS this exception occurs, may need to follow-up with the user on support forum.

junkiattan:
@initialshl 

I have emailed the person who submitted the error report to ask for more details. Will act on his reply

leeyh20:
Might want to see this https://support.microsoft.com/en-hk/help/285472/run-time-error--2147188160-on-activewindow-or-activepresentation-call

[...]

PR #1613, opened 2018-02-17
#1608 Fixes no currently active document window

leeyh20:
Issues are confirmed to be fixed: #1615
Issues are supposedly fixed: #1608 and #1543 

It might not only affect EnabledHandler but anything that tries to get selection without an active window. However, this method of catching the exception when using ActiveWindow will lead to NullPointerException if null checks are not made when using ActionFramework's CurrentSelection.

Also, another reason for this error could be that Sync Lab and Picture Slides Lab opens a PPT in the background (though this is not confirmed)

Thus this PR:

1. Greys out / disable menu and buttons when there is no presentation open.
2. Will catch exception for PowerPointCurrentPresentationInfo.CurrentSelection and return null if there is no active window / problems with getting Selection.
3. Adds null checks when PowerPointCurrentPresentationInfo.CurrentSelection.
4. Refactor OpenInBackground to use method Open instead with focus set to false so the original working window will be activated (and will still remain as the current active window). The refactoring of OpenInBackground will affect PictureSlidesLab and SyncLab.


PowerPointLabs/PowerPointLabs/Models/PowerPointCurrentPresentationInfo.cs
```diff
@@ -29,7 +29,16 @@ public static Selection CurrentSelection
         {
             get
             {
-                return Globals.ThisAddIn.Application.ActiveWindow.Selection;
+                Selection select = GetSelectionInActiveWindow();
+                if (select == null)
+                {
+                    // Could be that there is no active window, we can try to activate one
+                    if (ActivateWindow())
```
PPTLabsBot: MAJOR Merge this if statement with the enclosing one. [rule](https://sonarcloud.io/coding_rules#rule_key=csharpsquid%3AS1066)

PowerPointLabs/PowerPointLabs/Models/PowerPointCurrentPresentationInfo.cs
```diff
@@ -29,7 +29,13 @@ public static Selection CurrentSelection
         {
             get
             {
-                return Globals.ThisAddIn.Application.ActiveWindow.Selection;
+                Selection select = GetSelectionInActiveWindow();
+                // Could be that there is no active window, we can try to activate one
+                if (select == null && ActivateWindow())
+                {
+                    select = GetSelectionInActiveWindow();
```
junkiattan: For the case where Ln 32 returns a null for select, I don't think that running GetSelectionInActiveWindow in Ln 36 would change that situation again.

PowerPointLabs/PowerPointLabs/Models/PowerPointCurrentPresentationInfo.cs
```diff
@@ -57,5 +63,35 @@ public static IEnumerable<PowerPointSlide> SelectedSlides
                 return slides;
             }
         }
+
+        private static bool ActivateWindow()
```
junkiattan: I assume that this method checks if there is an active window right? Perhaps the name can be changed to IsWindowActive or HasActiveWindow to make the naming more clear.

leeyh20: It tries to activate the first window at Line 74. Thus it does what the function says, which is to try to activate a window. However, I am unable to confirm that this will work because it is difficult to replicate the situation that Prof is facing.

PowerPointLabs/PowerPointLabs/Models/PowerPointCurrentPresentationInfo.cs
```diff
@@ -57,5 +63,35 @@ public static IEnumerable<PowerPointSlide> SelectedSlides
                 return slides;
             }
         }
+
+        private static bool ActivateWindow() 
+        {
+            try 
+            {
+                Globals.ThisAddIn.Application.Activate();
+                if (Globals.ThisAddIn.Application.Active == Microsoft.Office.Core.MsoTriState.msoTrue && Globals.ThisAddIn.Application.Windows.Count >= 1)
```
junkiattan: In Ln 71 the application has been activated, hence it would always be true in Ln 72 for the condition of

`Globals.ThisAddIn.Application.Active == Microsoft.Office.Core.MsoTriState.msoTrue`

PowerPointLabs/PowerPointLabs/Models/PowerPointCurrentPresentationInfo.cs
leeyh20: Situation will be changed if Line 74 of activating a window works. However, I think i will change this PR to solve a more general problem too when there are no windows available at all (Window count is 0 and there is no active window) so that we can actually test this.

PowerPointLabs/PowerPointLabs/Models/PowerPointCurrentPresentationInfo.cs
leeyh20: I’m afraid for some reason the API does not work and the application does not become active. But on second thought, if the API does not work it’ll probably throw a COMexception and it will be caught. I’ll check that.

leeyh20:
Okay I have checked https://msdn.microsoft.com/en-us/vba/powerpoint-vba/articles/documentwindow-object-powerpoint and it says 
> The first member of the DocumentWindows collection, Windows(1), always returns the active document window.

So indeed it's still mysterious why there will be no active window. I will try the method from https://support.microsoft.com/en-hk/help/285472/run-time-error--2147188160-on-activewindow-or-activepresentation-call and also try to refactor the EnabledHandler such that we do not need to get the selection from active window.

leeyh20:
@junkiattan Hmm ok I'll refactor it to really check for an active window then. Thanks!

damithc:
Not sure if it is related. I have a vague memory of some of our labs e.g., shapesLab maintaining another open powepoint file in the background (hidden from the user).

leeyh20:
That’s true... It also could be that the background file wanted to be the active window instead but then it is hidden. Then we need to investigate how to ensure how we can maintain the actual ppt as the active window. Thanks Prof for the idea!

PowerPointLabs/PowerPointLabs/Models/PowerPointCurrentPresentationInfo.cs
junkiattan: Must we forcibly activate a window? Or can we just return null?

The exception happens because we try to get selection from an active window when there is no active window. So what I am saying is that perhaps if there is no active window, instead of getting its selection and catching possible exceptions, just check for the window itself and return null if there is no active window.

leeyh20:
Have updated the PR description to reflect the latest commits.

leeyh20:
image
Screenshot of the ribbon being greyed out when there is no open presentation.

leeyh20:
Merged with latest dev release.

> An exception of type 'System.Runtime.InteropServices.COMException' occurred in PowerPointLabs.dll but was not handled in user code
This happens when a new template slide is previewed.
image

This error is related to the no active document window error. This is caused by `PowerPoint.DocumentWindow activeWindow = pres.Application.ActiveWindow;` in the `ThisAddInPrensentationOpen` function. 

I have fixed this error in the latest commit.

PowerPointLabs/PowerPointLabs/ThisAddIn.cs
```diff
@@ -982,19 +982,23 @@ private void ThisAddInAfterNewPresentation(PowerPoint.Presentation pres)
...
-        private void ThisAddInPrensentationOpen(PowerPoint.Presentation pres)
+        private void ThisAddInPresentationOpen(PowerPoint.Presentation pres)
         {
-            PowerPoint.DocumentWindow activeWindow = pres.Application.ActiveWindow;
-            string tempName = pres.Name.GetHashCode().ToString(CultureInfo.InvariantCulture);
-
-            // if we opened a new window, register the window with its name
-            if (!_documentHashcodeMapper.ContainsKey(activeWindow))
+            // Windows count could be zero if presentation is opened as preview of template slides
+            if (pres.Application.Windows.Count > 0)
```
leeyh20: Only this line was added.

leeyh20:
@junkiattan Sorry, you might need to do a full review again because I have made many changes. Thanks!

PowerPointLabs/PowerPointLabs/TextCollection/ColorsLabText.cs
```diff
@@ -2,13 +2,16 @@
 {
     internal static class ColorsLabText
     {
-        public const string PaneTag = "ColorsLab";
-
-        public const string TaskPanelTitle = "Colors Lab";
+        #region Action Framework Tags
+        public const string RibbonMenuId = "ColorsLabButton";
```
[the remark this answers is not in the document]
leeyh20: It has to be the same name as the id strings in the XML file (Ribbon1.xml) and for these labs, the menu buttons are called buttons. I'm not sure if I should change how these labs work because they are legacy labs that are quite old.

PowerPointLabs/PowerPointLabs/Models/PowerPointPresentation.cs
```diff
@@ -415,26 +415,6 @@ public bool Create(bool withWidow, bool focus)
             return true;
         }
 
-        public bool CreateInBackground()
```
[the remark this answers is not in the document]
leeyh20: It is very similar to Create(), almost a copy-and-paste, except the fact that Create() allows for the working window to regain focus and this is exactly what I wanted to add to the CreateInBackground() function. Then I realised Create() already had it, thus I wanted to reduce code duplication.

PowerPointLabs/PowerPointLabs/Models/PowerPointPresentation.cs
```diff
@@ -477,32 +457,6 @@ public virtual bool Open(bool readOnly = false, bool untitled = false, bool with
             return true;
         }
 
-        public virtual bool OpenInBackground()
```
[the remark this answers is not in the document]
leeyh20: Similarly I wanted the original window to regain focus and Open() is exactly the same as this OpenInBackground() method except for the fact that Open() allows the original window to regain focus.

PowerPointLabs/PowerPointLabs/ThisAddIn.cs
```diff
@@ -992,15 +992,22 @@ private void ThisAddInAfterNewPresentation(PowerPoint.Presentation pres)
...
+                PowerPoint.DocumentWindow activeWindow = pres.Application.ActiveWindow;
+                string tempName = pres.Name.GetHashCode().ToString(CultureInfo.InvariantCulture);
+
+                // if we opened a new window, register the window with its name
+                if (!_documentHashcodeMapper.ContainsKey(activeWindow))
+                {
+                    _documentHashcodeMapper[activeWindow] = tempName;
+                }
+
+                // Refresh ribbon to grey out the menu / buttons if there are now at least one window
```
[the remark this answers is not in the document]
leeyh20: I'm not sure if an actual presentation can start out with no active window but okay I think this can be done. Shall test it.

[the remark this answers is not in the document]
leeyh20: Oh I did not want to enable the menu buttons if there is no active window so as to prevent the "no currently active document window" errors. Most of our PPTLabs features rely on an active window.

leeyh20:
@junkiattan Have addressed the changes. Could you help me test on PPT 2013 again? Thanks!

junkiattan:
I have tested on PPT2013 and found one error.

1. After I open PPT, PPL tabs would be active as expected.
1. Proceed to close the presentation, PPL tabs are disabled as expected.
1. Then proceed to open an existing presentation or create a new presentation, and the PPL tabs remain disabled (should be enabled)

Tested also on PPT2016. The same error happens, but the step 3 differs in that you can opem an existing presentation but the tabs remain disabled with a new presentation.

junkiattan:
Latest changes solved bug and runs fine on PPT2013/PPT2016 for both opening new/existing presentations

leeyh20:
@initialshl I have updated the comment. 
Issues confirmed to be fixed: #1615
Issues supposedly fixed: #1608 and #1543

Issue #1601, opened 2018-02-08
Error in Positions Lab tutorial


alexfjw:
> The Horizontal Distribution tutorial didn’t work correctly. It created duplicates as opposed to distributing by the order given in the tutorial.

Bug report from a user

alexfjw:
@junkiattan Think this belongs in your department ;)

leeyh20:
There's another bug report that was just submitted with the exact same problem.

> 
> I followed the tutorial and found some errors in Positions Lab. Objects are being duplicated. This occurs mainly in the Aling Horizontal and Distribute Horizontal buttons.
> 
> I am using Windows 10 x64 and Office Professional Plus 2016 (both in Portuguese - Brazil).

junkiattan:
@alexfjw 

I have checked on 2013 and 2016 versions. There does not seem to be an error in the Align operations, only in the Distribute Horizontal buttons.

Will look into it

damithc:
This one has similar problems (PPT2013)
![image](https://user-images.githubusercontent.com/1673303/36463164-f91b6442-1702-11e8-9f80-aa300730b093.png)

PR #1614, opened 2018-02-19
Error in Positions Lab tutorial #1601

junkiattan:
#1601 

The error possibly comes from the corrupted pictures used for the relevant slide.

I isolated the problem to the line in ExecutePositionAction where the code deletes the simulatedShapes. At first, I tested and realized that the code worked for shapes but not pictures (since pictures was used in the slides). However after going online and using images from the browser of the same type, the code runs correctly.

I have replaced the corrupted pictures with the same format and colour and tested it for both PPT2013 and PPT2016 and the function now works.


leeyh20:
Oh why was this closed?

PR #1617, opened 2018-02-22
#1544-Copy-Operation-Made-The-Text-Disappear

### LINKED ISSUE #1544: Copy operation made the text disappear
PPT 2013, Win 10, PL Version 4.1.1.2
Steps:
1. Select shape given below and click on the `copy` button
2. The text in the box disappeared (possibly changed to black color?)
3. After completing the copy operation,  I was able to get the text back with one `Ctrl+Z` without losing the copied styles

image

alexfjw:
fixes #1544 

Summary of issue: 
MS returns black when shape's text has more than one color.
This turns text black when we try to sync.

There was also an error in the code. 
We try to check if syncing of text color is possible, by syncing the shape text's color with itself. This sets the color to black.

Changes in code: 
- Use a more sensible sync check.
- Guess the color by clear the texting so only one color remains & checking the color afterwards.

### DIFF
--- a/PowerPointLabs/PowerPointLabs/SyncLab/ObjectFormats/FontColorFormat.cs
+++ b/PowerPointLabs/PowerPointLabs/SyncLab/ObjectFormats/FontColorFormat.cs
@@ -1,9 +1,10 @@
 ﻿using System;
 using System.Drawing;
-
-using Microsoft.Office.Interop.PowerPoint;
+using Microsoft.Office.Core;
 using PowerPointLabs.ActionFramework.Common.Log;
 using PowerPointLabs.Utils;
+using Shape = Microsoft.Office.Interop.PowerPoint.Shape;
+using Shapes = Microsoft.Office.Interop.PowerPoint.Shapes;
 
 namespace PowerPointLabs.SyncLab.ObjectFormats
 {
@@ -11,7 +12,7 @@ namespace PowerPointLabs.SyncLab.ObjectFormats
     {
         public static bool CanCopy(Shape formatShape)
         {
-            return Sync(formatShape, formatShape);
+            return formatShape.HasTextFrame == MsoTriState.msoTrue;
         }
 
         public static void SyncFormat(Shape formatShape, Shape newShape)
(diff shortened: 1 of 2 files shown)


alexfjw:
Works fine on 13 & 16

PowerPointLabs/PowerPointLabs/SyncLab/ObjectFormats/FontColorFormat.cs
```diff
@@ -1,17 +1,18 @@
...
 using PowerPointLabs.ActionFramework.Common.Log;
 using PowerPointLabs.Utils;
+using Shape = Microsoft.Office.Interop.PowerPoint.Shape;
+using Shapes = Microsoft.Office.Interop.PowerPoint.Shapes;
 
 namespace PowerPointLabs.SyncLab.ObjectFormats
 {
     class FontColorFormat
     {
         public static bool CanCopy(Shape formatShape)
```
leeyh20: Doesn't seem like this function is used anywhere..?

leeyh20:
LGTM. SyncLab test works on PPT 2010 and manual testing works too.

PowerPointLabs/PowerPointLabs/Utils/ShapeUtil.cs
```diff
@@ -984,6 +984,28 @@ public static TextRange ConvertTextRange2ToTextRange(TextRange2 textRange2)
             return shape == null ? null : shape.TextFrame.TextRange;
         }
 
+        /// <summary>
+        /// Useful when shape has text with more than one color
+        /// Getting color in that case without a workabout will return black
+        /// Assumes that shape has a TextRange
+        /// 
+        /// Unsuccessful workabouts:
```
leeyh20: Not sure if the unsuccessful workabouts should be mentioned in summary or not (put as comments?) since it makes the summary very long and as you hover on the method, it extends too long across the screen.

alexfjw: Good point. I'll shift it into the function

PowerPointLabs/PowerPointLabs/SyncLab/ObjectFormats/FontColorFormat.cs
alexfjw: Format object methods are accessed with reflection so the ide doesn't pick it up. See Format.cs's CanCopy method. Not sure if we will be refactoring this some time down the line to an oop approach. Either that or some comments in a README

Issue #1581, opened 2018-01-31
SyncLab: Copying SmartArt gives an error


alexfjw:
Accessing shape.Name throws an error for SmartArt, as it is not defined internally.
The current code does this when creating the SyncFormatDialog.
We should exercise more care when handling shape names.

alexfjw:
Synclab requires duplicating of Shapes.
SmartArt shapes cannot be duplicated.
We need some sort of workabout for this... no ideas so far though.

alexfjw:
#1581 prevents syncing of SmartArt altogether. Do note that the above 2 issues are not yet resolved in the code.

PR #1620, opened 2018-02-25
#1581-SyncLab-Copying-SmartArt-gives-an-error

alexfjw:
Temporary fix for #1581.
Show a message when copying SmartArt, to mention that SmartArt is not supported by SyncLab.
This prevents an error that occurs when the system attempts to copy SmartArt.


leeyh20:
LGTM! Tested on PPT 2010.

alexfjw:
This won't be fixed any time soon I believe. Requires undiscovered workabouts. Thanks, I'll look into adding the FT.

alexfjw:
Have added an FT for the message dialog. 
Also did some cleaning up for existing message dialog checks. (Hardcoded strings, unneeded imports)

initialshl:
Can we rerun all tests for all ppt versions before i approve this pr?

alexfjw:
Sure. Only test related files changed from that point but I'll just rerun the tests to be safe.

junkiattan:
Have ran the test on PPT2013 and it works.

leeyh20:
Runs on PPT2010 and it works.

PR #1627, opened 2018-03-08
#1626 Extend restore clipboard for actions that return a value

### LINKED ISSUE #1626: RestoreClipboardAfterAction should be able to return a value
As requested by @junkiattan, RestoreClipboardAfterAction should be able return a value so that we can use this with functions that return a value after modifying the clipboard in some way.

leeyh20:
Fixes #1626 
This PR also fixes flakiness of clipboard test.
Tested on PPT 2016.

@junkiattan Can help to test?


junkiattan:
@leeyh20 Looks good for PPT2013

PowerPointLabs/Test/FunctionalTest/ConvertToPictureTest.cs
```diff
@@ -66,12 +66,18 @@ private static void CheckIfClipboardIsRestored()
 
             // Paste whatever in clipboard
             Microsoft.Office.Interop.PowerPoint.ShapeRange newShape = actualSlide.Shapes.Paste();
+            // Set the shape location to 0 because the location of the shape being pasted is flaky
+            newShape.Top = 0;
```
initialshl: I think its better to set newshape position to copied position, and justify this "cheating" with the flakiness comment. Otherwise in case this method is used multiple times on a single slide, would be hard to differentiate the shapes

leeyh20: Ok :)

leeyh20: Good point! I have edited the test.

initialshl:
LGTM

PR #1583, opened 2018-01-31
#1403 Suggested new feature: Paste To Fit Slide

junkiattan:
#1403 

I have implemented the PasteToFit slide for PasteLab. I have also included the PasteToFit function for the contextual menu (right click)

In addition I am planning to also include FTs for this issue.

Picture of PasteToFit in contextual menu
pastetofit1

Picture of PasteToFit in PowerPointLabs Tab
pastetofit2

Original picture
pastetofit3

Paste to Fill for pic
pastetofit4

Paste to Fit for pic
pastetofit5

I have ran and passed all relevant FTs and UTs and updated all references in ThisAddIn.cs and Ribbon1.xml and other relevant files.

Note: I have not yet changed the icon image for PasteToFit. Currently it is the same as PasteToFill. Wanted to change the png from within VS but decided to consult you first @initialshl

Issue #1403, opened 2017-06-01
Suggested new feature: Paste To Fit Slide


initialshl:
As discussed today (for a future version):

A feature of Paste Lab. 
Paste the clipboard so that it fits the slide, similar to Paste To Fill Slide without the cropping.

junkiattan:
@initialshl ,

when you mention without cropping, do you mean it is allow to exceed the size of the slide? or will it fill up as best as it can wile maintaining aspect ratio without exceeding the boundaries of the slide?

Note: for the latter case, I think the results would be more obvious for images orginally smaller then the slide

initialshl:
It should be the latter

PR #1628, opened 2018-03-09
#1403 Reinsert paste to fit button

junkiattan:
A continuation from #1583 

Apparently a subsequent merged issue overrode changes to the Ribbon1.xml file for the original solution for #1403, which removed the PasteToFitSlide button from menu tab.

This fix simply adds the button in.


leeyh20:
Actually I think it might conflict with my #1608 because I am also changing the Ribbon.xml there with substantial changes...

Issue #1213, opened 2017-03-16
Sync Lab: Add support for Shape Effects


IanTeo:
Effects
- [x] Add support for Shadow
- [x] Add support for Reflection
- [x] Add support for Glow
- [x] Add support for Soft Edges
- [x] Add support for Bevel
- [x] Add support for 3D-rotation

alexfjw:
Mostly completed, but there are some labs that can be split into smaller components
Respective issues have been created for those labs.

PR #1642, opened 2018-03-21
#1623 replace with clipboard: preserve shape settings (existing sync formats only)

### LINKED ISSUE #1623: Replace with clipboard: preserve shape settings
When a picture is replaced with another using 'Replace with clipboard' feature, the new pictures seems to lose some of the properties of the previous one, such as the shadow effects. It would be good to preserve as many effects as possible.

alexfjw:
resolves #1623

- [x] preserve settings for shapes (fill, color, text, whatever synclab already has, excluding size), single shape
- [x] as above, for grouped shape
- [x] <s>remove redundant existing code in Replace With Clipboard, if any</s> (no redundant code found, the position code present is required to correctly shift the new image)

We currently lack code for shadows, bezels etc. See #1213 

Test Cases:
1) ensure that replacing a shape preserves all settings other than size (test replacing with both a single shape & > 1 shape in clipboard)
2) ensure that replacing a shape inside a group preserves all settings other than size (test replacing with both a single shape & > 1 shape in clipboard)

### DIFF
--- a/PowerPointLabs/PowerPointLabs/PasteLab/ReplaceWithClipboard.cs
+++ b/PowerPointLabs/PowerPointLabs/PasteLab/ReplaceWithClipboard.cs
@@ -29,6 +29,7 @@ namespace PowerPointLabs.PasteLab
 
                 slide.DeleteShapeAnimations(pastingShape);
                 slide.TransferAnimation(selectedShape, pastingShape);
+                ShapeUtil.ApplyAllPossibleFormats(selectedShape, pastingShape);
                 selectedShape.Delete();
 
                 return slide.ToShapeRange(pastingShape);
@@ -63,6 +64,7 @@ namespace PowerPointLabs.PasteLab
                         shapeAbove = shape;
                     }
                 }
+                ShapeUtil.ApplyAllPossibleFormats(selectedChildShape, shapeAbove);
 
                 // Remove selected child since it is being replaced
                 ShapeRange shapesToGroup = slide.ToShapeRange(selectedGroupShapeList);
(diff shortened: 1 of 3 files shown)


junkiattan:
Tested on PPT2016

Syncing works for
- Fill
- Text size / font
- Colour
- Line

Syncing do not yet work for 
- 3D formats
- Size

Currently I cannot select 2 shapes concurrently and use the "Replace with clipboard" function. Is that the intended behavior?

alexfjw:
>Syncing do not yet work for 3D formats
Picture effects are in a pending PR and not in dev-release yet.

>  Size
I didn't want to sync size as it may make things go out of shape, especially pictures and squares, circles. Apologies, the discussion was in the issue and not this PR. I think Prof Damith already made an issue for syncing shape too, suggesting adding another button / option #1656.
     
> Currently I cannot select 2 shapes concurrently and use the "Replace with clipboard" function. Is that the intended behavior?
This is the intended behavior of the function before this PR was introduced

junkiattan:
Tested on PPT2016 and syncing with multiple shapes in the clipboard works.

However PPT2013 had some issues. Issues only happen with >1 shapes in clipboard

Shapes that are cut/copied
capture

Desired shape to be replaced
capture2

Result
capture3

When I try to replace a shape in a group, it replaces the whole group, ending up with the same result as above.

junkiattan:
@leeyh20 Will need you to test on as many versions of ppt as you can. The issue I raised could be a computer dependent issue

alexfjw:
> However PPT2013 had some issues. Issues only happen with >1 shapes in clipboard

I can't seem to replicate this issue. Maybe it's an installation dependent issue again.
Let's see if @leeyh20 faces this on 2010.

leeyh20:
In my PPT 2010... it does not seem to work.
image

You also might want to merge with latest dev release first and check if everything is working because we have changed paste lab significantly with pictures hmm.

junkiattan:
@leeyh20 Is it possible to provide a better error description? Is the error you are facing the same as mine?

leeyh20:
The syncing does not seem to happen at all.

Issue #1644, opened 2018-03-22
SyncLab: Artistic Effect Fails to Sync, 2013


alexfjw:
We cannot sync artistic effect of pictures on 2013.
This problem does not occur on 2016.
I'm guessing this is due to the copying / saving mechanism for SyncLab. 
It could be that the copied shape has it's artistic effect set to permanent in 2013.

We will probably have to look into different ways of saving shapes in SyncLab to work around this.

alexfjw:
Have thought of a fix for this, will put it in after #1645 is merged.

Approach:
Sync artistic effect after SyncLab's storage mechanism copies the shape.
We will be transferring the artistic effect direct to the saved shape. 
This way, it will not be missed.

Note to avoid edge cases:
Manually clear the artistic effect of the copied object before syncing.
Add this as a function in PictureEffectsFormat.cs

leeyh20:
That's weird, it seems to work on PPT 2010... why does it happen in 2013?

alexfjw:
Exactly. Its fine on 2016 as well. No idea what's different in 2013

PR #1645, opened 2018-03-22
#1618 synclab refactor format objects

### LINKED ISSUE #1618: SyncLab: Refactor Format objects
Would be a good idea to refactor the format objects to OOP rather than sticking to the current util classes + reflection style.
The code has been giving quite a bit of confusion thus far for code reviews.
Makes sense to do this before adding more SyncLab formats like #1213

alexfjw:
Fixes #1618 

Format objects now follow OOP, inheriting from an abstract Format class rather than act as static utility classes. This gives us type safety!


alexfjw:
Also fixed a failing UT for copying placeholder pictures. 
Not sure why the failing UT missed our radar until now.

alexfjw:
FTs & UTs work on 2016. 
(Except the sync artistic effect UT, which seems to be something specific to my system, tested to be manually working on 2016)

Note that sync artistic effect does not work on 2013. See #1644

junkiattan:
Test works for PPT2013 and PPT2016

leeyh20:
Yes the failing UT is "TestSetArtisticEffect".
Besides that, the other tests passed in PPT 2010.

alexfjw:
For anyone giving a code review, here's a summary of the changes:
1) Format is an abstract class. All format classes now extend Format. 
2) Format also has it's equals & hashcode method modified. Format objects of the same class should be considered equal.
3) Fixing the failing placeholder picture test:
 PowerPointLabs/Test/UnitTest/PlaceHolderCopyTest.cs
PlaceHolder pictures were not copyable in the past. They are now and the test has been amended to show that.

alexfjw:
@Skaty @initialshl @IanTeo 
Ready for a review, the added commit only includes only changes from the csproj file

PR #1648, opened 2018-03-23
#1606 Fix pasting of web pictures

leeyh20:
Resolves #1606 which will allow web pictures to be pasted with all the Paste Lab functions without errors being thrown.

Side effect: This also resolves #1380 because it will paste slide objects as a picture instead.

Also changes RestoreClipboardAfterAction for web pictures in clipboard because it will now try to paste as a PNG picture, then bitmap picture and then try View.Paste(). The benefits of trying to paste as PNG picture is that the transparency of the picture will not be lost in case it is a PNG picture.

To test web pictures with RestoreClipboardAfterAction, use ConvertToPicture with web pictures in the clipboard.

To Test: 
1. Paste Lab functions with web pictures in clipboard (can try PNG pictures)
2. Paste Lab functions with slide objects in clipboard
3. ConvertToPicture with web pictures in clipboard (can try PNG pictures)

There is a weird powerpoint bug where sometimes it cannot detect that the web picture it is pasting is a PNG picture until I try to paste it using the clipboard pane first. This seems to be a problem with powerpoint itself because even the default paste / paste special does not allow me to paste the web picture as PNG. Thus in this case, Paste Lab will just have to paste it as bitmap picture (without transparency) :(

PR #1653, opened 2018-03-26
Positions Lab Bug (Distribute) #1640 

junkiattan:
#1640 

Current fix for distribute functions:

Distribute center, horizontal and vertical now works for both types of images. 

However Distribute radial and grid does not work, even after manual copy/paste of the images. This means for radial/grid, we cannot use copy/paste or the solution used to solve for distribute center, horizontal and vertical.

Furthermore, it has to be considered that all the methods work well for most images, hence we should minimize changes to the code that can affect this core functionality.

The basic workaround for radial/grid is to return an error message that indicates to the user some form of corrupted images.

Methods tried:
1. Corruption correction - failed because of the same error that caused the bug (i.e. `simulatedShapes.Delete()`)
1. Manually copy/paste and then using `ClipboardUtil` methods to replicate that in code - failed because it does not work for the Distribute Radial/Grid
1. Default powerpoint methods like `Shapes.Duplicate()` - failed for Distribute Radial /Grid as desired outcome of image arrangement is not achieved


PowerPointLabs/PowerPointLabs/PositionsLab/PositionsPaneWPF.xaml.cs
```diff
@@ -1353,114 +1360,122 @@ public void ExecutePositionsAction(Action<List<PPShape>, bool> positionsAction,
...
+                    if (simulatedShapes != null)
+                    {
+                        simulatedShapes.Delete();
+                        GC.Collect();
+                    }
                 }
             }
         }
 
         public void ExecutePositionsAction(Action<List<PPShape>, float, float> positionsAction, float dimension1, float dimension2, bool isPreview)
```
PPTLabsBot: CRITICAL Refactor this method to reduce its Cognitive Complexity from 18 to the 15 allowed. [rule](https://sonarcloud.io/coding_rules#rule_key=csharpsquid%3AS3776)

junkiattan: Ignoring this as it involves simplifying the method further which is not possible without changing the basic functionality of the current method

junkiattan:
@alexfjw 

Note for future reference:
I have removed duplicated shapes as best as I could by deleting the shapes manually one by one (checking if each of them has the "_Copy" string) under exception handling for `simulatedShapes.Delete()`

However this method would be insufficient in removing all duplicated shapes in the case where user tries to distribute a mixture of good/bad shapes. The workaround adopted would be to return an error message asking user to undo and then try the operation again with a different set of shapes.

Tested on PPT2013/2016 and it works as expected. Also removed the unneeded wrapper method. Please review one last time and approve if good to go.

junkiattan:
@alexfjw 

Actually we would need to show the dialogue even when items deleted == number of simulated shapes. This is because once you reach that line, it would mean you are dealing with corrupted shapes already, so either way you would have to show the dialog message.

alexfjw:
Good point. Thanks for clarifying

leeyh20:
Not sure if it helps, but my #1648 faced some error with shape range's get enumerator where some weird corrupted shape get enumerator would throw an error. Then I managed to delete it using shapes[1].Delete(); where shapes[1] is the shape range. I did this because I realised shape range has count of 2 which is highly weird cause I was very sure there was only one shape. Then when I deleted using shapes[1].Delete(); the shape really just got deleted.....

leeyh20:
In PPT 2010, after showing the error that these shapes are corrupted, the shapes are still being duplicated
image

If I select only three shapes and press Radial, it will say "Please select four shapes" error and then still show the "these shapes are corrupted" error. Is this intended? Should we just show one error?

PowerPointLabs/PowerPointLabs/PositionsLab/PositionsPaneWPF.xaml.cs
```diff
@@ -1099,8 +1099,34 @@ public void ExecutePositionsAction(Action<ShapeRange> positionsAction, bool isPr
...
+                                this.GetCurrentSlide().Shapes[1].Delete();
+                            }
+                        }
+                        // Remove any outlier extra shapes not deleted previously
+                        try
+                        {
+                            this.GetCurrentSelection().Delete();
+                        }
+                        catch (System.Runtime.InteropServices.COMException)
+                        {
```
PPTLabsBot: MAJOR Either remove or fill this block of code. [rule](https://sonarcloud.io/coding_rules#rule_key=csharpsquid%3AS108)

PowerPointLabs/PowerPointLabs/PositionsLab/PositionsPaneWPF.xaml.cs
```diff
@@ -1531,8 +1567,34 @@ private void ExecuteDistributeGrid(int rowLength, int colLength)
...
```
PPTLabsBot: MAJOR Either remove or fill this block of code. [rule](https://sonarcloud.io/coding_rules#rule_key=csharpsquid%3AS108)

PPTLabsBot: MAJOR Either remove or fill this block of code. [rule](https://sonarcloud.io/coding_rules#rule_key=csharpsquid%3AS108)

junkiattan:
@alexfjw @leeyh20 

I have included an additional check to remove duplicated shapes in the case when Distribute is called for a mixture of good and bad shapes.

I have also improved on the error message to make it more explicit.

I have tested on 2016/2013. However because of the scope of this bug, would be glad if you could double check on 2010/2013/2016 if possible.

leeyh20:
On PPT 2010, everything works but when I do Radial with four shapes for more than 4 or 5 times on the same shape, I start to get this error:

```
System.Runtime.InteropServices.COMException
Exception from HRESULT: 0x800A01A8
Stack Trace:
   at Microsoft.Office.Interop.PowerPoint.Shape.Duplicate()
   at PowerPointLabs.PositionsLab.PositionsPaneWpf.DuplicateShapes(ShapeRange range) in D:\Git Projects\PPLABS_JUNKIAT\PowerPointLabs\PowerPointLabs\PositionsLab\PositionsPaneWPF.xaml.cs:line 1726
   at PowerPointLabs.PositionsLab.PositionsPaneWpf.ExecutePositionsAction(Action`1 positionsAction, Boolean isPreview, Boolean isConvertPPShape) in D:\Git Projects\PPLABS_JUNKIAT\PowerPointLabs\PowerPointLabs\PositionsLab\PositionsPaneWPF.xaml.cs:line 1065

```

But I really have to do Radial a lot of times to get this.

PowerPointLabs/PowerPointLabs/PositionsLab/PositionsPaneWPF.xaml.cs
```diff
@@ -1099,8 +1099,42 @@ public void ExecutePositionsAction(Action<ShapeRange> positionsAction, bool isPr
...
+                        Only triggered for cases where Distribute is called for cases consisting of both
+                        corrupted and non-corrupted shapes
+                        */
+                        try
+                        {
+                            this.GetCurrentSelection().Delete();
+                        }
+                        catch (System.Runtime.InteropServices.COMException)
+                        {
+                            /*
```
alexfjw: Seems like the codestyle we follow recommend using /// or // rather than /* */. 
https://github.com/oss-generic/process/blob/master/codingStandards/CodingStandard-CSharp.adoc

PowerPointLabs/PowerPointLabs/PositionsLab/PositionsPaneWPF.xaml.cs
```diff
@@ -1531,8 +1575,42 @@ private void ExecuteDistributeGrid(int rowLength, int colLength)
...
+                        // Remove all simulated shapes manually
+                        for (int i = 0; i < simulatedShapes.Count; i++)
+                        {
+                            // This method to remove duplicated shapes might fail for non-corrupted shapes when mixing good/bad shapes
+                            if (this.GetCurrentSlide().Shapes[1].Name.Contains("_Copy"))
+                            {
+                                this.GetCurrentSlide().Shapes[1].Delete();
+                            }
+                        }
+                        /*
```
alexfjw: Seems like the codestyle we follow recommend using /// or // rather than /* */.
https://github.com/oss-generic/process/blob/master/codingStandards/CodingStandard-CSharp.adoc

PowerPointLabs/PowerPointLabs/PositionsLab/PositionsPaneWPF.xaml.cs
```diff
@@ -1099,8 +1099,42 @@ public void ExecutePositionsAction(Action<ShapeRange> positionsAction, bool isPr
...
```
alexfjw: Seems like the codestyle we follow recommend using /// or // rather than /* */.
https://github.com/oss-generic/process/blob/master/codingStandards/CodingStandard-CSharp.adoc

leeyh20:
@junkiattan Hmm i reselect the 4 or more shapes a lot of times and run Radial, and two of them were corrupted, two were normal autoshapes made with PPT.

leeyh20:
Okay it's only the PPT 2010 on my computer, should be a computer-dependent thing.

PowerPointLabs/PowerPointLabs/PositionsLab/PositionsPaneWPF.xaml.cs
IanTeo: Do you guys have the permission to edit SonarQube? It may be beneficial to decide which rules should be ignored in our project to avoid such problems, especially for future developers

junkiattan: I think we don't need to edit the SonarQube, not that we were given permissions to. This is because the problem in code quality is isolated for only certain functions in PositionsLab, which was code by previous few batches of students. It would be better to maintain the same rules now to maintain the same strictness in code standard in the overall project, and decide which to ignore on a case-by-case basis

IanTeo: As long as the current rules in place match with the current code quality. You are right that we can ignore it on a case by case basis for legacy codes.

PowerPointLabs/PowerPointLabs/PositionsLab/PositionsPaneWPF.xaml.cs
```diff
@@ -1353,114 +1383,124 @@ public void ExecutePositionsAction(Action<List<PPShape>, bool> positionsAction,
...
-            {
-                // invalid selection!
-                return;
-            }
-
-            ShapeRange simulatedShapes = null;
-
-            try
+            // Run twice to get the right operation
+            for (int noOfRuns = 0; noOfRuns < 2; noOfRuns++)
```
IanTeo: I don't really understand why running twice will get the right operation. Could you explain it to me?

junkiattan: @IanTeo its a PowerPoint thing. I determined it via manual testing and made sure it worked for different shapes/pictures. But ultimately I do not know why exactly running twice is needed.

PowerPointLabs/PowerPointLabs/PositionsLab/PositionsPaneWPF.xaml.cs
```diff
@@ -1353,114 +1383,126 @@ public void ExecutePositionsAction(Action<List<PPShape>, bool> positionsAction,
...
         public void ExecutePositionsAction(Action<List<PPShape>, float, float> positionsAction, float dimension1, float dimension2, bool isPreview)
         {
-            if (this.GetCurrentSelection().Type != PpSelectionType.ppSelectionShapes)
+            // Need to run the action 2 times because of the nature of PowerPoint default operations
+            // This has been determined via manual testing
+            for (int noOfRuns = 0; noOfRuns < 2; noOfRuns++)
             {
-                if (!isPreview)
+                if (this.GetCurrentSelection().Type != PpSelectionType.ppSelectionShapes)
                 {
```
PPTLabsBot: CRITICAL Refactor this method to reduce its Cognitive Complexity from 18 to the 15 allowed. [rule](https://sonarcloud.io/coding_rules#rule_key=csharpsquid%3AS3776)

junkiattan: Ignored as it involved changing the basic functionality of this method

PowerPointLabs/PowerPointLabs/PositionsLab/PositionsPaneWPF.xaml.cs
```diff
@@ -1531,8 +1573,38 @@ private void ExecuteDistributeGrid(int rowLength, int colLength)
...
+                        GC.Collect();
+                    }
+                    // Catch corrupted shapes
+                    catch (System.Runtime.InteropServices.COMException)
+                    {
+                        // Remove all simulated shapes manually
+                        for (int i = 0; i < simulatedShapes.Count; i++)
+                        {
+                            // This method to remove duplicated shapes might fail for non-corrupted shapes when mixing good/bad shapes
+                            if (this.GetCurrentSlide().Shapes[1].Name.Contains("_Copy"))
```
initialshl: Will this unintentionally delete some shapes if its name has _Copy? (Even when it should not be affected by this operation)

PowerPointLabs/PowerPointLabs/PositionsLab/PositionsPaneWPF.xaml.cs
```diff
@@ -1353,114 +1383,126 @@ public void ExecutePositionsAction(Action<List<PPShape>, bool> positionsAction,
...
                 }
             }
         }
 
         public void ExecutePositionsAction(Action<List<PPShape>, float, float> positionsAction, float dimension1, float dimension2, bool isPreview)
         {
-            if (this.GetCurrentSelection().Type != PpSelectionType.ppSelectionShapes)
+            // Need to run the action 2 times because of the nature of PowerPoint default operations
+            // This has been determined via manual testing
+            for (int noOfRuns = 0; noOfRuns < 2; noOfRuns++)
```
initialshl: Recommend using numOf instead of noOf

PowerPointLabs/PowerPointLabs/PositionsLab/PositionsPaneWPF.xaml.cs
```diff
@@ -1099,8 +1099,38 @@ public void ExecutePositionsAction(Action<ShapeRange> positionsAction, bool isPr
...
+                        GC.Collect();
+                    }
+                    // Catch corrupted shapes
+                    catch (System.Runtime.InteropServices.COMException)
+                    {
+                        // Remove all simulated shapes manually
+                        for (int i = 0; i < simulatedShapes.Count; i++)
+                        {
+                            // This method to remove duplicated shapes might fail for non-corrupted shapes when mixing good/bad shapes
+                            if (this.GetCurrentSlide().Shapes[1].Name.Contains("_Copy"))
```
initialshl: Will this accidentally delete shapes that should not be affected by this operation?

initialshl: Is the logic correct? Thought it should check shapes[i] instead of [1] everytime

junkiattan: duplicated shapes are somehow appended to the starts of the Shapes collection, hence they are referred to by shapes[1]. If you index it with i instead, it may delete the nonduplicated shapes at the back. Think of this as popping from the stack.

Furthermore, it is not code practice to delete shapes by their indexes while in a for loop, because it will mess up the delete operations

junkiattan: It has been tested and did not delete other unintended shapes.

I have tried to make it work as best as i could. The operation works well for when Distribute is performed for a set of "good" shapes and a set of "bad" shapes, but may be flaky for when there is a mixture of good/bad shapes. I have tried to solve this but faced API limitations. 

Hence the error message (Like for Radial and Grid) to get the user to press undo to ensure that state of the slides is reverted back to original.

PowerPointLabs/PowerPointLabs/PositionsLab/PositionsPaneWPF.xaml.cs
junkiattan: noted on the change

PowerPointLabs/PowerPointLabs/PositionsLab/PositionsPaneWPF.xaml.cs
```diff
@@ -1353,114 +1383,126 @@ public void ExecutePositionsAction(Action<List<PPShape>, bool> positionsAction,
...
         public void ExecutePositionsAction(Action<List<PPShape>, float, float> positionsAction, float dimension1, float dimension2, bool isPreview)
         {
-            if (this.GetCurrentSelection().Type != PpSelectionType.ppSelectionShapes)
+            // Need to run the action 2 times because of the nature of PowerPoint default operations
+            // This has been determined via manual testing
+            for (int numOfRuns = 0; numOfRuns < 2; numOfRuns++)
             {
-                if (!isPreview)
+                if (this.GetCurrentSelection().Type != PpSelectionType.ppSelectionShapes)
                 {
```
PPTLabsBot: CRITICAL Refactor this method to reduce its Cognitive Complexity from 18 to the 15 allowed. [rule](https://sonarcloud.io/coding_rules#rule_key=csharpsquid%3AS3776)

PowerPointLabs/PowerPointLabs/PositionsLab/PositionsPaneWPF.xaml.cs
initialshl: Hm then I think it should break out of loop as soon as it does not contain "Copy"? Otherwise it seems to be checking the same shape multiple times

junkiattan: I have added the breaks for the for loops. Thanx for that spot!

PR #1655, opened 2018-03-27
#1213 support for syncing shadow

alexfjw:
As title, part of #1213 
- [x] write code
- [x] write tests

Extra notes: 

I initially hid the shape's shadow in the SyncLab pane's display.
Decided against that after further consideration.
Showing the exact shape + all effects has been how we've handled the display all this while. 

The shadow image in SyncLab's dialog has been emphasized by enforcing shadow transparency = 0.3. 
Tried messing with other values for the shape but didn't get something which works universally.


PowerPointLabs/PowerPointLabs/SyncLab/ObjectFormats/ShadowEffectFormat.cs
```diff
@@ -0,0 +1,51 @@
...
+            destFormat.RotateWithShape = srcFormat.RotateWithShape;
+            destFormat.Blur = srcFormat.Blur;
+            destFormat.OffsetX = srcFormat.OffsetX;
+            destFormat.OffsetY = srcFormat.OffsetY;
+            destFormat.Transparency = srcFormat.Transparency;
+        }
+
+        public override Bitmap DisplayImage(Shape formatShape)
+        {
+            // change transparency to make shadow more visible
```
leeyh20: This is to make it more visible on the SyncLab Pane?

leeyh20: Oh just saw your github comment. Hmm I'm not sure if you need to write it clearer in the comment that it is **only** on the SyncLab Pane.

leeyh20:
I found one last bug. (sorry!!)

image

Changing the colour of the shadow to orange and then syncing the shadow does not preserve the colour of the shadow. It goes back to black.

leeyh20:
@alexfjw  Is it just me but there's a build error because PasteToFillSlideActionHandler cannot be found?

alexfjw:
Could have been a mistake in rebasing. Thanks i'll check it out

leeyh20:
Works now on PPT 2010.

leeyh20:
Sorry after testing with mixedType again, I get this error.
image
Preset shadows work perfectly.

image

alexfjw:
​Thanks, it should be a PPT thing where the API returns an out of bounds
value rather than 0.
I'll check it out in a bit.

alexfjw:
Further investigation reveals that many settings are unique to the shadow preset chosen. It would be strange to sync a shape's shadow with incompatible elements to another shape's shadow. 

One example is the inner shadow preset. The inner shadow lacks a size setting. Allowing syncing of individual settings makes users think that we do allow syncing of previously incompatible settings across shadow types.

Another concern is the shadow offset. I believe this comes along with the shadow preset chosen. However, users are not able to tweak this value manually. Splitting up the shadow effect would mean we have to map the offset values to each preset, because we cannot retrieve the preset reliably (Type returns Mixed when any one setting is modified). I think this is a little too big for this PR at the moment.

I do think that shadow is an important visual effect however and works as a group with the other visual effects. We should try to bring out all visual effects to users at the same time. 

Consequently, I will not be splitting the shadow effect settings out for syncing at this point.
I will however, make an issue to highlight these items after this PR is merged.

alexfjw:
New changes:
I've discovered that we are unable to sync custom perspective shadows. Custom perspective shadows are shadows that have been set with a perspective preset (there are 5), and the user has tweaked any shadow setting. 

We cannot sync them as the API tells us that they are an outer shadow (another type of shadow), and there is no easy way to find out exactly what shadow was applied. (other than guessing from image analysis, but that too is quite unreliable)

We can sync non-modified perspective shadows though. 

### UI Changes
Show a warning box when user chooses to copy the custom perspective shadow in the sync dialog.
The warning is shown for both custom perspective shadows & custom outer shadows.
Below is a screenshot of the change. 

The warning shows after the user presses 'OK' in the sync dialog, if the item has a custom perspective /outer shadow.

image

We cannot know which perspective shadow was selected, or if it was a outer shadow. Just apply as outer shadow, since outer shadow covers more cases. (more chance of getting what user wants)

image

Sync Dialog, just for reference:
image

leeyh20:
It works on PPT 2010! Hurray!

alexfjw:
Apologies, the rebase shifted my commit to after the comment. Just for clarification, YanHwa has already tested the latest version of this PR.

PowerPointLabs/PowerPointLabs/TextCollection/SyncLabText.cs
```diff
@@ -22,5 +22,10 @@ internal static class SyncLabText
         public const string ErrorPasteSelectionInvalid = "Please select at least one item to apply this format to.";
         public const string ErrorShapeDeleted = "Error in loading shape formats. Removing invalid formats from the list.";
         public const string ErrorSyncPaneNotOpened = "Error: SyncPane not opened.";
+
+        public const string WarningDialogTitle = "Warning";
+        public const string WarningSyncPerspectiveShadow =
+            "PowerPoint is unable to differentiate between custom perspective shadows and custom outer shadows.\n\n" +
```
initialshl: Should it be "PowerPointLabs"?

PR #1660, opened 2018-03-29
#1650 Unknown Office control ID: PictureCropToolsMenu

### LINKED ISSUE #1650: Unknown Office control ID: PictureCropToolsMenu
image

As above.

The second comment in this link seems to give a fix, but it's not trivial to implement. Probably needs a bit of planning to be done well.
https://stackoverflow.com/questions/9935539/outlook-addin-has-failed-to-find-office-control-by-id

junkiattan:
Fixes #1650 

Fix is to simply include in a built-in control that caters to all versions of PPT, PPT2010/2013/2016.

Tested and the error is removed.

@leeyh20 can help me test on PPT2010 to ensure that the error message does not appear when the "Show add-in user interface errors" option in the Options -> Advanced settings is checked?


leeyh20:
Tested on PPT 2010, yes the error is gone.
image


However, is it intended that the other default crop features like in the below picture are not included? 
image

junkiattan:
@leeyh20 

Yes it is intended. Observe that the other default crop features can be easily performed using current features in our labs, i.e. Crop Lab and Paste Lab.

Therefore there is no need to include those features. I think the most important default crop feature not provided explicitly is the basic crop.

leeyh20:
Just a suggestion, maybe the name of your PR should have #1650 at the front instead so that it is more consistent with the other PRs

PR #1664, opened 2018-04-01
#1644 SyncLab ArtisticEffect Fails to Sync

### LINKED ISSUE #1644: SyncLab: Artistic Effect Fails to Sync, 2013

alexfjw:
Fixes #1644 

### **Outline of Solution**

Since 2013 permanently applies ArtisticEffect to shapes after powerpoint in `SyncLabShapeStorage.cs` is saved, do the following:
Only for 2013, 
1) stash all ArtisticEffects when adding new shapes to SyncLabShapeStorage
2) load these ArtisticEffects back to the shape when retrieving shapes from SyncLabShapeStorage
3) clear the ArtisticEffects as appropriate, when shapes are deleted from SyncLabShapeStorage

This fix should not hinder ArtisticEffect syncing on 2010 & 2016.

### Others
PictureEffect was renamed to ArtisticEffect, to reflect the naming in the PowerPoint application, rather than VSTO function name.

The ArtisticEffect test has also been fixed. 
The failing test was due to a mistake in slide numbering.


PowerPointLabs/PowerPointLabs/SyncLab/SyncLabShapeStorage.cs
```diff
@@ -61,8 +62,18 @@ public string CopyShape(Shape shape, Format[] formats)
...
                 try
                 {
-                    shape.Copy();
+                    // workabout for 2013's artistic effect, (2010 & 2016 do not require this)
+                    // copied shapes will have their artistic effect permenantly set to the shape
+                    // remove the artistic effect before copying, and restore it after pasting
+                    Shape duplicate = shape.Duplicate()[1];
+                    ArtisticEffectFormat.RemoveArtisticEffect(duplicate);
+                    
+                    duplicate.Copy();
```
leeyh20: This will affect the clipboard right? Do you want to use the RestoreClipboardfAterAction function?

PowerPointLabs/Test/UnitTest/SyncLab/SyncLabArtisticEffectTest.cs
```diff
@@ -4,48 +4,47 @@
...
         {
-            return "SyncLab\\SyncLab_VisualEffects.pptx";
+            return "SyncLab\\SyncLab_ArtisticEffect.pptx";
         }
-        
+
         [TestMethod]
         [TestCategory("UT")]
-        public void TestReplaceArtisticEffect()
+        public void TestSetArtisticEffect()
```
leeyh20: Would it be better to name it TestSyncArtisticEffect so that it is consistent with the other synclab tests?

alexfjw: There's another method called TestSyncArtisticEffect, which does that.
ArtisticEffect is a little different from other formats as we have to replace all existing artistic effects on the destination object. This test focuses on the 'replacing' part.

PowerPointLabs/PowerPointLabs/SyncLab/ObjectFormats/ArtisticEffectFormat.cs
```diff
@@ -0,0 +1,114 @@
...
+            }
+        }
+        
+        private static bool Sync(Shape formatShape, Shape newShape)
+        {
+            try
+            {
+                // access PictureEffects, just to make sure shapes are compatible with ArtisticEffect
+                // will throw an exception otherwise
+                PictureEffects dest = newShape.Fill.PictureEffects;
```
PPTLabsBot: MINOR Remove this unused 'dest' local variable. [rule](https://sonarcloud.io/coding_rules#rule_key=csharpsquid%3AS1481)

PowerPointLabs/PowerPointLabs/SyncLab/ObjectFormats/ArtisticEffectFormat.cs
```diff
@@ -0,0 +1,114 @@
...
+                PictureEffects source = formatShape.Fill.PictureEffects;
```
PPTLabsBot: MINOR Remove this unused 'source' local variable. [rule](https://sonarcloud.io/coding_rules#rule_key=csharpsquid%3AS1481)

alexfjw:
@leeyh20 I've found a bug in my previous approach. 
Have used a different method to ensure ArtisticEffect syncs now.

PowerPointLabs/PowerPointLabs/SyncLab/ObjectFormats/ArtisticEffectFormat.cs
alexfjw: We access the effects just to see if an exception is thrown

PowerPointLabs/PowerPointLabs/SyncLab/ObjectFormats/ArtisticEffectFormat.cs
alexfjw: We access the effects just to see if an exception is thrown

leeyh20:
@alexfjw  Oops your syncing of artistic effect does not sync the options?

image

These special options are not synced. However, every artistic effect has slightly different options:

image

alexfjw:
Good catch there. It seems that each picture effect might have special parameters to set. 
I didn't notice it as it wasn't very visible in the ArtisticEffect API.
I'll create a new issue to highlight that.

leeyh20:
Okay then! It works on PPT 2010 for the default settings.

leeyh20:
Does it work in 2016?

alexfjw:
Yep, it does 👍

PowerPointLabs/Test/UnitTest/SyncLab/SyncLabArtisticEffectTest.cs
```diff
@@ -4,48 +4,47 @@
...
+        private void SyncArtisticEffect(string shapeToSync, int sourceSlideNumber, int expectedSlideNo)
         {
             Microsoft.Office.Interop.PowerPoint.Shape formatShape = GetShape(sourceSlideNumber, SourceShape);
 
-            Microsoft.Office.Interop.PowerPoint.Shape newShape = GetShape(sourceSlideNumber, shapeToPaste);
-            new PictureEffectsFormat().SyncFormat(formatShape, newShape);
+            Microsoft.Office.Interop.PowerPoint.Shape newShape = GetShape(sourceSlideNumber, shapeToSync);
+            new ArtisticEffectFormat().SyncFormat(formatShape, newShape);
 
             CompareSlides(sourceSlideNumber, expectedSlideNo);
```
IanTeo: Is this actually enough to compare if the sync actually succeeded? If I remember correctly, this comparison is quite lenient and sometimes we have to make the shape differences very obvious to test it this way. I would suggest we try to test if the attributes directly (if possible) like using IsSameShape, although we would have to add those attributes to be tested too.

PowerPointLabs/PowerPointLabs/SyncLab/ObjectFormats/ArtisticEffectFormat.cs
```diff
@@ -0,0 +1,114 @@
...
+            Bitmap image = GraphicsUtil.ShapeToBitmap(formatShape);
+            return image;
+        }
+
+        public static List<MsoPictureEffectType> GetArtisticEffects(Shape shape)
+        {
+            List<MsoPictureEffectType> artisticEffects = new List<MsoPictureEffectType>();
+            try
+            {
+                PictureEffects effects = shape.Fill.PictureEffects;
```
IanTeo: Is it possible to choose multiple artistic effects? How do you do it (I am using ppt 2013). If it is possible, then our tests are a bit lacking, we should have a test for this case.

PowerPointLabs/PowerPointLabs/SyncLab/ObjectFormats/ArtisticEffectFormat.cs
```diff
@@ -0,0 +1,114 @@
...
+        {
+            try
+            {
+                // access PictureEffects, just to make sure shapes are compatible with ArtisticEffect
+                // will throw an exception otherwise
+                PictureEffects dest = newShape.Fill.PictureEffects;
+                PictureEffects source = formatShape.Fill.PictureEffects;
+                
+                List<MsoPictureEffectType> effectTypes = GetArtisticEffects(formatShape);
+                ClearArtisticEffects(newShape);
```
IanTeo: Why do we have the effects of the new shape? Is it because it will not override the artistic effect if there is already one?

PowerPointLabs/PowerPointLabs/SyncLab/ObjectFormats/ArtisticEffectFormat.cs
alexfjw: I think you can do so by adding one artistic effect after another. Selecting a new effect adds a new one, rather than replaces the old one. There's a 'clear' button which clears all artistic effects.

PowerPointLabs/PowerPointLabs/SyncLab/ObjectFormats/ArtisticEffectFormat.cs
alexfjw: That's right. I'll add that as a comment to the file to make it clearer.

PowerPointLabs/Test/UnitTest/SyncLab/SyncLabArtisticEffectTest.cs
alexfjw: I think it's sufficient as the artistic effects we are using for the test greatly changes the look of each shape. I'll look at it once more and check back just to be sure.

alexfjw: Test for Replacing Effects:
image

The shape in the top left has it's effects changed after syncing.


Test for Setting Effects:
image

We set a checkered effect here. It also causes the overall lighting of the shape to change as well. 

I believe the change in images is large enough to be sufficient when we compare the slides.

alexfjw: Also note that ArtisticEffect is only applicable to picture type shapes. I'll be adding this in as a comment as well.

PowerPointLabs/PowerPointLabs/SyncLab/ObjectFormats/ArtisticEffectFormat.cs
IanTeo: So are you going to add tests for such a case? Or is it already covered by one of the tests?

PowerPointLabs/PowerPointLabs/SyncLab/ObjectFormats/ArtisticEffectFormat.cs
IanTeo: Continuing from #1640, discussion on SonarQube.
For example such errors, I remember we had a discussion on this before and concluded that Microsoft suggests to use unused variables to improve readability of the code. Such rules should be removed from SonarQube in that case.

PowerPointLabs/PowerPointLabs/SyncLab/ObjectFormats/ArtisticEffectFormat.cs
alexfjw: Deepest apologies Ian I've made a mistake on my statement that we can apply more than 1 artistic effect on picture, 
It appears that we can't actually stack artistic effects despite how it sounds in the API. It does however, make me wonder why the API treats ArtisticEffect as a collection of effects. 
I'll note this down as a comment in the ArtisticEffectFormat file.

alexfjw:
Thanks for the clarification on slide similarity Ian, 
I did not expect the slide similarity test to fail, but it did as you suspected.
I've changed the test to work on a more obvious artistic effect. (see screenshot below)

The 2 slides compared are the following, which gives a base difference of 96% without syncing.
I've bumped up the slide similarity threshold to 99% for these tests. 
Also, I've added some comments to the slide similarity test in the base SyncLab test file to highlight this.

image

PR #1666, opened 2018-04-01
#1213 support for syncing reflections

alexfjw:
Part of #1213

### **Outline of Solution**

- Adds support for reflections to SyncLab.

The preview image for the reflection item in SyncLab has been emphasized by using a fixed transparency, blur and shadow size.


junkiattan:
Tested on PPT2013/2016 and codes works as intended. Test provided also passes

Code LGTM

@leeyh20 Ready to check on 2010

leeyh20:
Works on PPT 2010.

PowerPointLabs/PowerPointLabs/SyncLab/ObjectFormats/ReflectionEffectFormat.cs
```diff
@@ -0,0 +1,73 @@
...
+            try
+            {
+                ReflectionFormat srcFormat = formatShape.Reflection;
+                ReflectionFormat destFormat = newShape.Reflection;
+
+                if (srcFormat.Type != MsoReflectionType.msoReflectionTypeMixed)
+                {
+                    // setting ReflectionFormat.Type automatically sets Reflection settings
+                    // there is no need to set them manually
+                    destFormat.Type = srcFormat.Type;
```
IanTeo: User can actually customize the reflection after setting the default one. Is this handled here? As seen in the else block below, there are actually 4 things that the user can configure. What do you think about allowing them to sync them each individually?

initialshl:
Good to have a peer code review before i review this

PowerPointLabs/PowerPointLabs/SyncLab/ObjectFormats/ReflectionTransparencyFormat.cs
```diff
@@ -0,0 +1,49 @@
...
+            if (!Sync(formatShape, newShape))
+            {
+                Logger.Log(newShape.Type + " unable to sync ReflectionTransparency format");
+            }
+        }
+
+        public override Bitmap DisplayImage(Shape formatShape)
+        {
+            return SyncFormatUtil.GetTextDisplay(
+                Math.Round(formatShape.Reflection.Transparency * 100).ToString() + "%",
```
leeyh20: Do you think it is needed to explain why need to multiply by 100?

alexfjw: Hmm.. the docs do say that it's a float between 0 & 1.
And we're displaying as a percentage so.. maybe not?

leeyh20: Oh if the docs mention then it's fine then!

alexfjw: Cancel that, the docs are a bit ambiguous. I'll update that in a bit

PR #1668, opened 2018-04-02
#1593 Inserting Shape should allow users to remain in PowerPointLabs tab

### LINKED ISSUE #1593: Inserting Shape should allow users to remain in PowerPointLabs tab
When users insert shapes from any one of the labs in PowerPointLabs, they should remain in that tab, instead of being transported away to the Format tab as with the normal insert shape

junkiattan:
Fixes #1593

**Outline of Solution**

**Tried solutions**

1. Tried to add custom callbacks to builtin controls directly. Did not work for both galleries or toggle buttons.

1. Because of API limitations, instead of trying to activate the Powerpointlabs tab, one way is to programmatically disable format tab instead by affecting its getVisible callback. However it is not possible to check during run time the current active tab and disable format tab during run time. 

**Final solution**: One workaround is to create a checkbox to let users have the freedom to specify whether they want or do not want the format tab to be disabled. Checkbox is a custom control so it is possible to add in a custom callback which disables or enables the format tab during run time.

Tested on PPT 2016/2013

untitled


PowerPointLabs/PowerPointLabs/Ribbon1.cs
```diff
@@ -136,6 +138,19 @@ public string GetCustomUI(string ribbonId)
         #region Ribbon Callbacks
         //Create callback methods here. For more information about adding callback methods, select the Ribbon XML item in Solution Explorer and then press F1
 
+        // Set the visibility of the Drawing Tools Format Tab
+        public bool GetVisibleFormatTab(Office.IRibbonControl control)
```
leeyh20: Is this function used in the code?

leeyh20: Is this method used at all in the code?

leeyh20:
On PPT 2010, the checkbox seems to make no difference to the format tab. But anyway in the first place, users are not teleported to the format tab when inserting a shape in PPT 2010.

junkiattan:
@leeyh20 
Thanx for your spot! I realised that because of previous merge conflicts with the xml files, it removed a line I originally added for the solution, which is why you saw a method that was created but not used. I have sinced rectified that, and also updated the labels to make the checkbox more clear.

Would appreciate an approval for 2010 if good to go.

leeyh20:
On PPT 2010, now the checkbox does enable/disable the format tab.

PowerPointLabs/PowerPointLabs/Ribbon1.cs
```diff
@@ -136,6 +138,19 @@ public string GetCustomUI(string ribbonId)
         #region Ribbon Callbacks
         //Create callback methods here. For more information about adding callback methods, select the Ribbon XML item in Solution Explorer and then press F1
 
+        // Set the visibility of the Drawing Tools Format Tab
+        public bool SetVisibleFormatTab(Office.IRibbonControl control)
```
initialshl: This method name is misleading since it doesnt set

junkiattan: Actually it does set the visibility of the format tab, just that the setting is not up to a direct input from the user, but dependent on the status of the checkbox

junkiattan:
@initialshl Screenshot added to PR

PowerPointLabs/PowerPointLabs/Ribbon1.cs
initialshl: Would it be better to call it toggle instead of set? Since without a parameter set may sound like setting it to true everytime

junkiattan:
@initialshl 

I have updated the naming of the method. Would appreciate your approval if this PR is good to go;)

IanTeo:
Just a comment: The purpose of the button is not very clear from the user's perspective.
From their point of view, why would they "Hide format tab"? It is a useful tab that they use to format their shapes. So when a user sees such a functionality they might not see the use case of it. I think a name change would make it clearer, or maybe a tooltip. What do you think @damithc?

junkiattan:
@IanTeo How about a tool tip message like this "Allows you to remain on the PowerPointLabs tab after inserting shapes"

initialshl:
Dont think users often see the supertips.
Some suggestions: "Maintain tab focus" or "Keep focus after action" etc

junkiattan:
@initialshl 

capture

Hows this? I have updated both labels to reflect the function better

initialshl:
I personally prefer "Maintain Tab Focus", "Stay on PowerPointLabs tab after inserting a shape", "Settings".
You (or UI/UX designer) can make the final decision

junkiattan:
@initialshl Ill keep Tab Management but include your suggestion on the more direct description for the supertip

PR #1669, opened 2018-04-02
#1213 support for glow

alexfjw:
A part of #1213 

### **Outline of Solution**

- Similar to other Sync PRs, adds support for Glow effect.

- The glow effect is emphasized in the sync preview.


alexfjw:
Thanks, I've rectified the issue. Seems like we need to do some bounds checking. The value must be 150 or lower. Otherwise, an exception is thrown.

junkiattan:
@leeyh20 Tested it works now for both PPT2013/2016

Awaiting approval for 2010

leeyh20:
Works on PPT 2010.

However, Glow is not very obvious on the SyncPane on the right?
(ignore the team viewer)
image

junkiattan:
@alexfjw 

I agree with @leeyh20, I feel that the glow could be more obvious

alexfjw:
I don't think we should edit the SyncPane though. 
One quick fix that comes to mind is to resize the shape itself. 
However, that will change the size of text and so on, which is pretty tricky to deal with.
Perhaps it'll be better to make a new issue to highlight this issue?

junkiattan:
@alexfjw 

Sure, that can be done

leeyh20:
Glow effect still works on PPT 2010 :)

alexfjw:
@IanTeo Think this is ready for another review 👍

PR #1671, opened 2018-04-03
#1213 support for syncing softedge

alexfjw:
A part of #1213

### **Outline of Solution**
Adds support for Soft Edge effect in SyncLab.
The image preview modifies the actual SoftEdge effect to make it more visible.


junkiattan:
Tested on 2013 and 2016. Syncing for Soft Edge works as expected.

In addition, provided test also works.

junkiattan:
@leeyh20 Can you do a check on 2010 and update the labels if everything is good to go?

leeyh20:
(as usual T___T) 

The options of "Size" is being overriden when I press Sync on the sync pane.

image

I set the "Size" to a very low value, but when I press Sync, it goes back to the default "Size".

alexfjw:
@leeyh20 I can't seem to replicate your problem. I've tried using decimals for soft edge (30.23) in the included test file, and it seems to work fine.

leeyh20:
Last bug!
Size is small but in the properties, it is overemphasised.

image

initialshl:
Would be good to have at least 1 approving code review before i start to give my review

alexfjw:
Merged dev-release again,
All tests are fine.

PR #1672, opened 2018-04-03
#1662 Make Agenda Reference Slide sync Design's Slide Master (Theme of the slide) with other agenda slides

### LINKED ISSUE #1662: Theme of the template slide is not synchronized into other agenda slides
image
 
#### Environment
PPT2013, Win10

#### Steps to replicate

1. Generate text agenda 
1. Change the design theme of the template slide 
1. Synchronize agenda slides

#### Expected behavior

All agenda slides should be changed to the theme used by the template slide

#### Actual behavior

Agenda slides remains the same (default) theme

leeyh20:
Fixes #1662

**Outline of Solution**
Fixed agenda reference slide's Design not syncing entire Slide Master (which is the theme of the slide) to other agenda slides by:
1. Cloning the Design of the reference slide and then renaming it.
2. If Design already exist then it is deleted first.

**Failed Methods**
- refSlide.GetNativeSlide().Background.Shapes() and then copying over and pasting to design.SlideMaster.Background.Shapes() created more problems
- refSlide.GetNativeSlide().Design.SlideMaster.Background.Shapes() and then copying over and pasting to design.SlideMaster.Background.Shapes() also created more problems
- Copying other Background.Fill types over to design.SlideMaster.Background.Fill was not working because often the types could only be "get" but not "set"

**Testing**
Tested on PPT 2016 and works well even after syncing agenda multiple times. To test this PR, should try **changing the theme of the agenda's reference slide** and then syncing agenda multiple times manually and ensure nothing is breaking (since the design is cloned...)
Passes Text Agenda tests on PPT 2016.

@damithc This should fix your issue!

### DIFF
--- a/PowerPointLabs/PowerPointLabs/Utils/SlideUtil.cs
+++ b/PowerPointLabs/PowerPointLabs/Utils/SlideUtil.cs
@@ -1,4 +1,5 @@
 ﻿using System.Collections.Generic;
+using System.Runtime.InteropServices;
 
 using Microsoft.Office.Core;
 using Microsoft.Office.Interop.PowerPoint;
@@ -7,6 +8,7 @@ using PowerPointLabs.Models;
 
 using Shape = Microsoft.Office.Interop.PowerPoint.Shape;
 using ShapeRange = Microsoft.Office.Interop.PowerPoint.ShapeRange;
+
 
 namespace PowerPointLabs.Utils
 {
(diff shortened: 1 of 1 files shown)


junkiattan:
Tested on PPT2013 and it works.

PowerPointLabs/PowerPointLabs/Utils/SlideUtil.cs
```diff
@@ -119,12 +121,13 @@ public static Design GetDesign(string designName)
         public static void CopyToDesign(string designName, PowerPointSlide refSlide)
         {
             Design design = GetDesign(designName);
-            if (design == null)
+            if (design != null)
             {
-                design = CreateDesign(designName);
+                design.Delete();
```
junkiattan: see if can add a try-catch, in case design == null and you get a NullReferenceException

@alexfjw Help to test on 2016

leeyh20: Hmm I don't think there will be a NullReferenceException because I check if design is not equals to null. However, ok I'll add a Try-Catch anyway because I'm not sure under what circumstances the design cannot be deleted. (There isn't an API documentation)

leeyh20:
@junkiattan  Have made the requested changes.

PR #1673, opened 2018-04-05
#1213 support for syncing bevel 

alexfjw:
A part of #1213 

### **Outline of Solution**

Adds support for syncing bevel effect. (and other related 3d effects)

No changes to preview, as bevel is a pretty complex item.
There's about 5 separate parts:
- top bevel
- bottom bevel
- depth
- contour
- material & lighting


junkiattan:
@alexfjw 

Function looks good on PPT2013/2016 and test provided works as well. Just a thought. As i was testing the chnaging the bevels for testing, I was wondering whether 3D formats like Material, Depth, etc can be synced independently instead of being tied to Bevel.

Because some people may change material, but leave bevel unchanged. In that case when they try to sync the formats, there isnt an option to sync just the material

alexfjw:
Good point there, what do you think of splitting it up into 
- bevel
- contour
- lighting
- material

junkiattan:
I think it would be optimal to split into 6 parts
- Top bevel
- Bottom bevel
- Depth
- Contour
- Material
- Lighting

This is because sync lab is about the control it gives to the users to sync over certain formats. Best to break it down as much as possible. However if its due to API limitations, I am fine with the proposedfew as well

alexfjw:
The test can be better split, to test each 3d effect component separately.
I've left that out in the interest of time. We do test out all effects in the existing test, but it won't be clear which has failed when the test fails.

junkiattan:
@alexfjw 

Will it be good to submit a new issue requesting for more in-depth test coveragefor syncing of 3d formats?

alexfjw:
Yea I think that's a good idea. I'll be doing that right after this PR merged. 👍

junkiattan:
@alexfjw I found 4 errors with the syncing of 3d formats

1. When you sync contours, if the size is <1, e.g. 0.1pts, the preview reflects 0pt. Best to check for Depth size as well
1. Syncing of angle in Lighting is absent
1. When bevel (top/down) is not specified/modified but user syncs everything under Visual Effects, when copied onto another shape, the bevel changes from "None" to "Round"
1. When a user tries to sync an unmodified shape, an exception is thrown as shown below
untitled

alexfjw:
Thanks for the help JunKiat.

> When you sync contours, if the size is <1, e.g. 0.1pts, the preview reflects 0pt. Best to check for Depth size as well

Has been addressed 

> Syncing of angle in Lighting is absent

Has been addressed 

> When bevel (top/down) is not specified/modified but user syncs everything under Visual Effects, when copied onto another shape, the bevel changes from "None" to "Round"

Has been addressed, looks like editing the bevel settings for a shape with bevel "None" will automatically set it to "Round". I've ignored syncing the bevel settings if the source shape has bevel of type "None". 

> When a user tries to sync an unmodified shape, an exception is thrown 

Has been addressed

Additional Notes:
Bevel type & bevel settings (inset, width) are not split, as they are closely related to the bevel type itself. I feel it is quite unlikely for users to only want to sync bevel width, since the same width on a different type has a very different appearance.

junkiattan:
@alexfjw 

Bugs 1-3 are fixed. However when syncing default shapes with no modifications, the depth size auto changes to 36pts even though it should be 0pt

alexfjw:
I've added the UTs for each 3d effect.

junkiattan:
I have tested on 2013 and the syncing of bevel works as expected.

alexfjw:
Fixed the merge conflict

PR #1694, opened 2018-04-18
#1691 Format Tab gets focus even when maintain tab focus tab is ticked

### LINKED ISSUE #1691: Format Tab gets focus even when maintain tab focus tab is ticked
PPT2013, Win10

Steps:
1. Tick the 'maintain tab focus' box
2. insert a new shape

Expected: PowerPointLabs tab retains focus
Actual: Format tab gets focus

junkiattan:
Fixes #1691 

Limitations:
- the ribbon only load once, even if new presentations are open, hence there is only a single "Format tab" object. As a result of that, the behaviour faced in the issue because of different presentations setting the visibility of the same Format Tab differently.

Solution
- Since the ribbon load only once, to set the boolean value when ribbon loads, and sync it across all the open/new presentations whenever there a checkbox in one of the presentation is clicked. Meaning if there is Presentation 1/2/3, and the checkbox is ticked for presentation 2, when user goes to presentation 1/3, the checkbox would reflect the same ticked status.

Tested on multiple new/open presentations in 2016 and it works.

To Test
- Open multiple new/existing presentations and see if they sync when you check the box on one of them


alexfjw:
Works good on 2013 👍

PowerPointLabs/PowerPointLabs/Ribbon1.cs
```diff
@@ -141,14 +141,26 @@ public string GetCustomUI(string ribbonId)
...
+            _ribbon.InvalidateControl("VisibleFormatShapes");
+        }
+
+        public void InitialiseVisibilityCheckbox()
+        {
+            _ribbon.InvalidateControl("VisibleFormatShapes");
+        }
+        
+        // Sets the default starting status of the checkbox (whether checked or not)
+        public bool SetVisibility(Office.IRibbonControl control)
```
leeyh20: I think it should be renamed to "GetPressed" or "GetPressedCheckbox" because it is not obvious that this is related to the checkbox.

junkiattan: method GetPressed for the control actually sets whether the checkbox is checked, not whether the box has been pressed. Hence the return value for it actually sets the state of the checkbox, hence the keyword "Set in the method"

leeyh20:
Boolean gets updated on PPT 2010 :)

PR #1705, opened 2018-06-12
#1638 Positions Lab Bug (Align)

### LINKED ISSUE #1638: Positions Lab Bug (Align)
I decided to break up #1601 into 3 components, because apparently certain images fail for all of the Align, Adjoin and Distribute operations, and each type of operation touches different parts of the code and is overall quite extensive to be placed under 1 issue.

This issue would solve for the **Align** operations. 

The file that is causing the issue: 
[to_print (1).pptx](https://github.com/PowerPointLabs/PowerPointLabs/files/1831929/to_print.1.pptx)

The facial images are the ones with the issue. The icon images are the images that work well (non-corrupted). Note that any change must work for both corrupted as well as non-corrupted images.

tankztz:
Fixes #1638 

*Fixed bugs for specific images using Positions Lab (Align part)

*Fixed bug for PositionsPaneWPF.xaml.cs -> DuplicateShapes function. Before this fix, the function will return two same shapes if the name of input shapes are same. Now this function give duplicated shapes a index number in its Name property. This may fix other bugs related.

*Added comments for ShapeRange.Duplicate() function. The return sequence of this function is not guaranteed.


tankztz:
@admin

PR #1712, opened 2018-07-10
#1701 adjust image compression levels

### LINKED ISSUE #1701: Adjust image compression levels
Currently the constants for [image compression level](https://github.com/PowerPointLabs/PowerPointLabs/blob/master/PowerPointLabs/PowerPointLabs/Utils/GraphicsUtil.cs#L35) as well as the [file size limit](https://github.com/PowerPointLabs/PowerPointLabs/blob/master/PowerPointLabs/PowerPointLabs/Utils/GraphicsUtil.cs#L36) in Graphics Utils already work well for large images.

However lets adjust the values so that the compression would apply to more images. This is because users might need to include a lot of small/medium sized images in their presentations as well, hence it would be good to apply a wider coverage of compression to keep overall file size small.

As shown in this [website](http://regex.info/blog/lightroom-goodies/jpeg-quality), setting a compression level of 75 still preserves image quality without visible artifacts, while further reducing image size, so compression level of 75 is still acceptable. We would reduce the max file size limit to apply compression to a wider range of image sizes.

To do:
1. Set compression level to 75
1. Set file size limit lower to 40000

tankztz:
Fixes #1701 
1. Set compression level to 75.
2. Set file size limit lower to 40000.
3. Tested with PasteToFillSlide method in Office 2017.

### DIFF
--- a/PowerPointLabs/PowerPointLabs/Utils/GraphicsUtil.cs
+++ b/PowerPointLabs/PowerPointLabs/Utils/GraphicsUtil.cs
@@ -32,8 +32,8 @@ namespace PowerPointLabs.Utils
         private const float TargetDpi = 96.0f;
         private static float dpiScale = 1.0f;
         // Heuristics for image compression obtained through testing
-        private const long targetCompression = 95L;
-        private const long fileSizeLimit = 75000L;
+        private const long targetCompression = 75L;
+        private const long fileSizeLimit = 40000L;
 
         // Static initializer to retrieve dpi scale once
         static GraphicsUtil()


alexfjw:
@tankztz could you edit your PR to follow the PR template? Thanks

tankztz:
@alexfjw Done, please check

alexfjw:
Could you do a visual check + size check and report your findings as well? 
We want to know if the settings really do make an impact, and how far we can go with adjusting them.

tankztz:
I checked the 2 settings with 3 different size of image. 

I didn't find any difference.

new_14_1 means new setting (Changed by me), 14 KB, 1st view

 
new_14_1
new_14_2
new_71_1
new_71_2
new_1720_1
new_1720_2
old_14_1
old_14_2
old_71_1
old_71_2
old_1720_1
old_1720_2

alexfjw:
Thanks, could you give a summary, in terms of size savings if we were to
use this approach?
The goal of the issue is ultimately to reduce the size of the powerpoint
file. (Rather than just changing the compression numbers)
If the savings aren't considerable, we can conclude that there isn't much
value in adjusting compression and close the issue.

tankztz:
For my own sample Presentation1.pptx

Original size: 465KB
Original size + 95 compression level image of 1720KB: 749KB
Original size + 75 compression level image of 1720KB: 579KB

The quality change on image is not visible. 

Thus I think we should keep this change.

alexfjw:
Great.

One last point which just came to mind:
Could you also see if changing the compression has effects on Zoom Lab's
features?

I'm not particularly familiar with ZoomLab, but I believe we do enlarge
images programatically ZoomLab. It would be best if we can verify that
changes in compression do not result in lower quality there.

tankztz:
@alex 
This change only affect paste to fill slides feature since it only have 2 reference both from PasteToFitSlide class.

damithc:
> I suppose we can assume users will not use paste to fit slide with zoom lab.

Well, some people might. :-p e.g., find a high-res image from interwebs, paste to fill slide, and then use zoom lab on that slide. 

Perhaps in the documentation we can warn about the loss of resolution when using paste-to-fill and instruct users to use a normal paste if they wish to keep the high resolution?

tankztz:
@damithc 
In this case, we can compare the this two pic:

new_1720_2:
https://user-images.githubusercontent.com/14175589/42500229-418108ae-8463-11e8-9df4-466bde5e75ee.png

old_1720_2:
https://user-images.githubusercontent.com/14175589/42500237-42a0a780-8463-11e8-92fd-011525bff0ca.png

I think the difference if negligible. 

btw, Zoom lab itself compress the image to a very low quality. 

normal:
zoom_1

After zoom to area:
zoom_2

damithc:
> I think the difference if negligible.

Thanks for checking @tankztz 

> btw, Zoom lab itself compress the image to a very low quality.

I think so. It's something we wanted to fix, but never got around to it.

tankztz:
I think this PR can be merged if no further reviews.

PR #1717, opened 2018-08-16
Change addresses to https

leeyh20:
**Outline of Solution**
Emergency fix to https:// due to nus computing changes regarding http links

PR #1723, opened 2018-09-29
#1721 tutorial does not open after installation

### LINKED ISSUE #1721: Tutorial file does not open after installation
PowerPointLabs gives an error message when trying to open the tutorial file immediately after installation.

LiXiaoooowei:
Fixes #1721

I provided a temporary fix to the issue by changing `Process.start("POWERPNT", sourceFile)` to `Process.start(sourceFile)`. This will open the link in broswer, and prompt user to choose whether to open it directly or save locally. User can choose to open directly and access the tutorial file.

This issue is likely caused by incompatible SSL/TLS configuration between server and the security protocol used by .NET. I have tried to read contents of the tutorial file as byte array using `WebClient.DownloadData` or `HTTPRequest.GetResponse` but neither works.

Another possible walkaround is to include tutorial.pptx file in the download package, so that `Process.start('POWERPNT', sourceFile)` can open the local pptx file directly.

### DIFF
--- a/PowerPointLabs/PowerPointLabs/app.config
+++ b/PowerPointLabs/PowerPointLabs/app.config
@@ -14,10 +14,10 @@
     <value>online</value>
    </setting>
    <setting name="ReleaseAddr" serializeAs="String">
-    <value>http://www.comp.nus.edu.sg/~pptlabs/download-78563/</value>
+    <value>https://www.comp.nus.edu.sg/~pptlabs/download-78563/</value>
    </setting>
    <setting name="DevAddr" serializeAs="String">
-    <value>http://www.comp.nus.edu.sg/~pptlabs/download/dev/</value>
+    <value>https://www.comp.nus.edu.sg/~pptlabs/download/dev/</value>
    </setting>
    <setting name="Version" serializeAs="String">
     <value>4.3.0.1</value>


initialshl:
Hi, thank you for your pull request. Have you tested that your fix works with no bugs?

On another note, I checked and found that some URI in our settings (https://github.com/PowerPointLabs/PowerPointLabs/blob/8d9045952c019603dad3341c2e7379b15932e296/PowerPointLabs/PowerPointLabs/app.config) are set as HTTP instead of HTTPS.

I typed http://www.comp.nus.edu.sg/~pptlabs/download-78563/ into my browser and it redirected me to SoC homepage, while HTTPS url works fine. Can you test if this issue can be fixed by changing the URLs to HTTPS?

LiXiaoooowei:
@initialshl Yes you are right. Changing `http` to `https` does the work. Tested on Office 2016.

initialshl:
@LiXiaoooowei Do you mind creating another PR with master as base branch? We need it as emergency fix for this public-facing error. Thanks!

initialshl:
Related merged PR: https://github.com/PowerPointLabs/PowerPointLabs/pull/1717

I will merge this PR for the dev-release branch, but should be synced in future.

damithc:
> I typed [http://www.comp.nus.edu.sg/~pptlabs/download-78563/](http://www.comp.nus.edu.sg/%7Epptlabs/download-78563/) into my browser and it redirected me to SoC homepage, while HTTPS url works fine. Can you test if this issue can be fixed by changing the URLs to HTTPS?

This behavior seems to be intermittent. SoC IT folks says it should never happen i.e., it should land in the original location with http switched automatically to https. But at times it does get redirected to SoC home page.

leeyh20:
Thanks for testing it! I was not confident of this fix previously thus I did not deploy it, but I will deploy it now.

PR #1725, opened 2018-12-05
Integrate Azure speech services to narrations lab

LiXiaoooowei:
Integrate NarrationsLab with Azure speech services. 
To try out Azure speech services:
  1. Go to [Azure Speech Service](https://azure.microsoft.com/en-us/services/cognitive-services/speech-services/)
  2. Click on "Try Speech Services" green button
  3. Select "7 day trial" on the pop-up window
  4. Follow the instructions to register for an Azure account
  5. Copy down the endpoint address and API key
  6. Click "Settings" in NarrationsLab
  7. Click on "Click to Log In" button on the pop up window
  8. Select a voice from dropdown list 
  9. Select Human Voice audio button
  10. Click "OK" button


damithc:
What to do after 7 day trial expires?

LiXiaoooowei:
> What to do after 7 day trial expires?

The API key and endpoint is valid for 30 days, after that, one needs to register for Azure speech service which is charged based on usage.

damithc:
> The API key and endpoint is valid for 30 days, after that, one needs to register for Azure speech service which is charged based on usage.

There's no free tier signup?

LiXiaoooowei:
> > The API key and endpoint is valid for 30 days, after that, one needs to register for Azure speech service which is charged based on usage.
> 
> There's no free tier signup?

There exists free-tier options according to the pricing below. 
capture

junkiattan:
Hey @LiXiaoooowei 

Great job linking NarrationsLab to Azure Speech Services!;)

Just wanted to check with you a few things:
1. I realized there is a "1 concurrent request" under the free tier. For the single powerpointlabs user, does it affect the usage of NarrationsLab?
1. Does the Free tier only apply to the 30 day period (including the 7 day trial)?
1. After the 30 days period, if a user start paying I assume he/she will be under the Standard tier. What if the user does not pay after 30 days? Can the user still use under the limited Free tier or will the API key and endpoint become unusable?
1. Which platforms have you tested this service (2016/2013/2010)? Would be glad if you could indicate this as well;)

Thank you!

leeyh20:
Hi @LiXiaoooowei Could you also post a screenshot of the changes that you have made on the NarrationsLab settings? If the user does not want to use the Azure Speech Services or the service is unavailable, will it auto revert to the default voices?

LiXiaoooowei:
@junkiattan A user can try Azure speech services for free without registering for an Azure account for 30 days. Upon exceeding 30 days, he/she can register for Azure account for free and continue using Cognitive Speech Services under free tier. In this case, a new endpoint and API key will be issued. There are limited usage for free tier. If user exceeds free tier usage, I assume he/she will be charged according to standard tier. 
I tested on ppt-2016.

leeyh20:
LGTM.  Would it be possible to make the "Click to Log In" button look the same as the "Ok" and "Cancel" buttons for consistency sake?

PowerPointLabs/PowerPointLabs/SpeechEngine/TextToSpeech.cs
```diff
@@ -59,11 +67,105 @@ public static void SpeakString(String textToSpeak)
...
+            {
+                RequestUri = new Uri(requestUri),
+                Text = textToSpeak,
+                VoiceType = humanVoice.voiceType,
+                Locale = humanVoice.Locale,
+                VoiceName = humanVoice.voiceName,
+                // Service can return audio in different output format.
+                OutputFormat = AudioOutputFormat.Riff24Khz16BitMonoPcm,
+                AuthorizationToken = "Bearer " + accessToken,
+            }, filePath).Wait();
```
leeyh20: Was just wondering, if the service returns an error after the async request, what kind of error would you display? 

For example, the user has already logged in but the user's internet suddenly goes down and we are unable to contact the service. Another case would be that the service suddenly starts returning errors to us. Would there be an error message to the user?  It would be good if you gave a bit more details for your solution on this PR, especially how you send an async request to the speech service and convert the response to the appropriate WAV format. Thanks!

PowerPointLabs/PowerPointLabs/NarrationsLab/ViewModel/Synthesize.cs
```diff
@@ -0,0 +1,245 @@
...
+            }
+        }
+
+        public class InputOptions
+        {
+            public InputOptions()
+            {
+                this.Locale = "en-us";
+                this.VoiceName = "Microsoft Server Speech Text to Speech Voice (en-US, ZiraRUS)";
+                // Default to Riff16Khz16BitMonoPcm output format.
```
leeyh20: Is there a reason why you would default to this? Is this format better?

PowerPointLabs/PowerPointLabs/SpeechEngine/TextToSpeech.cs
```diff
@@ -59,11 +68,106 @@ public static void SpeakString(String textToSpeak)
...
+                return;
+            }
+            string requestUri = UserAccount.GetInstance().GetUri();
+            var cortana = new Synthesize();
+
+            cortana.OnAudioAvailable += SaveAudioToWaveFile;
+            cortana.OnError += ErrorHandler;
+
+            // Reuse Synthesize object to minimize latency
+            cortana.Speak(CancellationToken.None, new Synthesize.InputOptions()
```
LiXiaoooowei: The default values are not currently used in code because attributes in InputOptions are always reset when issuing request.

PowerPointLabs/PowerPointLabs/SpeechEngine/TextToSpeech.cs
```diff
@@ -59,11 +68,106 @@ public static void SpeakString(String textToSpeak)
...
+                Logger.Log(ex.ToString());
+                Logger.Log(ex.Message);
+                return;
+            }
+            string requestUri = UserAccount.GetInstance().GetUri();
+            var cortana = new Synthesize();
+
+            cortana.OnAudioAvailable += SaveAudioToWaveFile;
+            cortana.OnError += ErrorHandler;
+
```
LiXiaoooowei: Upon audio is available, `SaveAudioToWaveFile` function will be called, which reads in stream input byte by byte, and finally convert to wav file. 
Any error (most likely network error) that occurred during the audio generation process is handled by function `ErrorHandler`.

PowerPointLabs/PowerPointLabs/SpeechEngine/TextToSpeech.cs
```diff
@@ -59,11 +68,106 @@ public static void SpeakString(String textToSpeak)
...
+        private static void SaveStringToWaveFileWithHumanVoice(string textToSave, string filePath)
+        {
+            string accessToken;
+            string textToSpeak = GetHumanSpeakNotesForText(textToSave);
+
+            try
+            {
+                Authentication auth = Authentication.GetInstance();
+                accessToken = auth.GetAccessToken();
+                Logger.Log("Token: " + accessToken);
```
LiXiaoooowei: Account authentication is performed before sending http request. If user account has expired or invalid, an error message will be displayed to user.

leeyh20:
@ChesterSng @blewjy @YuPeiHenry Could anyone test this on another version of PPT or at least try on their 2016 PPT? Thank you!

blewjy:
@LiXiaoooowei @leeyh20 Tested this on **PPT 2013** on Windows 10. Ran into some bugs...

**1. Unable to change back the narration to "Default Voice" after changing it to "Human-like Voice"**
***Steps to reproduce:***
a. Go to NarrationLab settings and login according to the instructions in the OP
b. Select "Human-like Voice" and choose a voice from the drop down. 
c. Click "OK"
d. Go back to NarrationLab settings and select "Default Voice". Click "OK" again.
e. Go back to NarrationLab settings - the selected voice is still "Human-like Voice". If you try to generate audio, it is still playing in the "Human-like Voice".

**2. After logging in and exiting out of settings, going back into settings again clears the input for the login page**
***Steps to reproduce:***
a. Go to NarrationLab settings and login according to the instructions in the OP
b. Before closing the settings window, select "Click to change user account". Notice that your login inputs are still there.
c. Close the settings window, then go back into NarrationLab settings.
d. Select "Click to change user account". The input fields are empty. This results in the next bug.

**3. Clicking "Confirm" in the login page with no input throws `NullReferenceException`**
***Steps to reproduce:***
a. Go to NarrationLab settings and login according to the instructions in the OP
b. Close the settings window, then go back into NarrationLab settings again. (This clears the "Choose access endpoint" field)
c. Select "Click to change user account", which will bring you back to the same login page.
d. Click "Confirm" without entering any input (both text input boxes should be blank), and `NullReferenceException` thrown.
```
System.NullReferenceException
  HResult=0x80004003
  Message=Object reference not set to an instance of an object.
  Source=PowerPointLabs
  StackTrace:
   at PowerPointLabs.NarrationsLab.Views.HumanVoiceLoginPage.ConfirmButton_Click(Object sender, RoutedEventArgs e) in C:\Users\Bryan\Desktop\xiaowei\PowerPointLabs\PowerPointLabs\PowerPointLabs\NarrationsLab\Views\HumanVoiceLoginPage.xaml.cs:line 45
```

It seems that we are not given the option to "log out" after logging in, I think it would be good if this feature was included. On a similar note, I'm not sure if we want the API key to be visible after the user has logged in, because for instance the PowerPoint could be on a shared computer and given that there's a paid option for this speech service, leaving the API key in the menu could pose some privacy issues? (Perhaps someone else would like to chime in on this... It's late, I might be thinking too much) 

Other than these issues, everything else seems to be working fine! I think the human voice is pretty realistic and users that use NarrationLab should like this feature.

LiXiaoooowei:
@blewjy Bugs fixed. Can you try it again?

ChesterSng:
Tested on ppt 2016 windows 10. The bugs that @blewjy encountered seems to be fixed for me!

Just wondering if the tags are supposed to work?
`[AfterClick]` is working, but the other tags are not working.

On a side note, the loading bar in the popup window overflows.

image

This is more apparent in using `Human-Like Voice` because it takes a longer time to load.

Other than these I think the feature is good! Just need to polish up the buttons!

damithc:
Maybe we should not call it 'Human-like voices' because the default voices are human voices too. Perhaps 'In-Computer Voices' and 'Cloud Voices'?

blewjy:
@ChesterSng I believe the other tags were already not working 100% of the time originally. Might need to address that in a separate issue.

@LiXiaoooowei just tested on PPT 2013. All the bugs I've listed above are working now! Just a new small bug with clicking logout then cancel:

**Steps to reproduce:**
a. Login with valid key, then select "Human-like voice" and click OK
b. Go back to NarrationLab settings and click logout, **but don't click OK, click "Cancel"**
c. Type some text into the notes and try to generate narration automatically, will have exception pop up:

```
System.ArgumentNullException
Value cannot be null.
Parameter name: key
Stack Trace:
   at System.Collections.Generic.Dictionary`2.FindEntry(TKey key)
   at System.Collections.Generic.Dictionary`2.get_Item(TKey key)
   at PowerPointLabs.NarrationsLab.Data.UserAccount.GetUri() in C:\Users\Bryan\Desktop\xiaowei\PowerPointLabs\PowerPointLabs\PowerPointLabs\NarrationsLab\Data\UserAccount.cs:line 44
   at PowerPointLabs.SpeechEngine.TextToSpeech.SaveStringToWaveFileWithHumanVoice(String textToSave, String filePath) in C:\Users\Bryan\Desktop\xiaowei\PowerPointLabs\PowerPointLabs\PowerPointLabs\SpeechEngine\TextToSpeech.cs:line 89
   at PowerPointLabs.SpeechEngine.TextToSpeech.SaveStringToWaveFiles(String notesText, String folderPath, String fileNameFormat) in C:\Users\Bryan\Desktop\xiaowei\PowerPointLabs\PowerPointLabs\PowerPointLabs\SpeechEngine\TextToSpeech.cs:line 51
   at PowerPointLabs.NarrationsLab.NotesToAudio.OutputSlideNotesToFiles(PowerPointSlide slide, String folderPath) in C:\Users\Bryan\Desktop\xiaowei\PowerPointLabs\PowerPointLabs\PowerPointLabs\NarrationsLab\NotesToAudio.cs:line 118
   at PowerPointLabs.NarrationsLab.NotesToAudio.EmbedSlideNotes(PowerPointSlide slide) in C:\Users\Bryan\Desktop\xiaowei\PowerPointLabs\PowerPointLabs\PowerPointLabs\NarrationsLab\NotesToAudio.cs:line 239
   at PowerPointLabs.NarrationsLab.NotesToAudio.EmbedSelectedSlideNotes() in C:\Users\Bryan\Desktop\xiaowei\PowerPointLabs\PowerPointLabs\PowerPointLabs\NarrationsLab\NotesToAudio.cs:line 83
   at PowerPointLabs.ActionFramework.NarrationsLab.AddNarrationsActionHandler.ExecuteAction(String ribbonId) in C:\Users\Bryan\Desktop\xiaowei\PowerPointLabs\PowerPointLabs\PowerPointLabs\ActionFramework\NarrationsLab\AddNarrations\AddNarrationsActionHandler.cs:line 33
   at PowerPointLabs.ActionFramework.Common.Interface.ActionHandler.Execute(String ribbonId) in C:\Users\Bryan\Desktop\xiaowei\PowerPointLabs\PowerPointLabs\PowerPointLabs\ActionFramework\Common\Interface\ActionHandler.cs:line 14
```

I think the expected behaviour when user clicks cancel after logging out should be that the user is already logged out. After clicking the logout button, the selection automatically changed to "Default Voice". This should remain when user selects "Cancel"

LiXiaoooowei:
@ChesterSng Hi Chester, thanks for testing! I have fixed the progress bar overflow error. However, the tags are API specific. Cloud voice API currently does not expose the same set of customization as system voice. 
@damithc Names changed. 
@blewjy Bug fixed.

However, cloud based narration sends http request to Azure server.  There is a max limit on the frequency of HTTP request. When a http request fails due to max limit reached (response code 429), I set a retry limit of 3 times with each time imposing an additional waiting time of 20 seconds. Typically, this behavior can appear after generating 16 content-intensive slides.

blewjy:
@LiXiaoooowei Looks good! Tested and working on PPT2013 👍

PowerPointLabs/PowerPointLabs/ActionFramework/NarrationsLab/AddNarrations/AddNarrationsActionHandler.cs
```diff
@@ -29,8 +36,15 @@ protected override void ExecuteAction(string ribbonId)
...
-
-            List<string[]> allAudioFiles = NotesToAudio.EmbedSelectedSlideNotes();
+            List<string[]> allAudioFiles = new List<string[]>();
+            try
+            {
+                allAudioFiles = NotesToAudio.EmbedSelectedSlideNotes();
+            }
+            catch
+            {
+                return;
```
YuPeiHenry: Will there be some exception added here later?

PowerPointLabs/PowerPointLabs/NarrationsLab/NotesToAudio.cs
```diff
@@ -69,7 +73,11 @@ public static List<string[]> EmbedSelectedSlideNotes()
             List<string[]> audioList = new List<string[]>();
 
             List<PowerPointSlide> slides = PowerPointCurrentPresentationInfo.SelectedSlides.ToList();
-
+            if (NotesToAudio.IsAzureVoiceSelected && UserAccount.GetInstance().IsEmpty())
+            {
+                MessageBox.Show("Invalid user account. Please log in again.");
+                throw new Exception();
```
YuPeiHenry: Will there be more contextual information added to the exception?

PowerPointLabs/PowerPointLabs/NarrationsLab/Storage/NarrationsLabStorageConfig.cs
```diff
@@ -0,0 +1,83 @@
...
+            {
+                FileStream file = File.Open(GetAccessKeyFilePath(), FileMode.Open);
+                XElement root = XElement.Load(file);
+                foreach (XElement el in root.Elements())
+                {
+                    user.Add(el.Name.LocalName, el.Value);
+                }
+                string key = user.ContainsKey("key") ? user["key"] : null;
+                string endpoint = user.ContainsKey("endpoint") ? user["endpoint"] : null;
+                if (key != null && endpoint != null)
```
YuPeiHenry: Is there any error handling supposed to be done if either is null?

PowerPointLabs/PowerPointLabs/NarrationsLab/ValueConverters/NarrationsLabSettingsPageValueConverter.cs
```diff
@@ -0,0 +1,35 @@
...
+                    return loginInstance;             
+                default:
+                    Debugger.Break();
+                    return null;
+            }
+        }
+
+        public override object ConvertBack(object value, Type targetType, object parameter, CultureInfo culture)
+        {
+            return null;
```
YuPeiHenry: Is the null return intended? Consider notImplemenetedException.

PowerPointLabs/PowerPointLabs/NarrationsLab/Views/NarrationsLabMainSettingsPage.xaml.cs
```diff
@@ -0,0 +1,203 @@
...
+using PowerPointLabs.TextCollection;
+
+namespace PowerPointLabs.NarrationsLab.Views
+{
+    /// <summary>
+    /// Interaction logic for NarrationsLabMainSettingsPage.xaml
+    /// </summary>
+    public partial class NarrationsLabMainSettingsPage : Page
+    {
+        public delegate void DialogConfirmedDelegate(string voiceName, AzureVoice azureVoiceName, bool isAzureVoiceSelected, bool preview);
```
YuPeiHenry: Should bool preview have a different name (eg, isPreviewing)?

PowerPointLabs/PowerPointLabs/NarrationsLab/Views/NarrationsLabMainSettingsPage.xaml.cs
```diff
@@ -0,0 +1,203 @@
...
+
+        private void OkButton_Click(object sender, RoutedEventArgs e)
+        {
+            try
+            {
+                string defaultVoiceSelected = RadioDefaultVoice.IsChecked == true ? voiceSelectionInput.Content.ToString() : null;
+                AzureVoice azureVoiceSelected = RadioAzureVoice.IsChecked == true ? (AzureVoice)voiceList.SelectedItem : null;
+                if (azureVoiceSelected == null && RadioAzureVoice.IsChecked == true)
+                {
+                    throw new Exception();
```
YuPeiHenry: Will there be any contextual information added?

leeyh20:
Thanks for testing on all three versions :)!

PowerPointLabs/PowerPointLabs/ActionFramework/NarrationsLab/AddNarrations/AddNarrationsActionHandler.cs
leeyh20: If ExecuteAction (part of ActionFramework) cannot be executed properly, the code should show an error message to the user. This seems to just return and do nothing,

leeyh20: After catching the exception, the code should read the exception message and show it to the user?

PowerPointLabs/PowerPointLabs/NarrationsLab/ViewModel/Synthesize.cs
```diff
@@ -0,0 +1,255 @@
...
+            };
+
+            var httpTask = client.SendAsync(request, HttpCompletionOption.ResponseHeadersRead, cancellationToken);
+
+            for (int i = 0; i < 3 && !httpTask.Result.IsSuccessStatusCode; i++)
+            {
+                Thread.Sleep(20000);
+                request = new HttpRequestMessage(HttpMethod.Post, inputOptions.RequestUri)
+                {
+                    Content = new StringContent(GenerateSsml(inputOptions.Locale, genderValue, inputOptions.VoiceName, inputOptions.Text))
```
leeyh20: Could this code be placed in a function considering that it is a duplicate of lines 59 to 64?

PowerPointLabs/PowerPointLabs/SpeechEngine/TextToSpeech.cs
```diff
@@ -1,68 +1,174 @@
...
+            }
+            catch (Exception e)
+            {
+                MessageBox.Show("Error generating audio files. ");
+                Debug.WriteLine(e.Message);
+            }
+        }
+        private static void ErrorHandler(object sender, GenericEventArgs<Exception> e)
+        {
+            Logger.Log("Unable to complete the TTS request: " + e.ToString());
```
leeyh20: Just wondering, would it be better to display this to the user instead of just logger.log?

PowerPointLabs/PowerPointLabs/ActionFramework/NarrationsLab/AddNarrations/AddNarrationsActionHandler.cs
LiXiaoooowei: Edited according to suggestion!

PowerPointLabs/PowerPointLabs/NarrationsLab/ValueConverters/NarrationsLabSettingsPageValueConverter.cs
LiXiaoooowei: Edited!

PowerPointLabs/PowerPointLabs/NarrationsLab/Views/NarrationsLabMainSettingsPage.xaml.cs
LiXiaoooowei: Edited!

PowerPointLabs/PowerPointLabs/NarrationsLab/Views/NarrationsLabMainSettingsPage.xaml.cs
LiXiaoooowei: Edited!

PowerPointLabs/PowerPointLabs/NarrationsLab/NotesToAudio.cs
LiXiaoooowei: Edited!

PowerPointLabs/PowerPointLabs/NarrationsLab/Storage/NarrationsLabStorageConfig.cs
LiXiaoooowei: Edited to delete storage file if either is null. User will then have to log in again upon clicking on settings page.

PowerPointLabs/PowerPointLabs/NarrationsLab/ViewModel/Synthesize.cs
LiXiaoooowei: Thanks for the suggestion! Refactored to a separate method.

PowerPointLabs/PowerPointLabs/SpeechEngine/TextToSpeech.cs
LiXiaoooowei: I chose not to display this error to user because:
    1. In the case of http request failure due to max limit reached (response code 429), 3 retry attempts are handled by the code behind.
    2. In the case of network failure, the function will simply return and do nothing.

Both types of failures will fire `ErrorHandler` method. If user chooses to generate narrations for multiple slides, then multiple http requests will be sent out at one time because multiple threads are called to handle async requests. In the case of case 1 or case 2 error, multiple MessageBox will be displayed to user (because every request failure will fire `ErrorHandler` once). That would be undesirable.

LiXiaoooowei:
@leeyh20 @YuPeiHenry Thank you for your suggestions on possible improvements on error handling and code refactoring. I have edited the code based on your suggestions. Please take a look!

PowerPointLabs/PowerPointLabs/SpeechEngine/TextToSpeech.cs
leeyh20: Oh I see.

leeyh20:
Note that @LiXiaoooowei says that .NET Framework would be upgraded to 4.5 following this PR, since 4.0 lacks a HTTP network assembly which is necessary for cloud based voices.

blewjy:
Tested again on PPT 2013. Feature working well, all tests passing.

PR #1728, opened 2018-12-12
#1384 Format name can overflow or go behind buttons

### LINKED ISSUE #1384: Format name can overflow or go behing buttons
image
image

YuPeiHenry:
Fixes #1384

Ready for review. (No tests written for this)

image

Format names were able to overflow into buttons previously. To address the issue:
1. Use TextBlock to trim text with ellipses.
2. Use 3 grid columns to partition and separate the image, text and the 3 buttons.
3. Tooltip is updated whenever the text is changed. (not using binding)

Minor issue with this method:
When resizing the panel on the right, the TextBlock does not resize to fully utilize the space. This may be addressed by not limiting the width of the TextBlock (Just a suggestion). The text will trail off the screen, but the user can scroll to access the buttons, shown below (Might not be better off). Edit: Removed suggested alternative's image to reduce ambiguity.

### DIFF
--- a/PowerPointLabs/PowerPointLabs/SyncLab/Views/SyncFormatPaneItem.xaml
+++ b/PowerPointLabs/PowerPointLabs/SyncLab/Views/SyncFormatPaneItem.xaml
@@ -7,14 +7,16 @@
              mc:Ignorable="d" 
              d:DesignHeight="60" d:DesignWidth="300"
              MouseDoubleClick="OnMouseDoubleClick">
-    <Grid x:Name="grid">
+    <Grid x:Name="grid" ClipToBounds="True">
         <Grid.ColumnDefinitions>
             <ColumnDefinition x:Name="col1" Width="60" MaxWidth="60"/>
             <ColumnDefinition x:Name="col2" Width="*"/>
             <ColumnDefinition x:Name="col3" Width="100" MaxWidth="100"/>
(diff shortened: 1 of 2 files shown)


alexfjw:
This approach makes the buttons for each item show up far far away when the text is long.
Having to scroll all the way right feels like a bigger pain point than not being able to read overflowing text.
Would it be possible to reposition the buttons or try something along those lines?

YuPeiHenry:
The 2nd image was a suggested alternative, the 1st image shows the current implementation, which does not require scrolling to read text.

alexfjw:
Is it possible to ensure that all buttons are displayed completely?
Otherwise LGTM.

YuPeiHenry:
I have updated maxwidth to 100, so that the buttons do not go off if the width of the panel was not modified, as shown below.
image

However, if the user scales the width of the panel, the buttons will go off screen.
image

leeyh20:
Could this be tested on both PPT 2016 and 2013?

leeyh20:
@YuPeiHenry Thanks for your contribution! Usually for this project, we will have certain members that have PPT 2013, PPT 2010 so that we can test on all three versions. However since this PR only involves the UI, I think it is possible that we can merge this PR even though it is only tested on PPT 2016. It will be great if someone had 2013 though.

PR #1742, opened 2019-01-21
#1741 Picture Slides Lab: Account for scale factor of windows

### LINKED ISSUE #1741: Opening Picture Slides will overflow on smaller screens (13")
Powerpoint 2016
13" Laptop
#### Steps to replicate
1.) Click on Picture Slides
#### Expected behavior
Picture Slides window (scaled to screen size) opens, centered on screen.
#### Actual behavior
Picture Slides window opens, off-centered and overflows screen.

Picture Slides Window open
Picture Slides Window on opening

Picture Slides Window overflows the width
Pictures Slides Window overflows width of screen (and probably height as well, can't see)

ChesterSng:
Fixes #1741

**Outline of Solution**

The issue is likely to happen when opening picture slides on a small-medium size screen (13" screen). 
The lines that caused this issue are:
image

They retrieve the windows resolution (1920x1080). However, smaller devices tend to have their scaling factors to be >100%, causing the actual resolution for the PowerPoint window to be less than 1920 and 1080.
image

My solution was to add a function to calculate the scaling factor of the device and then account for it in the `systemHeight` and `systemWidth` calculation.

After fix, on a laptop with 150% scaling, picture slides lab now open, centered and does not overflow screen. 
image

I couldn't find a simpler way to get the scale factor, if anyone has a simpler way, please let me know!

### DIFF
--- a/PowerPointLabs/PowerPointLabs/PictureSlidesLab/Views/PictureSlidesLabWindow.xaml.cs
+++ b/PowerPointLabs/PowerPointLabs/PictureSlidesLab/Views/PictureSlidesLabWindow.xaml.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.Collections.ObjectModel;
 using System.Collections.Specialized;
 using System.ComponentModel;
+using System.Drawing;
 using System.Linq;
 using System.Windows;
 using System.Windows.Controls;
@@ -78,6 +79,7 @@ namespace PowerPointLabs.PictureSlidesLab.Views
         private const double StandardWindowWidth = 1200.0;
         private const double StandardWindowHeight = 700.0;
         private const double StandardPrewienGridWidth = 560.0;
+        private const float StandardDpi = 96f;
 
         # endregion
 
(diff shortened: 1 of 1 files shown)


Skaty:
Seems good. @blewjy @ChesterSng Have you guys checked that the tests passes on all scaling factors?

junkiattan:
Hey @blewjy @ChesterSng 

Nice job! Please do try to test it on other versions of Powerpoint and vary their scaling factors there as well;)

leeyh20:
@YuPeiHenry  Thanks for testing on PPT 2010!

leeyh20:
@ChesterSng Since the Standard DPI is hardcoded to 96 dpi, was wondering will it be better to try to obtain the raw DPI of the monitor?

Will https://docs.microsoft.com/en-us/uwp/api/windows.devices.display.displaymonitor
RawDpiX and RawDpiY work?

ChesterSng:
@leeyh20 Hi Yan Hwa, thanks for your suggestion!

I could not access this library inside Visual Studio :( 

However, to my knowledge, the 100% setting on the display settings corresponds to 96dpi, so hardcoding this value should work.

leeyh20:
@ChesterSng True, this API is for windows apps and not for NET Framework... LGTM! 

If we ever upgrade to NET Framework 4.7, we could support something called DPI Awareness Mode as shown in https://docs.microsoft.com/en-us/dotnet/framework/winforms/high-dpi-support-in-windows-forms . However, I think we are still running on NET Framework 4.6 so this PR is fine for now.

ChesterSng:
@leeyh20 Hi Yan Hwa, I read the article. 

Under the section **DPI Unaware**, it says that 

> Whenever these applications are run on a screen with a display scale greater than 100% (> 96 DPI), Windows will stretch the application bitmap to the expected physical size, although this results in the application being blurry.

I think 100% still refers to 96 DPI, and thus using it as the baseline and calculating the scaling factor will make the pop-up window DPI aware(?).

blewjy:
Tested on PPT 2013, working nicely at all scaling factors. PictureSlidesLab tests passed.

PR #1746, opened 2019-01-26
#1432 Close background ppts on closing of main window

### LINKED ISSUE #1432: PowerPoint open in background after closing
Currently, we have 3 "background" powerpoints:
- Shapes Lab
- Sync Lab
- Picture Slides Lab

When any of the two are open together, they will not close properly when PowerPoint is closed. This is because the codes check if 2 ppt are open in their respective onclose methods.

We should think of a way to ensure that all the powerpoints are closed properly when the user closes the main active powerpoint.

blewjy:
Fixes #1432 

From the original issue, if user were to open at least 2 of the 3 labs with background presentations (SyncLab, ShapesLab and PictureSlidesLab), closing the main PowerPoint window does not automatically close all the background ppts, you have to manually close them from the Task Manager or by clicking "Stop" in Visual Studio (assuming you're running from VS).

Also noted in the original issue, it seems that the methods that attempt to close PictureSlidesLab and ShapesLab will check if there are only 2 opened presentations before closing them. This is why when there are 2 or more of these labs open (makes 3 including the main presentation), they don't close properly. I'm not too sure why this check was included in the first place, but removing it seemed like the best way to fix this issue for now. Please update here if the check is required, we'll have to find another solution for this problem then.

**Outline of Solution**
~~1. When user closes the main ppt window, the `PresentationClose` event will be triggered. Our handler method (`ThisAddInPresentationClose()`) attempts to close PictureSlidesLab and all the other active panes in the presentation. For some unknown reason, closing the SyncLab pane when it is the only pane opened will close the background ppt as well, but when ShapesLab and SyncLab panes are open together, closing the pane doesn't not close the background ppt.
2. The next event called is `WindowDeactivate`. Our handler method (`ThisAddInApplicationOnWindowDeactivate()`) will then attempt to close SyncLab and ShapesLab background ppts here, if they are still opened.~~
**[UPDATE]**
When user closes the main ppt window, `ThisAddInPresentationClose()` will be triggered and it will attempt to close all 3 background ppts (SyncLab, ShapesLab & PictureSlidesLab), if they are open. In the case that this handler is called but the presentation that is closing is NOT the main window, execution will simply proceed and skip the closing methods.

**[UPDATE 2]**
I have removed the logic for checking for PPT2010 when closing presentation. I have found that even without that check, all the presentations and windows close as they should on PPT 2010.

Related FTs and UTs are passing. I have tested this on both PPT2016 and PPT2013, Windows 10.

### DIFF
--- a/PowerPointLabs/PowerPointLabs/ThisAddIn.cs
+++ b/PowerPointLabs/PowerPointLabs/ThisAddIn.cs
@@ -1054,6 +1054,11 @@ namespace PowerPointLabs
         {
             Trace.TraceInformation("Closing " + pres.Name + "...");
 
+
+            // We need to check if there is only one window AND the active window's presentation 
+            // has the same name as the one we are closing because it is possible for background 
+            // presentation (those without windows) to close and trigger this event as well.
+            // We only want to shut down PPTLabs if we are closing the main presentation.
             if (Application.Windows.Count == 1 &&
                 Application.ActiveWindow.Presentation.FullName == pres.FullName)
             {
--- a/PowerPointLabs/PowerPointLabs/ThisAddIn.cs
+++ b/PowerPointLabs/PowerPointLabs/ThisAddIn.cs
@@ -57,6 +57,7 @@ namespace PowerPointLabs
         private const string SlideXmlSearchPattern = @"slide(\d+)\.xml";
         private const string TempFolderNamePrefix = @"\PowerPointLabs Temp\";
         private const string ShapeGalleryPptxName = "ShapeGallery";
+        private const string SyncLabPptxName = "Sync Lab - Do not edit";
         private const string TempZipName = "tempZip.zip";
 
         private string _deactivatedPresFullName;
@@ -485,6 +486,15 @@ namespace PowerPointLabs
             }
         }
 
+        private void ShutDownImageSearchPane()
+        {
+            PictureSlidesLab.Views.PictureSlidesLabWindow pictureSlidesLabWindow = Globals.ThisAddIn.Ribbon.PictureSlidesLabWindow;
+            if (pictureSlidesLabWindow != null && pictureSlidesLabWindow.IsOpen)
+            {
+                pictureSlidesLabWindow.Close();
+            }
+        }
+
         private void RemoveTaskPanes(PowerPoint.DocumentWindow activeWindow)
         {
             if (!_documentPaneMapper.ContainsKey(activeWindow))
@@ -785,6 +795,51 @@ namespace PowerPointLabs
                 recorder.ForceStopEvent();
             }
         }
+
+        private void CloseSyncLab()
+        {
+            if (_isClosing)
+            {
+                // If sync lab open, then close it.
+                PowerPoint.Presentation syncLabPpt = GetOpenedSyncLabPresentation();
+                if (syncLabPpt != null)
+                {
+                    syncLabPpt.Close();
+                    Trace.TraceInformation("Sync Lab Ppt closed.");
+                }
+            }
+
+        }
+
+        private PowerPoint.Presentation GetOpenedSyncLabPresentation()
+        {
+            foreach (PowerPoint.Presentation presentation in Application.Presentations)
+            {
+                if (presentation.Name.Contains(SyncLabPptxName))
+                {
+                    return presentation;
+                }
+            }
+            return null;
+        }
+
+        private void CloseShapesLab()
+        {
+            // in this case, we are closing the last client presentation,
+            // therefore we can close the shape gallery
+            if (_isClosing &&
+                ShapePresentation != null &&
+                ShapePresentation.Opened)
+            {
+                if (string.IsNullOrEmpty(ShapesLabConfig.DefaultCategory))
+                {
+                    ShapesLabConfig.DefaultCategory = ShapePresentation.Categories[0];
+                }
+
+                ShapePresentation.Close();
+                Trace.TraceInformation("Shape Gallery terminated.");
+            }
+        }
         # endregion
 
         # region Powerpoint Application Event Handlers
@@ -840,21 +895,8 @@ namespace PowerPointLabs
 
             _deactivatedPresFullName = pres.FullName;
 
-            // in this case, we are closing the last client presentation,
-            // therefore we can close the shape gallery
-            if (_isClosing &&
-                Application.Presentations.Count == 2 &&
-                ShapePresentation != null &&
-                ShapePresentation.Opened)
-            {
-                if (string.IsNullOrEmpty(ShapesLabConfig.DefaultCategory))
-                {
-                    ShapesLabConfig.DefaultCategory = ShapePresentation.Categories[0];
-                }
-
-                ShapePresentation.Close();
-                Trace.TraceInformation("Shape Gallery terminated.");
-            }
+            CloseSyncLab();
+            CloseShapesLab();
         }
 
         private void ThisAddInApplicationOnWindowActivate(PowerPoint.Presentation pres, PowerPoint.DocumentWindow wn)
@@ -1099,15 +1141,6 @@ namespace PowerPointLabs
             Ribbon.RefreshRibbonControl(TimerLabText.RibbonMenuId);
             Ribbon.RefreshRibbonControl(AgendaLabText.RibbonMenuId);
             Ribbon.RefreshRibbonControl(PictureSlidesLabText.RibbonMenuId);
-        }
-
-        private void ShutDownImageSearchPane()
-        {
-            PictureSlidesLab.Views.PictureSlidesLabWindow pictureSlidesLabWindow = Globals.ThisAddIn.Ribbon.PictureSlidesLabWindow;
-            if (pictureSlidesLabWindow != null && pictureSlidesLabWindow.IsOpen && Application.Presentations.Count == 2)
-            {
-                pictureSlidesLabWindow.Close();
-            }
         }
 
         private void ThisAddInShutdown(object sender, EventArgs e)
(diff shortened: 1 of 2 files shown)


ChesterSng:
@blewjy 
https://github.com/PowerPointLabs/PowerPointLabs/blob/5f1d44f7424be193c7fd111f1ee98070b5cad8e8/PowerPointLabs/PowerPointLabs/SyncLab/Views/SyncPaneWPF.xaml.cs#L51-L62

These lines of code seems to be the reason why synclab is not closing properly when the pane is closed. 

However I'm guessing the `==2` was implemented because when opening two PowerPoint applications and both PowerPoint applications use either `SyncLab`, `PictureSlidesLab` or `ShapesLab`, one background application is opened for both of them. 

**Steps to reproduce:**
1. Open PowerPoint 
    1.1 Open `SyncLab`, copy anything to the clipboard.
2. Open another PowerPoint 
    2.1. Open `SyncLab` and close it.
3. `SyncLab` stops working on the original PowerPoint window (cannot copy, cannot apply).

This problem can be reproduced using `ShapesLab` as well.

I'm thinking if the problem can be better solved by first searching if there is another main powerpoint window open, if there is, we don't close the background window. If there is no other main powerpoint window, then we close all the background windows.

blewjy:
@ChesterSng Great catch on the `SyncPane_Closing()`!

I was able to reproduce the error from my original code, thanks for testing!

> I'm thinking if the problem can be better solved by first searching if there is another main powerpoint window open. If there is, we don't close the background window, if there is no other main powerpoint window, then we close all the background windows.

I took your suggestion here and updated the PR. I also decided to move all the background ppt closing logic into `ThisAddIn.cs`, as such I removed the `shapeStorage.Close()` call from `SyncLabPane`. Closing the pane will now only close the dialog if it's there. 

With this PR, we should be able to now close all background ppts properly for any number of open ppt windows. In the case of multiple windows, background ppts will only close when the very last ppt is closed. This check is done at the beginning of the `ThisAddInPresentationClose()` handler:
https://github.com/PowerPointLabs/PowerPointLabs/blob/f17e79bc760df9d3e9fbb98af1aab822fc413f15/PowerPointLabs/PowerPointLabs/ThisAddIn.cs#L1057-L1062

Thanks!

blewjy:
Tested and working on PPT2013

leeyh20:
Is it working on PPT 2016?

ChesterSng:
Hi @blewjy the bug seems to be fixed. I'm having a bit of trouble testing on my laptop now, will test again on 2016 later!

PowerPointLabs/PowerPointLabs/ThisAddIn.cs
```diff
@@ -1022,7 +1052,14 @@ private void ThisAddInPresentationOpen(PowerPoint.Presentation pres)
 
         private void ThisAddInPresentationClose(PowerPoint.Presentation pres)
         {
-            Trace.TraceInformation("Closing " + pres.Name);
+            Trace.TraceInformation("Closing " + pres.Name + "...");
+
+            if (Application.Windows.Count == 1 &&
```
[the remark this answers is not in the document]
blewjy: Done, thanks!

PowerPointLabs/PowerPointLabs/ThisAddIn.cs
```diff
@@ -785,6 +793,44 @@ private void BreakRecorderEvents()
...
+                syncLabPpt.Close();
+                Trace.TraceInformation("SyncLab terminated.");
+            }
+        }
+
+        private PowerPoint.Presentation GetOpenedSyncLabPresentation()
+        {
+            foreach (PowerPoint.Presentation presentation in Application.Presentations)
+            {
+                if (presentation.Name.Contains(SyncLabPptxName))
```
ChesterSng: Is there a particular reason why `.Equals()` is not used instead?

PowerPointLabs/PowerPointLabs/ThisAddIn.cs
```diff
@@ -1033,34 +1075,41 @@ private void ThisAddInPresentationClose(PowerPoint.Presentation pres)
                 ShapePresentation.Close();
             }
 
-            // special case: if we are closing ShapeGallery.pptx, no other action will be done
-            if (pres.Name.Contains(ShapeGalleryPptxName))
+            // special case: if we are closing 'ShapeGallery.pptx' or 'Sync Lab - Do not edit.pptx', no other action will be done
+            if (pres.Name.Contains(ShapeGalleryPptxName) || pres.Name.Contains(SyncLabPptxName))
```
ChesterSng: Same here

PowerPointLabs/PowerPointLabs/ThisAddIn.cs
```diff
@@ -1022,7 +1052,19 @@ private void ThisAddInPresentationOpen(PowerPoint.Presentation pres)
...
+            // presentation (those without windows) to close and trigger this event as well.
+            // We only want to shut down PPTLabs if we are closing the main presentation.
+            if (Application.Windows.Count == 1 &&
+                Application.ActiveWindow.Presentation.FullName == pres.FullName)
+            {
+                // If this current window we are closing is the last window, then PptLabs should terminate.
+                _pptLabsShouldTerminate = true;
+            }
 
             if (IsApplicationVersion2010() &&
```
leeyh20: Since there is this part of the code only for PPT 2010, we definitely need to test this PR with PPT 2010. And is this part of the code still needed? It seems to be shutting down shapes lab presentation only if presentations.count == 2

PowerPointLabs/PowerPointLabs/ThisAddIn.cs
leeyh20: Just wondering would ThisAddInPresentationClose be triggered when picture slides lab's window is closed?

PowerPointLabs/PowerPointLabs/ThisAddIn.cs
```diff
@@ -834,33 +880,16 @@ private void ThisAddInStartup(object sender, EventArgs e)
...
-            Trace.TraceInformation(pres.Name + " terminating...");
-            Trace.TraceInformation(string.Format("Is Closing = {0}, Count = {1}", _isClosing,
-                Application.Presentations.Count));
-
+            Trace.TraceInformation(pres.Name + " (Presentation) and " + wn.Caption + " (Window) deactivated.");
             _deactivatedPresFullName = pres.FullName;
-
-            // in this case, we are closing the last client presentation,
-            // therefore we can close the shape gallery
-            if (_isClosing &&
```
leeyh20: I noticed that the presentations (this whole code is not run) are now not closed OnWindowDeactivate, is it confirmed okay?

PowerPointLabs/PowerPointLabs/ThisAddIn.cs
blewjy: Yes, it would be triggered. In this case, we are only checking for ShapesLab and SyncLab's background ppts here because we don't want `ShutDownSyncLab()` and `ShutDownShapesLab()` function to be called more than once. It's alright for PictureSlidesLab because we have a way to track whether PictureSlidesLab is open through Ribbon1, so that logic is already handled in `ShutDownPictureSlidesLab()`.

The behaviour of the background ppts for all three labs differ slightly, hence all these closing methods for each lab also differ slightly. For instance, for PSL the background ppt is closed upon closing of the PSL window. For sync & shapes lab, the background ppt remains open till the end once user clicks on the lab from the ribbon.

leeyh20: That sounds good, thanks! :)

PowerPointLabs/PowerPointLabs/ThisAddIn.cs
blewjy: Yup, this is ok because `OnWindowDeactivate()` is of lower priority than `PresentationClose()` (This priority is defined in `ThisAddInStartup()`). Upon clicking 'X' on the main window, `PresentationClose()` is called and all background ppts are closed here. By the time code execution reaches `OnWindowDeactivate()`, all background ppt should be closed already.

PowerPointLabs/PowerPointLabs/ThisAddIn.cs
```diff
@@ -485,6 +486,15 @@ private void ShutDownColorPane()
             }
         }
 
+        private void ShutDownPictureSlidesLab()
+        {
+            PictureSlidesLab.Views.PictureSlidesLabWindow pictureSlidesLabWindow = Globals.ThisAddIn.Ribbon.PictureSlidesLabWindow;
```
leeyh20: This script is already `ThisAddIn` itself right? Maybe you could simplify this by changing it to just `Ribbon.PictureSlidesLabWindow` ?

PowerPointLabs/PowerPointLabs/ThisAddIn.cs
blewjy: My bad... For some reason i seemed to have missed out this part of the code :P. I'll have to investigate 2010 a little more...

PowerPointLabs/PowerPointLabs/ThisAddIn.cs
blewjy: Done!

PowerPointLabs/PowerPointLabs/ThisAddIn.cs
blewjy: This code section has been removed. Does not affect PPT2010, windows and presentations still close as they should.

blewjy:
@leeyh20 I have tested this on PPT2010 and there seem to be no problems. That part of the code that checks for PPT2010 is not necessary for the windows and presentations to close properly on PPT2010.

@YuPeiHenry could you help me double check this? I tested on my own PPT2010 but would be good if you could help to test it too!

leeyh20:
Sounds great! :)

YuPeiHenry:
Working on PPT2010.

PR #1747, opened 2019-01-27
#1272 Shows details of copied formats in tooltips

### LINKED ISSUE #1272: Show details of copied formats in tooltip
As discussed in #1224, the format list doesn't show the formats that are checked for copying.

Suggested solution is to display the copied formats in a tooltip when the format is hovered over.

YuPeiHenry:
Fixes #1272

Ready for review (no tests were added). Tested and working on PPT 2016.

**Outline of Solution**
A tooltip was added to the shape grid in the list. The tooltip's values are changed whenever the name or properties are edited. The property names are compiled recursively, so that if a child property's parent is checked, the child property's name will not be included. This reduces clutter as well.

Screenshot:
image

Example of how property names are picked:
EG.
[x] A
[x] -  A1
[x] -  A2
[ ] B
[ ] -  B1
[x] -  B2
[ ] C
[x] -  C3

Outputs: A, B2, C3

### DIFF
--- a/PowerPointLabs/PowerPointLabs/SyncLab/Views/SyncFormatPaneItem.xaml
+++ b/PowerPointLabs/PowerPointLabs/SyncLab/Views/SyncFormatPaneItem.xaml
@@ -22,7 +22,7 @@
         </Grid.ColumnDefinitions>
         <Image x:Name="imageBox" Grid.Column="0" Margin="10,10,0,10" OpacityMask="{DynamicResource {x:Static SystemColors.ControlBrushKey}}" HorizontalAlignment="Left" Width="40" Height="40"/>
         <Canvas x:Name="canvas" Grid.Column="1" >
-            <TextBlock x:Name="textBlock" Margin="0,20,0,20" MaxWidth="{Binding ActualWidth, ElementName=canvas}" HorizontalAlignment="Stretch" VerticalAlignment="Center" IsHitTestVisible="False"/>
+            <TextBlock x:Name="textBlock" Margin="0,20,0,20" MaxWidth="{Binding ActualWidth, ElementName=canvas}" HorizontalAlignment="Stretch" VerticalAlignment="Center" TextTrimming="CharacterEllipsis" IsHitTestVisible="False"/>
         </Canvas>
         <Button x:Name="pasteButton" Grid.Column="2" Margin="0,10,65,10" Click="PasteButton_Click" HorizontalAlignment="Right" MinWidth="25" MaxWidth="25" MinHeight="25" MaxHeight="25">
             <Image x:Name="pasteImage" Source="pack://siteoforigin:,,,/Resources/SyncLabPasteButton.png" Stretch="Fill"/>
--- a/PowerPointLabs/PowerPointLabs/SyncLab/Views/SyncFormatPaneItem.xaml
+++ b/PowerPointLabs/PowerPointLabs/SyncLab/Views/SyncFormatPaneItem.xaml
@@ -10,7 +10,7 @@
     <Grid x:Name="grid" ClipToBounds="True" Background="Transparent">
         <Grid.ToolTip>
             <StackPanel>
-                <TextBlock x:Name="toolTipName" FontWeight="Bold" FontSize="14" Margin="0,0,0,5"/>
+                <TextBlock x:Name="toolTipName" FontWeight="Bold" FontSize="14" Margin="0,0,0,5" MaxWidth="400" TextWrapping="Wrap"/>
                 <Border BorderBrush="Silver" BorderThickness="0,1,0,0" Margin="0,8" />
                 <TextBlock x:Name="toolTipBody"/>
             </StackPanel>
--- a/PowerPointLabs/PowerPointLabs/SyncLab/Views/SyncFormatPaneItem.xaml
+++ b/PowerPointLabs/PowerPointLabs/SyncLab/Views/SyncFormatPaneItem.xaml
@@ -7,7 +7,16 @@
              mc:Ignorable="d" 
              d:DesignHeight="60" d:DesignWidth="300"
              MouseDoubleClick="OnMouseDoubleClick">
-    <Grid x:Name="grid" ClipToBounds="True">
+    <Grid x:Name="grid" ClipToBounds="True" Background="Transparent">
+        <Grid.ToolTip>
+            <StackPanel>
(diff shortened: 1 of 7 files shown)


PowerPointLabs/PowerPointLabs/SyncLab/Views/SyncFormatPaneItem.xaml
```diff
@@ -7,15 +7,22 @@
...
+        </Grid.ToolTip>
         <Grid.ColumnDefinitions>
             <ColumnDefinition x:Name="col1" Width="60" MaxWidth="60"/>
             <ColumnDefinition x:Name="col2" Width="*"/>
             <ColumnDefinition x:Name="col3" Width="100" MaxWidth="100"/>
         </Grid.ColumnDefinitions>
         <Image x:Name="imageBox" Grid.Column="0" Margin="10,10,0,10" OpacityMask="{DynamicResource {x:Static SystemColors.ControlBrushKey}}" HorizontalAlignment="Left" Width="40" Height="40"/>
         <Canvas x:Name="canvas" Grid.Column="1" >
-            <TextBlock x:Name="textBlock" Margin="0,20,0,20" MaxWidth="{Binding ActualWidth, ElementName=canvas}" HorizontalAlignment="Stretch" VerticalAlignment="Center" TextTrimming="CharacterEllipsis"/>
+            <TextBlock x:Name="textBlock" Margin="0,20,0,20" MaxWidth="{Binding ActualWidth, ElementName=canvas}" HorizontalAlignment="Stretch" VerticalAlignment="Center" IsHitTestVisible="False"/>
```
blewjy: Any reason why you removed `TextTrimming` here? If I enter a very very long name for the SyncLab entry, the title view on the pane no longer has the ellipses to signify that it's a long title, it just gets truncated.

Also, if I enter a very very long title, the tooltips box gets stretched all the way across the whole screen. While this doesn't affect any functionality, I think it will look nicer if we restricted the tooltips box to a max width that looks nice.

Screenshot of the stretched tooltips:
tooltip-box-too-long

YuPeiHenry:
@blewjy I have included the tooltip update on editing for any SyncLab formats.

YuPeiHenry:
Thanks for the suggestions. I think the tooltip part is not too difficult, like suggested here https://stackoverflow.com/questions/14033238/wpf-setting-tooltip-maxwidth. As for the duplicate names, I did not realise this problem.

Maybe putting the root is sufficient. I will add this in.

YuPeiHenry:
I have added textwrap with maxwidth 400 to deal with long names. I have also added the root names.

blewjy:
@YuPeiHenry LGTM! tested on 2013. Thanks!

leeyh20:
LGTM. Can someone provide a screenshot of the final tooltip to show how it looks like? Would it be difficult to add a test case for this?

PowerPointLabs/PowerPointLabs/SyncLab/Views/SyncFormatPaneItem.xaml.cs
```diff
@@ -123,10 +131,54 @@ public String Text
...
+            }
+        }
+
+        #endregion
+
+        #region Helper Functions
+
+        private void UpdateToolTipBody()
+        {
+            String toolTipBodyText = "";
```
alexfjw: Don't think it really makes a difference since tooltips are likely to be short but maybe its better to use a StringBuilder instead? (same for line 158)

YuPeiHenry: Good idea!

YuPeiHenry:
@leeyh20 I believe it won't be difficult to implement a test. What should be the expected output to compare against? (hardcoded strings, or string that reference current `Sync` categories?)

YuPeiHenry:
@leeyh20 I have added tests for the tooltips of SyncLab.

blewjy:
Tested on PPT2010, feature is working and tests passing.

PowerPointLabs/Test/UnitTest/SyncLab/SyncLabFormatPaneItemTest.cs
```diff
@@ -0,0 +1,129 @@
...
+        private FormatTreeNode[] SelectFormats(FormatTreeNode[] formats, string[] formatNames)
+        {
+            foreach (string name in formatNames)
+            {
+                getEndNode(formats, name).IsChecked = true;
+            }
+            return formats;
+        }
+
+        private FormatTreeNode getEndNode(FormatTreeNode[] formats, string name)
```
leeyh20: Just some styling issues, I believe the function name should be in caps for the first letter?

PowerPointLabs/Test/UnitTest/SyncLab/SyncLabFormatPaneItemTest.cs
```diff
@@ -0,0 +1,129 @@
...
+                if (nodes[index].Name.Equals(name))
+                {
+                    return index;
+                }
+                index--;
+            }
+            return index;
+        }
+
+        private void testFollowingNodes(string[] nodeNames)
```
leeyh20: This test too.

PowerPointLabs/Test/UnitTest/SyncLab/SyncLabFormatPaneItemTest.cs
leeyh20: Maybe will be good to also provide some function comments to describe what the functions does

PowerPointLabs/PowerPointLabs/SyncLab/Views/SyncFormatPanelItemStub.cs
```diff
@@ -0,0 +1,18 @@
+﻿using System.Drawing;
+
+namespace PowerPointLabs.SyncLab.Views
+{
+    class SyncFormatPaneItemStub : SyncFormatPaneItem
+    {
+        public SyncFormatPaneItemStub(FormatTreeNode[] formats) :
```
leeyh20: Nice stub!

PowerPointLabs/Test/UnitTest/SyncLab/SyncLabFormatPaneItemTest.cs
YuPeiHenry: Yes the first letter should be caps. I will also add comments since it might not be immediately obvious what the function does.

YuPeiHenry:
@leeyh20 I have fixed the naming violations and added a short description to some functions. Apologies for the slow response.

PR #1748, opened 2019-01-28
#1604 Disable SyncLab/ShapeLab buttons when nothing or illegal object is selected

### LINKED ISSUE #1604: ShapesLab / SyncLab: Add button should be disabled when nothing/illegal object is selected
As discussed in #1566, we will discuss whether the button on the pane should be disabled or not when nothing / something illegal is selected.  Related to #1573, where we decided that for ribbon buttons, we will grey them out.
Popups are more expressive, giving more information to the user while disabled buttons are less distracting. We could use tooltips with disabled buttons to give the same information to the user.

SyncLab might need to change too to standardise this.

@alexfjw @junkiattan We can discuss more in this issue.

ChesterSng:
Fixes #1604 

**In Progress**
1. Need to change some tests
2. Need to standardise disabled button appearances

**Outline of Solution**

Added an event handler in `SyncPane` and `ShapesLab` to handle selection change. When nothing/illegal objects are selected, the `Copy` and `Add` button will be disabled. The tooltip on mouse over will still show as below. 

image


PowerPointLabs/PowerPointLabs/ShapesLab/CustomShapePane.cs
```diff
@@ -1685,5 +1747,7 @@ private void InitToolTipControl()
             toolTip1.SetToolTip(addShapeButton, ShapesLabText.AddShapeToolTip);
         }
         #endregion
+
```
ChesterSng: Remove these lines

blewjy:
@ChesterSng I think if we can have a popup to alert the user that they were attempting to add an illegal object will be good. If we were just to disable it for illegal objects the user might just assume that the lab is not working as it should. Having a response for illegal object should be more informative for the user.

ChesterSng:
Thanks @blewjy for your suggestion! I agree with you that disabling the button even when there is a selection can confuse users into thinking that the lab doesn't work. 

I have changed the behaviour of the buttons:
1. Button is enabled and litted up when there is a selection (refers to any selection within the slide) 
2. Button is disabled and greyed out when there are no selections. 

Error message will pop up when the user tries to use the lab with an illegal object as before.

PR #1760, opened 2019-01-31
#1604 Disable SyncLab/ShapesLab button when there is no selection

### LINKED ISSUE #1604: ShapesLab / SyncLab: Add button should be disabled when nothing/illegal object is selected

ChesterSng:
Fixes #1604 

**Outline of Solution**

Added an event handler in `SyncPane` and `ShapesLab` to handle selection change. When **no** objects are selected, the `Copy` and `Add` button will be disabled. The tooltip on mouse over will still show as below (mouse hidden in screenshot). 

1. This is when shapes add button is disabled.
shapesbuttondisabled

2. This is when shapes add button is enabled when there is a selection.
shapesbuttonenabled

3. This is when the mouse hovers over the disabled button.
shapesbuttondisabledwithtooltip

1. This is when sync copy button is disabled. 
syncbuttondisabled

2. This is when sync copy button is enabled when there is a selection.
syncbuttonenabled

3. This is when the mouse hovers over the disabled button.
syncbuttondisabledtooltip


PowerPointLabs/PowerPointLabs/TextCollection/ShapesLabText.cs
```diff
@@ -69,6 +69,6 @@ internal static class ShapesLabText
         public const string FolderDialogDescription = "Select the directory that you want to use as the default.";
         public const string ErrorFolderNonEmpty = "Please select an empty folder as default saving folder.";
 
-        public const string AddShapeToolTip = "Adds a shape to Shapes Lab.";
+        public const string AddShapeToolTip = "Adds a shape to Shapes Lab. \nStart by selecting a shape.";
```
blewjy: Since you changed the original tooltip, when the button is active, it still says to "Start by selecting a shape". Perhaps we should split into a `ButtonEnabledToolTip` and `ButtonDisabledToolTip`.

PowerPointLabs/PowerPointLabs/ShapesLab/CustomShapePane.cs
```diff
@@ -1677,6 +1712,41 @@ private void AddShapeButton_Click(object sender, EventArgs e)
...
+                    {
+                        toolTip1.Show(ShapesLabText.AddShapeToolTip, ctrl, ctrl.Width / 2, ctrl.Height / 2);
+                        toolTip1.Tag = ctrl;
+                        _toolTipShown = true;
+                    }
+                }
+            }
+            else
+            {
+                ctrl = toolTip1.Tag as Control;
```
blewjy: Is it possible not to reuse this `ctrl` variable here? Issue is when I first read through this part of the code, I was wondering why you are checking `ctrl != null` again on line 1742 when you already did that in line 1727. This may cause unnecessary confusion to other developers in the future.

PowerPointLabs/PowerPointLabs/ShapesLab/CustomShapePane.cs
```diff
@@ -1677,6 +1712,41 @@ private void AddShapeButton_Click(object sender, EventArgs e)
 
             AddShapeFromSelection(selection, addIn);
         }
+
+        // A disabled button cannot respond to any events.
+        // Thus we register the event to the pane and when the mouse moves over
+        // the button, the tool tip will display.
+        private void CustomShapePane_MouseMove(object sender, MouseEventArgs e)
```
YuPeiHenry: The button can be embedded inside a grid, with hitTest enabled and has the tooltip for the button. This is easier done with WPF (in fact I see that tooltip service can remove the need for a grid), not sure if its difficult to do currently.

PowerPointLabs/PowerPointLabs/ThisAddIn.cs
```diff
@@ -960,7 +961,7 @@ private void ThisAddInSelectionChanged(PowerPoint.Selection sel)
                 PowerPoint.Shape sh = null;
                 try
                 {
-                    sh = sel.ShapeRange[1];
+                    sh = sel.ShapeRange[1];
```
YuPeiHenry: Trailing whitespace

PowerPointLabs/PowerPointLabs/ThisAddIn.cs
```diff
@@ -978,6 +979,21 @@ private void ThisAddInSelectionChanged(PowerPoint.Selection sel)
 
             }
 
+
+            // When there is no selection on the slide, disable the add/copy buttons on
+            // customshapepane and syncpane respectively
+            if (GetActivePane(typeof(SyncPane)) != null)
```
YuPeiHenry: In C# 8, you can do optional chaining. Here, the entire block can be simply:
`(GetActivePane(typeof(SyncPane))?.Control as SyncPane).UpdateOnSelectionChange();`
I see many existing occurrences of this problem, might open an issue on this for codeQuality. Need your input on this.

PowerPointLabs/PowerPointLabs/ThisAddIn.cs
```diff
@@ -978,6 +979,21 @@ private void ThisAddInSelectionChanged(PowerPoint.Selection sel)
...
+
+            // When there is no selection on the slide, disable the add/copy buttons on
+            // customshapepane and syncpane respectively
+            if (GetActivePane(typeof(SyncPane)) != null)
+            {
+                SyncPane syncPane = GetActivePane(typeof(SyncPane)).Control as SyncPane;
+                syncPane.UpdateOnSelectionChange();
+            }
+
+            if (GetActivePane(typeof(CustomShapePane)) != null)
```
YuPeiHenry: Can use optional chaining here.

PowerPointLabs/PowerPointLabs/ShapesLab/CustomShapePane.cs
YuPeiHenry: Might want to ignore this if its too difficult, current implementation is fine too,

PowerPointLabs/PowerPointLabs/ThisAddIn.cs
leeyh20: We are currently using [ .NET Framework 4.6 for dev release ](https://github.com/PowerPointLabs/PowerPointLabs/blob/master/doc/ProjectSetUp.md) which might be using only C# 6.0. I'm not sure if upgrading the framework version is difficult or not, we did not try yet.

PowerPointLabs/PowerPointLabs/ShapesLab/CustomShapePane.cs
leeyh20: I think @YuPeiHenry's suggestion is to have a different implementation for this tooltip of the disabled button, so that you won't have to constantly track the movement of the mouse (which seems inefficient in a way).

leeyh20: I also think it could be easier to do it in WPF so... might want to change this implementation when moving over to WPF.

ChesterSng: @leeyh20 Yeah the `SyncLab` implementation is much simpler!

PowerPointLabs/PowerPointLabs/ThisAddIn.cs
```diff
@@ -978,6 +979,21 @@ private void ThisAddInSelectionChanged(PowerPoint.Selection sel)
 
             }
 
+
+            // When there is no selection on the slide, disable the add/copy buttons on
+            // customshapepane and syncpane respectively
+            if (GetActivePane(typeof(SyncPane)) != null)
+            {
+                SyncPane syncPane = GetActivePane(typeof(SyncPane)).Control as SyncPane;
+                syncPane.UpdateOnSelectionChange();
```
leeyh20: Why not pass the selection `sel` to this method as a parameter too?

PowerPointLabs/PowerPointLabs/SyncLab/Views/SyncPaneWPF.xaml.cs
```diff
@@ -72,6 +74,21 @@ public int FormatCount
             }
         }
 
+        public void UpdateCopyButtonEnabledStatus()
+        {
+            Selection selection = this.GetCurrentSelection();
```
leeyh20: Why not use a passed selection of shapes rather then try to get the current selection again?

leeyh20:
Could you also try to make tests that check whether the buttons are enabled / disabled?

PowerPointLabs/PowerPointLabs/ThisAddIn.cs
ChesterSng: This is because the selection can be retrieved inside `SyncPaneWPF.xaml.cs` but not in `ShapesLab`. I can change it such that both methods take in `sel`!

ChesterSng:
@leeyh20 Okay sure!

PowerPointLabs/PowerPointLabs/ThisAddIn.cs
leeyh20: Thanks! Hopefully nothing goes wrong with that! Just felt it is slightly better to use `sel` instead of `this.GetCurrentSelection()` because we have received #1543  bug reports before about `this.GetCurrentSelection()` sometimes not working in certain environments (not sure why...)

ChesterSng:
@leeyh20 I'm trying to write the tests to get the status of the button when there is a selection/no selection etc, but I'm having some trouble accessing the status of the button. 

` public bool GetCopyButtonEnabledStatus()
        {
            bool isCopyButtonEnabled = false;
            if (_pane != null)
            {
                UIThreadExecutor.Execute(() =>
                {
                    isCopyButtonEnabled = _pane.SyncPaneWPF1.copyButton.IsEnabled;
                });
            }
            return isCopyButtonEnabled;
        }
`

Somehow the variable isCopyButtonEnabled does not get updated after the execute call (always returns false).

leeyh20:
@ChesterSng Besides adding wait, have you tried doing the check for `isCopyButtonEnabled` directly without putting it inside UIThreadExecutor.Execute? Not sure if it will work
The `ShapesLabTest` is it failing randomly?

leeyh20:
@ChesterSng If the tests continue to fail, it might be because you need to clean the project or clear the cache or something. Or do ask another team member to help to test your branch !

ChesterSng:
@leeyh20 If I try to access the status without using the `UIThreadExecutor`, it will throw an exception saying that I'm trying to access something that belongs to another thread. 

The `ShapesLabTest` is failing all the time. I will clean and rebuild the project to try on PowerPoint 2013.

ChesterSng:
@YuPeiHenry Need help with testing on powerpoint 2010!

ChesterSng:
@leeyh20 
> @ChesterSng If the tests continue to fail, it might be because you need to clean the project or clear the cache or something. Or do you think the tests are failing because of another PR to the dev-release branch?
`ShapesLabTest` now passes on 2013

PowerPointLabs/Test/FunctionalTest/BaseFunctionalTest.cs
```diff
@@ -81,8 +81,8 @@ protected static void CheckIfClipboardIsRestored(Action action, int actualSlideN
             ShapeRange newShape = actualSlide.Shapes.Paste();
 
             // Check if pasted shape is the same as the shape added to clipboard originally
-            Assert.AreEqual(shapeNameToBeCopied, newShape.Name);
-            Assert.AreEqual(shapeToBeCopied.Count, newShape.Count);
+            //Assert.AreEqual(shapeNameToBeCopied, newShape.Name);
```
leeyh20: @ChesterSng Does the tests pass if you uncomment these lines?

leeyh20: This seems to be affecting another test.

ChesterSng: @leeyh20 oops I forgot to uncomment them. Uncommented and ShapesLabTest and SyncLabTest no problem!

YuPeiHenry:
Tested and passing on PPT2010.

PowerPointLabs/PowerPointLabs/PowerPointLabs.csproj
```diff
@@ -2073,6 +2073,9 @@
   <ItemGroup>
     <None Include="Resources\FillSlide.png" />
   </ItemGroup>
+  <ItemGroup>
+    <Content Include="Resources\AddToCustomShapesDisabled.png" />
```
leeyh20: Probably this should be None Include instead, which is causing the deployment error. Still testing.

PR #1761, opened 2019-01-31
#195 Increase resolution of images in ZoomLab

### LINKED ISSUE #195: Increase resolution of 'slide as image' conversion
_From [dam...@gmail.com](https://code.google.com/u/114919422028123491955/) on February 15, 2014 14:14:26_

When we convert a slide to an image (e.g. zoom in), the resolution of the image determines the quality of the animation. saving the image as a file first and inserting that seems to give me a better quality than using the clipboard method 'paste as picture'
Apparently, the resolutions can be changed http://support.microsoft.com/kb/827745

_Original issue: http://code.google.com/p/powerpointlabs/issues/detail?id=195_

blewjy:
Fixes #195 

Issue was regarding resolution of images in `ZoomLab`, particularly 'Drill Down' and 'Step Back'. During the animation, it can be seen quite obviously that the quality of the images during the animation has been reduced greatly. This was due to the fact that some of the images during the animation were generated by copying and pasting the slides as images.
 
### **Outline of Solution**
This PR changes the copying and pasting to export and insert, as suggested in the issue. We already have an existing export slide as `.png` image method in `GraphicsUtil`, so that is what I used. The slide is exported as an image to a temporary destination, then immediately added back into the target slide as an image. This results in a greater quality image as opposed to just copying and pasting. Greater quality image in turn gives a smoother-looking animation. This is more evident in higher resolution images.

Screenshots of result below. 'Original' refers to the picture before the resolution drop. 'Before fix' refers to the picture as a result of copy and paste. 'After fix' refers to the picture as a result of exporting and inserting. 

**Original vs. Before fix:**
original  before

**Original vs. After fix:**
original  after

Tested and working on PPT2013.

### DIFF
--- a/PowerPointLabs/PowerPointLabs/Utils/FileDir.cs
+++ b/PowerPointLabs/PowerPointLabs/Utils/FileDir.cs
@@ -139,6 +139,16 @@ namespace PowerPointLabs.Utils
             return CopyFolder(oldPath, newPath) && DeleteFolder(oldPath);
         }
 
+        public static string GetTemporaryPngFilePath()
+        {
+            string tempDirectory = Path.GetTempPath();
+            Directory.CreateDirectory(tempDirectory);
+
+            string randomFileName = "pptlabs_" + Path.GetRandomFileName();
+            string randomFilePath = Path.Combine(tempDirectory, randomFileName);
+            return Path.ChangeExtension(randomFilePath, "png");
+        }
+
         /// <summary>
         /// This function sets attribute of all files in a folder and its sub-folder 
         /// to normal.
--- a/PowerPointLabs/PowerPointLabs/ZoomLab/AutoZoom.cs
+++ b/PowerPointLabs/PowerPointLabs/ZoomLab/AutoZoom.cs
@@ -457,17 +457,28 @@ namespace PowerPointLabs.ZoomLab
 
         private static PowerPoint.Shape AddSlideAsShape(PowerPointSlide slideToAdd, PowerPointSlide targetSlide)
         {
-            // Export the slide as .png to a temporary location, then add it to shapes.
-            // This yields a higher quality image compared to copy-pasting slide as image.
-            string tempFilePath = FileDir.GetTemporaryPngFilePath();
-            GraphicsUtil.ExportSlide(slideToAdd, tempFilePath);
-            PowerPoint.Shape slideAsShape = targetSlide.Shapes.AddPicture2(tempFilePath,
-                                                                             Microsoft.Office.Core.MsoTriState.msoFalse,
-                                                                             Microsoft.Office.Core.MsoTriState.msoTrue,
-                                                                             0,
-                                                                             0);
-            FileDir.DeleteFile(tempFilePath);
-            return slideAsShape;
+            try
+            {
+                // Export the slide as .png to a temporary location, then add it to shapes.
+                // This yields a higher quality image compared to copy-pasting slide as image.
+                string tempFilePath = FileDir.GetTemporaryPngFilePath();
+                GraphicsUtil.ExportSlide(slideToAdd, tempFilePath);
+                PowerPoint.Shape slideAsShape = targetSlide.Shapes.AddPicture2(tempFilePath,
+                                                                                 Microsoft.Office.Core.MsoTriState.msoFalse,
+                                                                                 Microsoft.Office.Core.MsoTriState.msoTrue,
+                                                                                 0,
+                                                                                 0);
+                FileDir.DeleteFile(tempFilePath);
+                return slideAsShape;
+            }
+            catch (Exception)
+            {
+                // It is possible that there could permissions-related issues that cause user to be unable to create/delete files.
+                // In that case, we proceed with copy-pasting the slide as image.
+                slideToAdd.Copy();
+                PowerPoint.Shape slideAsShape = targetSlide.Shapes.PasteSpecial(PowerPoint.PpPasteDataType.ppPastePNG)[1];
+                return slideAsShape;
+            }
         }
     }
 }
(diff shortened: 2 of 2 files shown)


blewjy:
There was also a suggestion by Prof @damithc to include a context menu option `Save Slide as Image` (I believe that would go under Shortcuts Lab). I intend to implement that in a separate issue and PR.

PowerPointLabs/PowerPointLabs/Utils/FileDir.cs
```diff
@@ -139,6 +139,16 @@ public static bool MoveFolder(string oldPath, string newPath)
             return CopyFolder(oldPath, newPath) && DeleteFolder(oldPath);
         }
 
+        public static string GetTemporaryPngFilePath()
+        {
+            string tempDirectory = Path.GetTempPath();
+            Directory.CreateDirectory(tempDirectory);
+
+            string randomFileName = "pptlabs_" + Path.GetRandomFileName();
```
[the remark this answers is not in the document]
blewjy: Well, it's not a necessity, but there are a bunch of other files in the `AppData/Local/Temp` folder that do not necessarily belong to PPTLabs, and it would be good to differentiate. In fact, it seems that PictureSlidesLab also saves some temp files here under the folder name `pptlabs_pictureSlidesLab`, so it's probably good to standardise.

leeyh20:
> 
> 
> There was also a suggestion by Prof @damithc to include a context menu option `Save Slide as Image` (I believe that would go under Shortcuts Lab). I intend to implement that in a separate issue and PR.

Maybe u would like to update your original PR message to say that you are only fixing the issue partially?

leeyh20:
Instead of Shortcuts lab, the `Save Slide as Image ` could also be part of Save Lab? But do check with Prof @damithc

PowerPointLabs/PowerPointLabs/Utils/FileDir.cs
```diff
@@ -139,6 +139,16 @@ public static bool MoveFolder(string oldPath, string newPath)
             return CopyFolder(oldPath, newPath) && DeleteFolder(oldPath);
         }
 
+        public static string GetTemporaryPngFilePath()
+        {
+            string tempDirectory = Path.GetTempPath();
+            Directory.CreateDirectory(tempDirectory);
```
leeyh20: Not sure if it is possible, but could an exception be caught if there is insufficient permissions to create a directory or a new file? There could be PPTLabs users that are in more protected environments or maybe they did not run powerpoint with administrator permissions.

leeyh20: Thus if they do not have sufficient permissions, copying and pasting actually works better for them.

leeyh20:
Do test with 2016 since everyone has it! :)

damithc:
> Instead of Shortcuts lab, the `Save Slide as Image ` could also be part of Save Lab? But do check with Prof @damithc

Can we cross-list the feature in both?

PowerPointLabs/PowerPointLabs/Utils/FileDir.cs
blewjy: Good point, I didn't think about that! Ok, I have added a try catch for exceptions as a result of creating/deleting files. The execution then proceeds with copy and paste.

PowerPointLabs/PowerPointLabs/ZoomLab/AutoZoom.cs
```diff
@@ -460,5 +454,20 @@ private static void SelectAllShape(List<PowerPoint.Shape> shapes)
...
+            // Export the slide as .png to a temporary location, then add it to shapes.
+            // This yields a higher quality image compared to copy-pasting slide as image.
+            string tempFilePath = FileDir.GetTemporaryPngFilePath();
+            GraphicsUtil.ExportSlide(slideToAdd, tempFilePath);
+            PowerPoint.Shape slideAsShape = targetSlide.Shapes.AddPicture2(tempFilePath,
+                                                                             Microsoft.Office.Core.MsoTriState.msoFalse,
+                                                                             Microsoft.Office.Core.MsoTriState.msoTrue,
+                                                                             0,
+                                                                             0);
+            FileDir.DeleteFile(tempFilePath);
```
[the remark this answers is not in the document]
blewjy: Addressed.

blewjy:
> Can we cross-list the feature in both?

Since Save Lab already has a feature to let users save selected slides as a new PPT, I think it's an appropriate place to put this new feature. I guess a possible solution is to put it under Save Lab, but to add it as a context menu option when a number of slides are selected and user right clicks.

Similar to how Paste Lab also shows up on the context menu here.

save-slide-as-image-context-menu

PowerPointLabs/PowerPointLabs/Utils/FileDir.cs
leeyh20: I only thought about that because of the numerous bug reports filed regarding insufficient permissions on google groups 🔢

PowerPointLabs/PowerPointLabs/ZoomLab/AutoZoom.cs
```diff
@@ -460,5 +454,31 @@ private static void SelectAllShape(List<PowerPoint.Shape> shapes)
...
+                // Export the slide as .png to a temporary location, then add it to shapes.
+                // This yields a higher quality image compared to copy-pasting slide as image.
+                string tempFilePath = FileDir.GetTemporaryPngFilePath();
+                GraphicsUtil.ExportSlide(slideToAdd, tempFilePath);
+                PowerPoint.Shape slideAsShape = targetSlide.Shapes.AddPicture2(tempFilePath,
+                                                                                 Microsoft.Office.Core.MsoTriState.msoFalse,
+                                                                                 Microsoft.Office.Core.MsoTriState.msoTrue,
+                                                                                 0,
+                                                                                 0);
+                FileDir.DeleteFile(tempFilePath);
```
leeyh20: Should FileDir.DeleteFile have its own try catch?
I'm worried that if everything succeeds except delete file, then there will be a duplicate of the shape on the slide. This is because the adding of the picture to the slide has already succeeded, then deletefile throws an exception but yet the code will go to the "catch" block and then run the copy and paste code. This results in two shapes on the slide?

blewjy: Hmm, I was operating under the assumption that if the user was able to create a file, user should also be able to delete that same file. But I guess it wouldn't hurt to be a little safer. Made the changes as suggested!

PR #1779, opened 2019-02-12
#1778 Port ColorsLab to WPF

### LINKED ISSUE #1778: Port ColorsLab over to WPF
Currently, ColorsLab and ShapesLab are the only labs that are still using Windows Forms for the UI while the rest of the Labs with panes are using WPF. This issue is for porting over ColorsLab code into WPF. Mentioned in #478.

There is a change in behaviour of ColorsLab that I would like to propose, and it has been discussed offline with a couple of us but I would like to put it here for all to comment:

**Current behaviour:**
Currently, to apply a colour to a shape (be it font, fill or outline), user has to click and drag from one of the three buttons at the top of the pane to their desired colour to apply. I thought that this is not very intuitive, and if it weren't for the docs, it would take awhile for me to figure out how this works. 
colors-lab-8

**Proposed:**
I think to make it easier and more straightforward, the user will simply select their desired colour, then click on one of the buttons to apply the colour to the shape. This should be more intuitive than the eyedropper function. We can then add a fourth button in the top row for users to select colours with the eyedropper as well. (Eyedropper not in the screenshot below)
colors-lab-new

If there are any concerns, do mention here! Thanks

blewjy:
Fixes #1778. Ready for review.

~As mentioned in the issue, this PR has the proposed functionality for the buttons, which is where user selects the colour before clicking the button to apply the color to the shape. This feature can be easily changed back to the old behaviour if that is desired.~

This PR will port ColorsLab from the WindowsForms to the WPF so as to keep the implementation consistent throughout the labs.

- [x] Base pane and UI in WPF
- [x] Main selected color rectangle & color picker
- [x] Button functionality (click-and-drag to eyedrop color)
- [x] Sliders + matching colors
- [x] Drag-and-drop to favourite colors
- [x] Save, load, clear & reload favourtie colors
- [x] Eyedropper tool
- [x] Rewrite tests
- [x] Color Information Dialog

This new implementation of ColorsLab follows the old implementation very closely, with some minor exceptions. 

**Previous implementation:** To change color of the main rectangle, either click to open color picker dialog or click and drag to eyedrop.
**New implementation:** Instead of dragging the rectangle to eyedrop, there is now a new eyedropper button on the top row. This allows user to drag the main color rectangle to the favourite colors panel as well.

Aside from a couple of bug fixes, other main features remain the same.


blewjy:
> Actually, there is an alternative we can consider, which might alleviate the unintuitiveness of the feature UX.
We can lock into 'magnifying glass + live preview' mode upon clicking on the button (instead of requiring to click and drag). We can go back to normal mode upon clicking the button again, or picking a color, or pressing `ESC` key.

From the issue, there was a suggestion from Prof to change the button functionality to a click to "lock-into-mode" and click again to select, instead of the initial drag-to-pick color. 

After much research and testing, I found that while allowing the user to click to go into the color picking mode was more intuitive, it might not be a feasible feature to implement, at least not without some hacky solutions. This is because after the user locks into that mode, clicking on a location outside of PowerPoint causes PowerPoint to lose focus. While the colour can still be successfully picked, it makes for a pretty bad experience for the user. 

As such, I will be reverting back to the old behaviour for now, which is the click drag and release behaviour. Releasing mouse button outside of PowerPoint does not cause it to lose focus, hence this method works.

ChesterSng:
Hi @blewjy great job on this PR! Colors Lab FT passing on PowerPoint 2016. 

Putting the title of the icons in the colorslab pane will help users understand the icon quicker, instead of using the tooltip only.

Found a few bugs:  

**1.) Error when dragging the color from color box into textbox**
_Steps to reproduce:_
1. Select and drag from the colorpicker (or any color box)
2. Release inside any textbox (even shapes that have textbox)
catastrophic failure

**2.) Main colour changes even when eye dropper tool is not used properly**
_Steps to reproduce:__
1. Change the main color.
2. Click on the `Eye Dropper Tool`, don't drag.
3. Error appears but main colour changed.

_Main color will be changed to the last color that was applied to it successfully by the eyedropper tool_

Compared to the behaviour of other tools, they do not change the colour of the text/shape/line when there is an error. 
main colour change with error

blewjy:
Hey @ChesterSng, thanks for testing!

**1.) Error when dragging the color from color box into textbox**
It seems that this error is due to PowerPoint not having the proper support for such drag and drop operations from an external source into its textbox/shapes. I got around this by catching the exception thrown and inserting it into the `TextRange` of the textbox/shape instead. [Link for more info](https://social.msdn.microsoft.com/Forums/vstudio/en-US/16fbc569-fd8b-4447-8d4e-c524954dcb78/wpf-listbox-drop-and-powerpoint?forum=wpf)

**2.) Main colour changes even when eye dropper tool is not used properly**
Fixed.

I also added labels to the buttons as suggested. Thanks!

YuPeiHenry:
@blewjy Tested and working on PPT2010 (including FT). There appears to be a minor issue with save/load on favorite colors. The file extension is unspecified.

Original:
image

Current:
image

blewjy:
@YuPeiHenry good catch! Fixed it!

PowerPointLabs/PowerPointLabs/DataSources/ColorDataSource.cs
```diff
@@ -259,7 +259,61 @@ public HSLColor ThemeColorTen
...
+            get
+            {
+                return themeColorElevenValue;
+            }
+            set
+            {
+                if (value != this.themeColorElevenValue)
+                {
+                    this.themeColorElevenValue = value;
+                    OnPropertyChanged("themeColorEleven");
```
YuPeiHenry: Seems like there are a lot of colors. I'm not sure if its possible (and not too difficult) to change this to use an array, but it would be good.

PowerPointLabs/PowerPointLabs/ActionFramework/ColorsLab/ColorsLabActionHandler.cs
```diff
@@ -12,11 +13,17 @@ class ColorsLabActionHandler : ActionHandler
...
         {
-            CustomTaskPane colorPane = 
-                this.RegisterTaskPane(typeof(ColorPane), ColorsLabText.TaskPanelTitle);
-            if (colorPane != null)
+            this.RegisterTaskPane(typeof(ColorsLabPane), ColorsLabText.TaskPanelTitle);
+            CustomTaskPane colorPane = this.GetTaskPane(typeof(ColorsLabPane));
+            // if currently the pane is hidden, show the pane
+            if (!colorPane.Visible)
             {
-                colorPane.Visible = !colorPane.Visible;
```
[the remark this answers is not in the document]
blewjy: This is required to hide and show the colorPane when user clicks on the main ribbon button

PowerPointLabs/PowerPointLabs/ColorsLab/HSLColor.cs
```diff
@@ -44,6 +44,12 @@ public string ToRGBString()
             return String.Format("R: {0:#0.##} G: {1:#0.##} B: {2:#0.##}", color.R, color.G, color.B);
         }
 
+        public string ToHEXString()
+        {
+            Color color = (Color)this;
+            return "#" + color.R.ToString("X2") + color.G.ToString("X2") + color.B.ToString("X2");
```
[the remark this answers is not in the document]
blewjy: Good idea! I have refactored this into `ColorHelper.cs`. Makes more sense this way, thanks for the suggestion!

PowerPointLabs/PowerPointLabs/DataSources/ColorDataSource.cs
blewjy: This `ColorDataSource.cs` is mostly code from the old implementation of ColorsLab which i reused. Because all of them are using bindings with the UI and calling that `OnPropertyChanged` method, I don't think using a simple array will work for individual objects, i.e. they won't get updated as expected. 

One possible solution we can look into is using `ObservableCollections`, however I have not researched much into this just yet. Perhaps this could be something to consider refactoring in the near future, but as of now I don't think this affects functionality much!

blewjy:
Tested and working on PPT2013. This PR is ready for mentors' review.

junkiattan:
@alexfjw @leeyh20 

Code LGTM. On my end, I have not touched WPF to the extent of this PR.
Great initiative to maintain consistency with the labs though. Im sure this would make future developments alot more easier to code and maintain;)

PowerPointLabs/PowerPointLabs/ColorsLab/HSLColor.cs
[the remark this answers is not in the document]
leeyh20: Instead of still using `ToString("X2")`

PowerPointLabs/PowerPointLabs/Converters/ColorPane/SelectedColorShiftLuminosityByFactor.cs
```diff
@@ -0,0 +1,32 @@
...
+    [ValueConversion(typeof(HSLColor), typeof(string))]
+    class SelectedColorShiftLuminosityByFactor : IValueConverter
+    {
+        public static SelectedColorShiftLuminosityByFactor Instance = new SelectedColorShiftLuminosityByFactor();
+
+        public object Convert(object value, Type targetType, object parameter, System.Globalization.CultureInfo culture)
+        {
+            if (value == null)
+            {
+                Color color = (HSLColor)value;
```
leeyh20: Was wondering if `value == null` then converting it to `HSLColor` wouldn't it be still null? Or is there a default value to the `HSLColor`?

PowerPointLabs/PowerPointLabs/FunctionalTestInterface.Impl/Controller/ColorsLabController.cs
```diff
@@ -16,114 +23,364 @@ class ColorsLabController : MarshalByRefObject, IColorsLabController
...
+            }
+
+            if (_pane != null && rect != null)
+            {
+                UIThreadExecutor.Execute((Action)(() =>
+                {
+                    rect.RaiseEvent(
+                        new MouseButtonEventArgs(Mouse.PrimaryDevice, Environment.TickCount, MouseButton.Left)
+                        {
+                            RoutedEvent = UIElement.MouseDownEvent
```
leeyh20: This routed event code seems to be repeated quite a few times, would it be possible to refactor into a function?

PowerPointLabs/PowerPointLabs/TextCollection/ColorsLabText.cs
```diff
@@ -14,17 +14,16 @@ internal static class ColorsLabText
         public const string TaskPanelTitle = "Colors Lab";
 
         public const string MainColorBoxTooltip = "Choose the main color: " +
-                                   "\r\nDrag the button to pick a color, " +
+                                   "\r\nYou can drag the box to favorites palatte, " +
```
leeyh20: Maybe start the sentence with "Drag the box..." so that it is more consistent?

PowerPointLabs/PowerPointLabs/Converters/ColorPane/SelectedColorShiftLuminosityByFactor.cs
blewjy: That's right, `HSLColor` when implicitly converted to `System.Drawing.Color` will handle `null` by setting the `Color` to `rgb(0, 0, 0)`. 

`Color` is a struct in C#, so not allowed to set as `null`. If I'm not wrong, there will be a `NullReferenceException` thrown if we attempt to set `Color` to `null`. But since we have our own `HSLColor`, this is handled here.

PowerPointLabs/PowerPointLabs/FunctionalTestInterface.Impl/Controller/ColorsLabController.cs
blewjy: Good idea, done!

PowerPointLabs/PowerPointLabs/TextCollection/ColorsLabText.cs
blewjy: Done

PowerPointLabs/PowerPointLabs/ColorsLab/HSLColor.cs
```diff
@@ -44,6 +44,12 @@ public string ToRGBString()
             return String.Format("R: {0:#0.##} G: {1:#0.##} B: {2:#0.##}", color.R, color.G, color.B);
         }
 
+        public string ToHEXString()
```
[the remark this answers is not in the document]
blewjy: Oops, changed!

PowerPointLabs/PowerPointLabs/ActionFramework/ColorsLab/ColorsLabActionHandler.cs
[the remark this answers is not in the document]
blewjy: No difference... I think I took this from one of the other labs. I've shortened this block of code.

leeyh20:
Thanks for this PR! 👍

PR #1780, opened 2019-02-13
Tooltips lab: Complete Assigning Tooltip Functionality

ChesterSng:
This PR completes the functionality to `Assign Tooltip`.

When selecting multiple shapes, the first shape selected will be the `Trigger Shape`, the rest of the shapes will be the `Animated Shapes`. All of the `Animated Shapes` will appear on the click of `Trigger Shape` in slideshow mode.

Opens the Animation Pane on successful assigning of tooltip


blewjy:
@ChesterSng Ah ok, I think I managed to find the code to check.

```c#
if (!this.GetApplication().CommandBars.GetPressedMso("AnimationCustom"))
{
	this.GetApplication().CommandBars.ExecuteMso("AnimationCustom");
}
```

This should work.

blewjy:
Hey @ChesterSng, a couple more things:

1. If there's no selection but the user clicks `Assign Tooltip`, there's no feedback to the user. Would be good if we can have the similar popup when only one shape is selected.

2. If only one shape is selected, there will be a popup to ask user to select at least 2 shapes. After that popup is dismissed, Animation Pane is opened even though no animations were added. I think for this specific action, animation pane should remain closed.

ChesterSng:
> 1. If there's no selection but the user clicks `Assign Tooltip`, there's no feedback to the user. Would be good if we can have the similar popup when only one shape is selected.

I think we can disable the button in the future so that when nothing is selected, the button is greyed out. 

> 
2. If only one shape is selected, there will be a popup to ask user to select at least 2 shapes. After that popup is dismissed, Animation Pane is opened even though no animations were added. I think for this specific action, animation pane should remain closed.

Yeah I agree that it should be closed, modified the code so that it will remain closed when there are no tooltips are added.

PR #1787, opened 2019-02-24
New menu buttons: Create Callout and Create Trigger

blewjy:
This PR adds 2 new menu buttons in addition to simply `Create Tooltip`, user can now choose to select one or multiple shapes as their pre-existing callout/trigger, then to use `Create Callout` or `Create Trigger` to have our default shapes generated for them with the trigger animations attached.


ChesterSng:
Hi @blewjy, thanks for making the changes as discussed! 

Have some suggestions: 

1. Disable `Create Callout` and `Create Trigger` buttons when there is no selection, so it works like `Assign Tooltip` button. 

2. Subseqent usage of `Create Callout`, `Create Trigger` and `Create Tooltip` should not be placed in the exact same position. 

**Now**
On second use of `Create Tooltip`: 
image

**Suggested**
image

This will be similar to when you are trying to paste the same shape multiple times on the slide. 

I'm thinking we can further expanding this menu: 

1. One trigger - One Callout __(Supported)__
2. One trigger - Many Callouts __(Can only be done by `Assign Tooltip` manually for now)__
3. Many triggers - One Callout __Is this necessary?__

Point no.2 seems to be a reasonable use for `Tooltips Lab`.

damithc:
> 3\. Many triggers - One Callout _Is this necessary?_

No need for this kind of advanced features in the first version. Just do a MVP version out into the dev release. No need for tests or proper documentation either. It's important that I get to dogfood the feature as early as possible.

PR #1789, opened 2019-02-26
#1374 Disable rotation mode when exception is caught

### LINKED ISSUE #1374: Positions Lab Rotate grouped shapes can become unselected
Expected behavior: the shape selections should not change.

Furthermore, if the pivot object is deleted after the other objects are unselected, dragging other objects will randomly position the rotating objects.

ChesterSng:
Fixes #1374 partially.

**Outline of Solution**
I am unable to diagnose why grouped shapes might become unselected after releasing the mouse. 
`_selectedRange` always returns the right group of shapes upon manual checking in the [mouse up handler](https://github.com/ChesterSng/PowerPointLabs/blob/98030db1e4a08efa905f413dae23a9a6220ac143/PowerPointLabs/PowerPointLabs/PositionsLab/PositionsPaneWPF.xaml.cs#L356-L362).

However, solved the issue caused by the deletion of the reference shape. 

_Before_
An exception will be caught however, the Rotation Mode is not disabled in the caught exception block.

_After_
Disable rotation mode when exception is caught.

### DIFF
--- a/PowerPointLabs/PowerPointLabs/PositionsLab/PositionsPaneWPF.xaml.cs
+++ b/PowerPointLabs/PowerPointLabs/PositionsLab/PositionsPaneWPF.xaml.cs
@@ -375,7 +375,7 @@ namespace PowerPointLabs.PositionsLab
                                                                        : null;
             try
             {
-                if (button.IsMouseOver)
+                if (button == null || button.IsMouseOver)
                 {
                     DisableRotationMode();
                     return;
--- a/PowerPointLabs/PowerPointLabs/PositionsLab/PositionsPaneWPF.xaml.cs
+++ b/PowerPointLabs/PowerPointLabs/PositionsLab/PositionsPaneWPF.xaml.cs
@@ -370,12 +370,11 @@ namespace PowerPointLabs.PositionsLab
 
         void _leftMouseDownListener_Rotation(object sender, SysMouseEventInfo e)
         {
-            try
-            {
-                ToggleButton button = ((bool)rotationButton.IsChecked) ? rotationButton :
+            ToggleButton button = ((bool)rotationButton.IsChecked) ? rotationButton :
                              ((bool)duplicateRotationButton.IsChecked) ? duplicateRotationButton
                                                                        : null;
-
+            try
+            {
                 if (button.IsMouseOver)
                 {
                     DisableRotationMode();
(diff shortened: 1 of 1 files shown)


leeyh20:
@damithc Hi Prof, is disabling the rotation mode acceptable for this situation?

damithc:
@ChesterSng what is exact behavior from the user's POV?

ChesterSng:
Hi Prof @damithc and @leeyh20, 

When using `rotate` with a group of shapes, the group of shapes will become unselected sometimes. 

If users delete the reference point shape, it will result in unexpected behaviours as shown here: 
Users will not be able to deselect the shape by clicking outside of it because Rotation Mode is still enabled.
grouped shapes error

An exception will be thrown when accessing the reference point shape in this line in `PositionsPaneWPF.xaml.cs` https://github.com/PowerPointLabs/PowerPointLabs/blob/eebd9c4a49273773aa0aec163dba42feed8c9446/PowerPointLabs/PowerPointLabs/PositionsLab/PositionsPaneWPF.xaml.cs#L396

However the catch exception simply logs the exception, but Rotation Mode is still enabled. 

After making the disabling Rotation Mode in the exception, if the user deletes the reference shape, the rotation mode will end and user will be able to drag/deselect the shapes.

grouped shapes error fixed

damithc:
Thanks for the explanation @ChesterSng 
Yes, the proposed fix sounds reasonable.

leeyh20:
Sounds good! Could someone do a peer-review for this PR?

PowerPointLabs/PowerPointLabs/PositionsLab/PositionsPaneWPF.xaml.cs
```diff
@@ -370,12 +370,11 @@ void _leftMouseUpListener_Rotation(object sender, SysMouseEventInfo e)
 
         void _leftMouseDownListener_Rotation(object sender, SysMouseEventInfo e)
         {
-            try
-            {
-                ToggleButton button = ((bool)rotationButton.IsChecked) ? rotationButton :
+            ToggleButton button = ((bool)rotationButton.IsChecked) ? rotationButton :
                              ((bool)duplicateRotationButton.IsChecked) ? duplicateRotationButton
```
YuPeiHenry: This code might not be modified in your PR, but I feel that instead of assigning null to button, it may be better to return straight? Because calling button.IsMouseOver on null knowing that it will be null feels weird. (But it seems the previous author intended for this because an error will be logged when button gets set to null, seems to promote code reuse in a rather unusual way).

leeyh20:
LGTM!

PR #1790, opened 2019-02-27
Tooltips Lab: Change settings of Tooltipslab to show not implemented yet

ChesterSng:
Disable settings for Tooltips Lab


leeyh20:
I have a suggestion, could someone start a PR from tooltips-lab branch to the dev branch (but it will be WIP work in progress) so that we can easily see the gradual changes of the tooltips-lab branch? I'm concerned that the final PR will have so many changes it will be difficult for us to review.

ChesterSng:
@leeyh20 yeah we are planning to do that after this pr is merged :)

PR #1795, opened 2019-02-28
New Lab: Tooltips lab

ChesterSng:
This new lab is done by @YuPeiHenry, @blewjy and @ChesterSng.

`Tooltips Lab` will help users of PowerPointLab easily create Tooltips with the `Create Menu` and `Assign Tooltip` functionalities:

**Create Menu**
1. `Create Tooltip` 
 - Create a trigger and callout with trigger animations.
 - Ignores any selection

create tooltip

2. `Create Callout`
 - Need to have selection (can have multiple)
 - One selection: Callout created above trigger, attached to selection as trigger
 - Multiple selections: Callouts created, attached to each shape as trigger

create callout

3. `Create trigger`
 - Need to have selection (can have multiple)
 - One selection: Trigger created below callout, callout attached to trigger
 - Multiple selections: Triggers created below callouts, callouts attached to triggers

create trigger

-----
**Assign Tooltip**
- Must have more than one shape in selection
- First shape will be trigger
- All other shapes will be callout
- All shapes will be attached to first shape as trigger and will appear and disappear together on first and second clicks.
assign tooltip

- Subsequent assigning of tooltip to same trigger will cause all callout shapes to appear and disappear on first and second click. 

Note: A loop can be created using `Assign Tooltip` causing all shapes to not be visible on the slide. However, the original powerpoint application allows for this behaviour and thus we will ignore this.

Note: This version of `Tooltips Lab` will not come with settings and we are open to suggestions on what should come in the settings menu.


leeyh20:
Do you need to restart the Appveyor build?

ChesterSng:
@leeyh20, restarting the build will not help, tried many times. The AppVeyor build will fail on the `tooltips-lab` branch because of permission errors but the PR to `dev-release` is fine.

PowerPointLabs/PowerPointLabs/ActionFramework/TooltipsLab/Create/CreateTrigger/CreateTriggerActionHandler.cs
```diff
@@ -0,0 +1,47 @@
...
+
+            if (selection.Type != PowerPoint.PpSelectionType.ppSelectionShapes)
+            {
+                MessageBox.Show("Please select 1 or more shapes as your callout shape.");
+                return;
+            }
+
+            foreach (PowerPoint.Shape selectedShape in selection.ShapeRange)
+            {
+                // TODO: Adding a trigger to an existing shape that is already part of a trigger animation as the "callout" (whether created by user or by TooltipsLab)
```
leeyh20: This will be addressed in a future issue? :)

PowerPointLabs/PowerPointLabs/TooltipsLab/CreateTooltip.cs
```diff
@@ -0,0 +1,70 @@
...
+                height);
+
+            return triggerShape;
+        }
+
+        public static PowerPoint.Shape GenerateCalloutWithReferenceTriggerShape(PowerPointSlide currentSlide, PowerPoint.Shape triggerShape)
+        {
+            float midpointX = ShapeUtil.GetMidpointX(triggerShape);
+
+            // TODO: To be moved to TooltipsLabConstants in the future.
```
leeyh20: Yes, should immediately move these to `TooltipsLabConstants` in another issue / PR?

PowerPointLabs/PowerPointLabs/TooltipsLab/CreateTooltip.cs
```diff
@@ -0,0 +1,70 @@
...
+                triggerShape.Top - (float)(1.125 * height) - 10,
+                width,
+                height);
+
+            return callout;
+        }
+
+        public static PowerPoint.Shape GenerateTriggerShape(PowerPointSlide currentSlide)
+        {
+            // TODO: These are all hardcoded values for now. They should be replaced in the future.
```
leeyh20: Replaced in another issue / PR?

PowerPointLabs/PowerPointLabs/ActionFramework/TooltipsLab/AssignTooltip/AssignTooltipActionHandler.cs
```diff
@@ -0,0 +1,43 @@
...
+                AssignTooltip.AddTriggerAnimation(currentSlide, selection);
+                // Open the animation custom pane if it is not opened
+                if (!this.GetApplication().CommandBars.GetPressedMso("AnimationCustom"))
+                {
+                    this.GetApplication().CommandBars.ExecuteMso("AnimationCustom");
+                }
+            }
+            catch (Exception)
+            {
+
```
leeyh20: Seems like not very nice to catch a generic exception and then leave this part totally empty... consider passing the error message with the exception and showing a message box with the error message?

PowerPointLabs/PowerPointLabs/ActionFramework/TooltipsLab/AssignTooltip/AssignTooltipActionHandler.cs
```diff
@@ -0,0 +1,43 @@
...
+            if (!ShapeUtil.IsSelectionShape(selection))
+            {
+                return;
+            }
+
+            try
+            {
+                AssignTooltip.AddTriggerAnimation(currentSlide, selection);
+                // Open the animation custom pane if it is not opened
+                if (!this.GetApplication().CommandBars.GetPressedMso("AnimationCustom"))
```
leeyh20: Maybe `"AnimationCustom"` can be in `TooltipsLabConstants` too?

PowerPointLabs/PowerPointLabs/ActionFramework/TooltipsLab/Create/CreateCallout/CreateCalloutActionHandler.cs
```diff
@@ -0,0 +1,44 @@
...
+            }
+
+            foreach (PowerPoint.Shape selectedShape in selection.ShapeRange)
+            {
+                PowerPoint.Shape callout = CreateTooltip.GenerateCalloutWithReferenceTriggerShape(currentSlide, selectedShape);
+                PowerPoint.Shape calloutGroup = AddTextbox.AddTextboxToCallout(currentSlide, callout);
+                AssignTooltip.AddTriggerAnimation(currentSlide, selectedShape, calloutGroup);
+            }
+            
+            if (!this.GetApplication().CommandBars.GetPressedMso("AnimationCustom"))
```
leeyh20: `AnimationCustom` is used again here.

PowerPointLabs/PowerPointLabs/ActionFramework/TooltipsLab/Create/CreateTooltip/CreateTooltipActionHandler.cs
```diff
@@ -0,0 +1,33 @@
...
+namespace PowerPointLabs.ActionFramework.TooltipsLab
+{
+    [ExportActionRibbonId(TooltipsLabText.CreateTooltipTag)]
+    class CreateTooltipActionHandler : ActionHandler
+    {
+        protected override void ExecuteAction(string ribbonId)
+        {
+            this.StartNewUndoEntry();
+
+            PowerPointSlide currentSlide = this.GetCurrentSlide();
```
leeyh20: Consider adding a guard clause that checks if `currentSlide `is null?

PowerPointLabs/PowerPointLabs/ThisAddIn.cs
```diff
@@ -978,6 +978,7 @@ private void ThisAddInSlideSelectionChanged(PowerPoint.SlideRange sldRange)
                     prev = presentation.Slides[slideIndex - 1];
                 }
             }
+            Ribbon.RefreshRibbonControl("CreateMenu");
```
leeyh20: Could this be called "CreateTooltipMenuButton" instead, if CreateMenu is only relevant to TooltipsLab? This button name seems to be inconsistent with the rest.

PowerPointLabs/PowerPointLabs/TooltipsLab/AttachTriggerAnimation.cs
```diff
@@ -0,0 +1,142 @@
...
+            Sequences sequences = currentSlide.TimeLine.InteractiveSequences;
+            // A set is used here so no duplicate shapes will be added
+            ISet<Shape> shapesToAnimateSet = new HashSet<Shape>(shapesToAnimate);
+
+            // Find the existing sequence that has the triggerShape
+            for (int i = 1; i <= sequences.Count; i++)
+            {
+                Sequence sequence = sequences[i];
+                // Iterate from the back because of deletion
+                for (int j = sequence.Count; j >= 1; j--)
```
leeyh20: `sequence[0] `does not need to be deleted? Is it because the index starts from 1?

PowerPointLabs/PowerPointLabs/TooltipsLab/CreateTooltip.cs
```diff
@@ -0,0 +1,70 @@
...
+using PowerPointLabs.Models;
+using PowerPointLabs.TextCollection;
+using PowerPointLabs.Utils;
+using PowerPointLabs.Views;
+
+using PowerPoint = Microsoft.Office.Interop.PowerPoint;
+
+namespace PowerPointLabs.TooltipsLab
+{
+    internal static class CreateTooltip
```
leeyh20: Because this is an entirely new class, I suggest adding some documentation comments to briefly describe the class if possible? `<summary> </summary> `

PowerPointLabs/PowerPointLabs/TooltipsLab/AttachTriggerAnimation.cs
```diff
@@ -0,0 +1,142 @@
...
+        }
+
+        private static void AddTriggerAnimation(PowerPointSlide currentSlide, Shape triggerShape, List<Shape> newShapesToAnimate)
+        {
+            TimeLine timeline = currentSlide.TimeLine;
+            // TODO: Get effect from settings
+            MsoAnimEffect fadeEffect = MsoAnimEffect.msoAnimEffectFade;
+
+            // Get the shapes that are already associated with trigger shape
+            List<Shape> shapesToAnimate = RemoveAnimationsInInteractiveSequence(currentSlide, triggerShape, newShapesToAnimate);
```
leeyh20: The name "`RemoveAnimationsInInterativeSequence`" seems a bit misleading as it seems like its function is to GetShapes of some kind, but yet in the name there is no suggestion of getting shapes, sounds like it is just removing..?

PowerPointLabs/PowerPointLabs/TooltipsLab/AttachTriggerAnimation.cs
ChesterSng: yes the index starts from 1 here

leeyh20: Ok!

PowerPointLabs/PowerPointLabs/ActionFramework/TooltipsLab/Create/CreateTooltip/CreateTooltipActionHandler.cs
ChesterSng: The button will be disabled when there are no slides, but will add a guard clause!

PowerPointLabs/PowerPointLabs/TooltipsLab/AttachTriggerAnimation.cs
ChesterSng: I understand your concern here, as much as I would like to make the function have a "single repsonsibility" it doesn't seem possible here. 

What do you suggest i rename the function, `GetShapesAndRemoveAnimationsInInteractiveSequence`?

leeyh20: `GetShapesInInteractiveSequenceWithAnimationsRemoved` ?

PowerPointLabs/PowerPointLabs/ActionFramework/TooltipsLab/Create/CreateTrigger/CreateTriggerActionHandler.cs
blewjy: Turns out that this was intended behaviour. I added this todo wrongly, haha

PowerPointLabs/PowerPointLabs/ThisAddIn.cs
blewjy: This naming actually follows the convention for the rest of the labs. "CreateMenu" is actually a Menu, not a Button. "CreateMenu" expands into 3 buttons, "CreateTooltipButton", "CreateCalloutButton" and "CreateTriggerButton".

PowerPointLabs/PowerPointLabs/TooltipsLab/CreateTooltip.cs
blewjy: Initially I added this comment because the values are just some default values that I came up with. We decided that these values were appropriate, this comment should have been removed.

PowerPointLabs/PowerPointLabs/ActionFramework/TooltipsLab/Create/CreateMenu/CreateMenuEnabledHandler.cs
```diff
@@ -0,0 +1,19 @@
...
+
+namespace PowerPointLabs.ActionFramework.TooltipsLab
+{
+    [ExportEnabledRibbonId(TooltipsLabText.CreateMenuTag)]
+    class CreateMenuEnabledHandler : EnabledHandler
+    {
+        protected override bool GetEnabled(string ribbonId)
+        {
+            PowerPointSlide currentSlide = this.GetCurrentSlide();
+            return currentSlide != null;
```
LiXiaoooowei: Hi, just noticed a small issue here. A `null reference exception` will be thrown when the code is running from Visual Studio if current slide is null. The exception is caused by the underlying implementation of `this.GetCurrentSlide()` function.  

Suggested way to fix this:
- Use a boolean variable to indicate whether `createMenu` should be enabled, and alter the variable value when slide selection is changed. In this way, empty slide selection can be handled explicitly. OR
- Change `PowerPointCurrentPresentationInfo.CurrentSlide` to handle null value explicitly.

YuPeiHenry: @LiXiaoooowei How do you replicate the problem? I can't seem to get the exception.

LiXiaoooowei: > @LiXiaoooowei How do you replicate the problem? I can't seem to get the exception.

I think it probably depends on your exception settings for visual studio. I set it to break when `C++ Exceptions Platform::NullReferenceException` is thrown.

PR #1798, opened 2019-03-01
#1756 Prevent illegal adjustment of GridSplitter

### LINKED ISSUE #1756: Adjusting separator in PictureSlidesLab throws exception
## Bug Report  
Adjusting the separator/splitter in PictureSlidesLab throws unhandled exception.
picture-slides-lab-center-pane-1

Windows 10, PowerPoint 2013
#### Steps to replicate
1. Open PictureSlidesLab. If you don't see two panes (left one with slide preview, right one with pictures), then add a picture, and you should be able to see the two panes.
2. Drag separator/splitter all the way to the left until the end of the window, you should encounter an exception. (Follow screenshot above)

```
System.ArgumentException
  HResult=0x80070057
  Message='-8.5' is not a valid value for property 'Width'.
  Source=PowerPointLabs
  StackTrace:
   at PowerPointLabs.PictureSlidesLab.Views.PictureSlidesLabWindow.Splitter_OnDragDelta(Object sender, DragDeltaEventArgs e) in C:\Users\Bryan\Desktop\PowerPointLabs\PowerPointLabs\PowerPointLabs\PictureSlidesLab\Views\PictureSlidesLabWindow.xaml.cs:line 525
```

ChesterSng:
Fixes #1756 

There were no minimum widths for the grid columns so the splitter was able to exit the screen. Also, splitter can resize grid columns to an unusable width.

**Outline of Solution**

1. Set minimum Widths for the grid columns

2. Add checks for dragging of splitter so `StylesPreviewGrid` (the area which the slides are previewed) will be the same size as the grid column 

_The minimum widths 300 and 400 are selected because they are the minimum width where all functionalities will be on the screen_

__Minimum width of 300 for previewing styles__
minimum preview styles

__Minimum width of 400 for the picture selector__
minimum flyout

### DIFF
--- a/PowerPointLabs/PowerPointLabs/PictureSlidesLab/Views/PictureSlidesLabWindow.xaml.cs
+++ b/PowerPointLabs/PowerPointLabs/PictureSlidesLab/Views/PictureSlidesLabWindow.xaml.cs
@@ -532,7 +532,21 @@ namespace PowerPointLabs.PictureSlidesLab.Views
         /// <param name="e"></param>
         private void Splitter_OnDragDelta(object sender, DragDeltaEventArgs e)
         {
-            StylesPreviewGrid.Width = StylesPreviewGrid.ActualWidth + e.HorizontalChange;
+            double newWidth = StylesPreviewGrid.ActualWidth + e.HorizontalChange;
+            // Prevent StylesPreviewGrid from becoming too small
+            if (newWidth < StylesPreviewGrid.MinWidth)
+            {
+                StylesPreviewGrid.Width = StylesPreviewGrid.MinWidth;
+            }
+            // Prevent StylesPreviewGrid from overflowing grid column 0
+            else if (newWidth > Column0.Width.Value)
+            {
+                StylesPreviewGrid.Width = Column0.Width.Value;
+            }
+            else
+            {
+                StylesPreviewGrid.Width = newWidth;
+            }
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


PowerPointLabs/PowerPointLabs/PictureSlidesLab/Views/PictureSlidesLabWindow.xaml.cs
```diff
@@ -532,7 +532,21 @@ private void ImageSelectionListBox_OnKeyDown(object sender, KeyEventArgs e)
         /// <param name="e"></param>
         private void Splitter_OnDragDelta(object sender, DragDeltaEventArgs e)
         {
-            StylesPreviewGrid.Width = StylesPreviewGrid.ActualWidth + e.HorizontalChange;
+            double newWidth = StylesPreviewGrid.ActualWidth + e.HorizontalChange;
```
YuPeiHenry: I think it might be better if you extract this into a Setter. This way, others will not unintentionally violate this rule.

ChesterSng: Are you suggesting extracting `StylesPreviewGrid.ActualWidth + e.HorizontalChange` into another method?

YuPeiHenry: @ChesterSng Actually I meant the code below, sorry. The <minWidth and >maxWidth part.

ChesterSng: Hi @YuPeiHenry, I considered your suggestion but considering that adjusting the splitter window is the only method that will adjust the width of `StylesPreviewGrid`, I think it is not necessary to create a separate setter for it. It is sufficient to contain the entire logic in this function.

PR #1800, opened 2019-03-02
#1797 Add checkbox to ribbon for user preference to compress images

### LINKED ISSUE #1797: 'Crop to shape' lowers image resolution
## Bug Report  
Bug report from user
Not specified
#### Description
I'm currently using PowerPointLabs for some basic design work, and I've found that when using the 'crop to shape' feature, the image quality is significantly worsened.

Is there a way to use this feature without this occurring?

blewjy:
Fixes #1797 

**Outline of Solution**
Previously, when using `Crop to Shape`, high resolution images will be compressed after the crop. This is because we were exporting the images at 96ppi. This PR bumps this up to 330ppi to maintain a certain amount of quality on the pictures.

Before:
croptoshape-before
After:
croptoshape-after

There is also a checkbox added to the main ribbon to allow users to decide if they would like images to be compressed. If this is checked, the image exporting resolution will be set back down to the original 96ppi.

Checkbox:
compress-images-checkbox

Note: this PR also affects other labs that use the same exporting and import method, including `EffectsLab`, thus the FT slides for `EffectsLab` had to be updated.

### DIFF
--- a/PowerPointLabs/PowerPointLabs/Ribbon1.cs
+++ b/PowerPointLabs/PowerPointLabs/Ribbon1.cs
@@ -160,15 +160,7 @@ namespace PowerPointLabs
         {
             ShouldCompressImages = pressed;
             _ribbon.InvalidateControl("ShouldCompressImagesCheckbox");
-
-            if (ShouldCompressImages)
-            {
-                GraphicsUtil.PictureExportingRatio = 96.0f / 72.0f;
-            }
-            else
-            {
-                GraphicsUtil.PictureExportingRatio = 330.0f / 72.0f;
-            }
+            GraphicsUtil.ShouldCompressPictureExport(ShouldCompressImages);
         }
 
 
--- a/PowerPointLabs/PowerPointLabs/Ribbon1.cs
+++ b/PowerPointLabs/PowerPointLabs/Ribbon1.cs
@@ -188,6 +188,12 @@ namespace PowerPointLabs
             return DisableFormatTab;
         }
 
+        // Sets the default starting status of the checkbox (whether checked or not)
+        public bool SetImageCompression(Office.IRibbonControl control)
+        {
+            return ShouldCompressImages;
+        }
+
         public void RibbonLoad(Office.IRibbonUI ribbonUi)
         {
             ActionHandlerFactory = new ActionHandlerFactory();
--- a/PowerPointLabs/PowerPointLabs/Ribbon1.cs
+++ b/PowerPointLabs/PowerPointLabs/Ribbon1.cs
@@ -12,6 +12,7 @@ using PowerPointLabs.ActionFramework.Common.Log;
 using PowerPointLabs.ELearningLab.Service;
 using PowerPointLabs.PictureSlidesLab.Views;
 using PowerPointLabs.TextCollection;
+using PowerPointLabs.Utils;
 
 using Office = Microsoft.Office.Core;
 using PowerPoint = Microsoft.Office.Interop.PowerPoint;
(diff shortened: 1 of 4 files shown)


ChesterSng:
Hi @blewjy, good job on solving the issue! Tested on PowerPoint 2016
Image now retain its resolution after `Crop` functions.

There is an issue when checking `Compress Images`:
**Compress Image becomes unchecked after switching tabs**
1. Check `Compress Images`
2. Go to any other ribbon tab i.e. `Home`
3. Go back to `PowerPointLabs` tab.
4. `Compress Images` is now unchecked. 
However, if you do `Crop` functions, the images are now lower resolution, which means even though it is unchecked, `Compress Images` is still active. 

I also have a suggestion for the tooltip of `Compress Images`: 

People might not understand the what functionalities it affects even though it states `image-related functions`, not sure if it is important to include a list of features that it will affect for better clarity.

Alternatively, maybe a `?` icon can be added to beside the Compress Images option so people who click it can get a more detailed explanation of the option:

image

blewjy:
Hey @ChesterSng, thanks for your review!

> I think you need to add a getPressed parameter and a function to use ShouldCompressImages like in Maintain Tab Focus.

You're right about this, adding the function solved that issue.

> People might not understand the what functionalities it affects even though it states image-related functions, not sure if it is important to include a list of features that it will affect for better clarity.
Alternatively, maybe a ? icon can be added to beside the Compress Images option so people 
who click it can get a more detailed explanation of the option:

To keep it consistent with the other buttons, I decided to add more detail to the tooltip instead. I listed out the features that were affected by this compress image setting.

leeyh20:
Was it tested on 2016?

ChesterSng:
@leeyh20 Yup tested!

PowerPointLabs/PowerPointLabs/Ribbon1.cs
```diff
@@ -152,17 +155,45 @@ public void ToggleVisibility(Office.IRibbonControl control, bool pressed)
...
 
+        // Toggles the boolean controlling whether images should be compressed.
+        public void ToggleImageCompression(Office.IRibbonControl control, bool pressed)
+        {
+            ShouldCompressImages = pressed;
+            _ribbon.InvalidateControl("ShouldCompressImagesCheckbox");
+
+            if (ShouldCompressImages)
+            {
+                GraphicsUtil.PictureExportingRatio = 96.0f / 72.0f;
```
leeyh20: it will be better if these ratio values will be stored in GraphicsUtil as a constant so that if anyone wants to find these values then they will directly go to GraphicsUtil, not Ribbon1.

blewjy: Done! moved them into GraphicsUtil

PowerPointLabs/PowerPointLabs/Utils/GraphicsUtil.cs
```diff
@@ -28,9 +28,14 @@ internal static class GraphicsUtil
...
+        public static float PictureExportingRatio = 330.0f / 72.0f;
         private const float TargetDpi = 96.0f;
         private static float dpiScale = 1.0f;
+
+        // Picture exporting ratios
+        private const float PictureExportingRatioHigh = 330.0f / 72.0f;
+        private const float PictureExportingRatioCompressed = 96.0f / 72.0f;
+
         // Heuristics for image compression obtained through testing
         private const long targetCompression = 75L;
```
leeyh20: I realised that these variables are lower-case while the above const float is upper-case for the first letter, could you standardise them according to https://github.com/oss-generic/process/blob/master/codingStandards/CodingStandard-CSharp.adoc ?

blewjy: Done! Changed to camel case for private variabels.

PR #1802, opened 2019-03-07
Tooltips Lab: Refactor code based on PR comments

ChesterSng:
- Rename AssignTriggerAnimation.cs to AssignTooltip.cs
- Change AssignTooltip.AddTriggerAnimation to return `bool` so exception is not needed to prevent opening of animation pane
- Add summary to AssignTooltip.cs


blewjy:
LGTM

PR #1803, opened 2019-03-07
Move constants to TooltipsLabConstants + refactor

blewjy:
- Added guard clauses for checking of active slide selection
- Remove expired comments
- Move all constants and strings to `TooltipsLabConstants` and `TooltipsLabText`


ChesterSng:
LGTM 👍

PR #1805, opened 2019-03-07
#1769 Replace .Copy() and .Paste() in ZoomToArea with .Duplicate()

### LINKED ISSUE #1769: Error when using 'zoom to area'
v4.4.0.0 Dev
PPT 2016 Win10

Steps:
* Use zoom-to-area in the attached slide.
[zoomLaberror.pptx](https://github.com/PowerPointLabs/PowerPointLabs/files/2840603/zoomLaberror.pptx)

>System.Runtime.InteropServices.COMExceptionShapes (unknown member) : Invalid request.  Clipboard is empty or contains data which may not be pasted here.Stack Trace:   at Microsoft.Office.Interop.PowerPoint.Shapes.Paste()   at PowerPointLabs.ZoomLab.ZoomToArea.GetBestFitShape(PowerPointSlide currentSlide, Shape zoomShape)   at PowerPointLabs.ZoomLab.ZoomToArea.<>c__DisplayClass3_0.<GetEditedShapesForZoomToArea>b__0(Shape zoomShape)   at System.Linq.Enumerable.WhereSelectListIterator`2.MoveNext()   at System.Collections.Generic.List`1..ctor(IEnumerable`1 collection)   at System.Linq.Enumerable.ToList[TSource](IEnumerable`1 source)   at PowerPointLabs.ZoomLab.ZoomToArea.GetEditedShapesForZoomToArea(PowerPointSlide currentSlide, List`1 zoomRectangles)   at PowerPointLabs.ZoomLab.ZoomToArea.AddZoomToArea()

ChesterSng:
Fixes #1769 

**Outline of Solution**

After researching and experimenting on the problem, it seems an exception will be thrown on some machines because `ZoomToArea` uses a lot of `.Copy()` and `.Paste()`. It might be due to concurrent access of clipboard resulting in `AccessViolationException` being thrown. 

Replaced `.Copy()` and `.Paste()` with `.Duplicate()` in functions that do not require the use of clipboard, i.e. copying a shape from one slide and pasting it to another. This seemed to solve the issue on my machine. 

Also, `.PasteSpecial()` throws exception consistently on my machine as well, fixed the problem by exporting the shape as an image first and then adding it back into the slide like in PR #1761 

Previously, my `ZoomToArea` FT was failing due to this error, so need more people to test on this PR to make sure the fix really works. 

This PR also affects `CropLab#CropToShape` because it is used in `ZoomToArea`. Previously, `CropToShape` also threw a `AccessViolationException` and changing it to use `.Duplicate()` fixed it.

### DIFF
--- a/PowerPointLabs/PowerPointLabs/Models/PowerPointMagnifyingSlide.cs
+++ b/PowerPointLabs/PowerPointLabs/Models/PowerPointMagnifyingSlide.cs
@@ -1,9 +1,10 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
-
+using System.Windows;
 using PowerPointLabs.AnimationLab;
 using PowerPointLabs.CropLab;
+using PowerPointLabs.Utils;
 using PowerPointLabs.ZoomLab;
 
 using Office = Microsoft.Office.Core;
@@ -144,9 +145,33 @@ namespace PowerPointLabs.Models
             cropShape.Select();
             PowerPoint.Selection sel = Globals.ThisAddIn.Application.ActiveWindow.Selection;
             PowerPoint.Shape croppedShape = CropToShape.Crop(zoomSlideCopy, sel, magnifyRatio: magnifyRatio);
-            croppedShape.Cut();
+            try
+            {
+                string tempFilePath = FileDir.GetTemporaryPngFilePath();
+                Utils.GraphicsUtil.ExportShape(croppedShape, tempFilePath);
+                zoomSlideCroppedShapes = _slide.Shapes.AddPicture2(tempFilePath,
+                    Office.MsoTriState.msoFalse,
+                    Office.MsoTriState.msoTrue,
+                    0,
+                    0);
+                croppedShape.Delete();
+                try
+                {
+                    FileDir.DeleteFile(tempFilePath);
+                }
+                catch (Exception)
+                {
+                    // If the file cannot be deleted, we continue without deletion.
+                }
 
-            zoomSlideCroppedShapes = _slide.Shapes.PasteSpecial(PowerPoint.PpPasteDataType.ppPastePNG)[1];
+            }
+            catch (Exception)
+            {
+                // Revert to normal copy and pasting if unable to create file.
+                croppedShape.Cut();
+                zoomSlideCroppedShapes = _slide.Shapes.PasteSpecial(PowerPoint.PpPasteDataType.ppPasteMetafilePicture)[1];
+            }
+
             zoomSlideCroppedShapes.Name = "PPTLabsMagnifyAreaGroup" + DateTime.Now.ToString("yyyyMMddHHmmssffff");
             Utils.ShapeUtil.FitShapeToSlide(ref zoomSlideCroppedShapes);
             zoomSlideCopy.Delete();
--- a/PowerPointLabs/PowerPointLabs/Models/PowerPointMagnifyingSlide.cs
+++ b/PowerPointLabs/PowerPointLabs/Models/PowerPointMagnifyingSlide.cs
@@ -169,7 +169,7 @@ namespace PowerPointLabs.Models
             {
                 // Revert to normal copy and pasting if unable to create file.
                 croppedShape.Cut();
-                zoomSlideCroppedShapes = _slide.Shapes.PasteSpecial(PowerPoint.PpPasteDataType.ppPasteMetafilePicture)[1];
+                zoomSlideCroppedShapes = _slide.Shapes.PasteSpecial(PowerPoint.PpPasteDataType.ppPastePNG)[1];
             }
 
             zoomSlideCroppedShapes.Name = "PPTLabsMagnifyAreaGroup" + DateTime.Now.ToString("yyyyMMddHHmmssffff");
--- a/PowerPointLabs/PowerPointLabs/Models/PowerPointMagnifyingSlide.cs
+++ b/PowerPointLabs/PowerPointLabs/Models/PowerPointMagnifyingSlide.cs
@@ -101,9 +101,7 @@ namespace PowerPointLabs.Models
         //Return shape which helps to calculate the amount of zoom-in animation
         private PowerPoint.Shape GetReferenceShape(PowerPoint.Shape shapeToZoom)
         {
-            shapeToZoom.Copy();
-
-            PowerPoint.Shape referenceShape = _slide.Shapes.Paste()[1];
+            PowerPoint.Shape referenceShape = shapeToZoom.Duplicate()[1];
             referenceShape.LockAspectRatio = Office.MsoTriState.msoTrue;
             if (referenceShape.Width > referenceShape.Height)
             {
--- a/PowerPointLabs/PowerPointLabs/CropLab/CropToShape.cs
+++ b/PowerPointLabs/PowerPointLabs/CropLab/CropToShape.cs
@@ -49,8 +49,8 @@ namespace PowerPointLabs.CropLab
                 Shape shape = hasManyShapes ? shapeRange.Group() : shapeRange[1];
                 float left = shape.Left;
                 float top = shape.Top;
-                shape.Cut();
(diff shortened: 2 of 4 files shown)


ChesterSng:
@leeyh20 are u aware of any reason why `.copy()` and `.paste()` is used instead of `.duplicate()` previously? Based on testing, there seems to be no differences. Moreover, the clipboard of the user will not be affected when `.duplicate()` iis used, which is a plus.

ChesterSng:
@blewjy Yup I am getting the exact same error too. Have updated the FT slides!

blewjy:
@ChesterSng FT passes for me on PPT2013 windows 10 now. I personally have not been able to reliably reproduce the initial error, so I won't be able to comment on whether this fixes the original bug.

ChesterSng:
@YuPeiHenry can you help me test this too? Thanks!

leeyh20:
> 
> 
> @leeyh20 are u aware of any reason why `.Copy()` and `.Paste()` is used instead of `.duplicate()` previously? Based on testing, there seems to be no differences. Moreover, the clipboard of the user will not be affected when `.Duplicate()` is used, which is a plus.

I'm actually not very sure... There was an issue with .Duplicate() in another lab because it failed to work, but we could not figure out why.

LiXiaoooowei:
> @leeyh20 are u aware of any reason why `.Copy()` and `.Paste()` is used instead of `.duplicate()` previously? Based on testing, there seems to be no differences. Moreover, the clipboard of the user will not be affected when `.Duplicate()` is used, which is a plus.

Sometimes `.Duplicate()` throws exception if the shape to be copied does not exist in the current slide.

leeyh20:
@ChesterSng But yes I am aware `.PasteSpecial()` throws an exception for certain machines, and sometimes it also depends on what is in the clipboard.

PowerPointLabs/PowerPointLabs/Models/PowerPointMagnifyingSlide.cs
```diff
@@ -146,9 +145,33 @@ private void AddZoomSlideCroppedPicture(float magnifyRatio = 1.0f)
...
+                {
+                    // If the file cannot be deleted, we continue without deletion.
+                }
+
+            }
+            catch (Exception)
+            {
+                // Revert to normal copy and pasting if unable to create file.
+                croppedShape.Cut();
+                zoomSlideCroppedShapes = _slide.Shapes.PasteSpecial(PowerPoint.PpPasteDataType.ppPasteMetafilePicture)[1];
```
leeyh20: It used to be `PasteSpecial(PowerPoint.PpPasteDataType.ppPastePNG)` in the old code, is `ppPasteMetafilePicture` better? I think it used to throw exceptions for my machine, and I had to use `ppPastePNG` instead.

YuPeiHenry: Is there documentation somewhere that states which methods throw exceptions?

YuPeiHenry:
@ChesterSng ZoomToArea FT is failing for me on PPT2016.

PowerPointLabs/PowerPointLabs/Models/PowerPointMagnifyingSlide.cs
ChesterSng: i used `ppPasteMetaFilePicture` too, it doesn't help with solving the problems

ChesterSng:
> Sometimes .Duplicate() throws exception if the shape to be copied does not exist in the current slide.

That is right, it cannot copy a shape from another slide to the current slide. 

@YuPeiHenry what is the error that is produced? Previously ZoomToArea fails on my desktop with PowerPoint 2016 too, because slide similarity is not enough. After applying this PR, it still fails with the same problem.

YuPeiHenry:
The dev-release gives the error that Bryan reported. This branch gives the error below:

image

ChesterSng:
It seems like ZoomToArea will fail on the current `dev-release` branch as well on some machines.

leeyh20:
Is it very flaky?

ChesterSng:
Yeah it seems like the slides generated by one machine might not work on another (might be due to screen resolutions?). I'll try again to generate the slides from another machine and see if it works. If it doesn't we probably have to address this in another issue.

YuPeiHenry:
One solution might be resizing the PowerPoint window to certain aspect ratio, or certain size (maybe 800x600?) for the test itself.

ChesterSng:
Thanks @YuPeiHenry, that might be a possible solution, let me try that. 

In addition, the differences in powerpoint generate seem to differ only when on these two conditions: 

1.  Multiple shapes are selected 
**and**
2.  Generating animations on separate slides enabled 

The other tests have no problem.

ChesterSng:
@YuPeiHenry, unfortunately, standardising the resolution did not work. Looks like its not a problem with different resolution of powerpoint. 

Reverted changes to the powerpoint file. Will open another issue for this ZoomToArea FT instead.

PowerPointLabs/PowerPointLabs/Models/PowerPointMagnifyingSlide.cs
```diff
@@ -146,9 +145,33 @@ private void AddZoomSlideCroppedPicture(float magnifyRatio = 1.0f)
...
-            croppedShape.Cut();
+            try
+            {
+                string tempFilePath = FileDir.GetTemporaryPngFilePath();
+                Utils.GraphicsUtil.ExportShape(croppedShape, tempFilePath);
+                zoomSlideCroppedShapes = _slide.Shapes.AddPicture2(tempFilePath,
+                    Office.MsoTriState.msoFalse,
+                    Office.MsoTriState.msoTrue,
+                    0,
+                    0);
```
YuPeiHenry: Entering a single 0 on a new line looks a little out of place to me. @leeyh20 Is this a common coding practice for C#? I'm not sure.

PowerPointLabs/PowerPointLabs/Models/PowerPointMagnifyingSlide.cs
```diff
@@ -146,9 +145,33 @@ private void AddZoomSlideCroppedPicture(float magnifyRatio = 1.0f)
...
+                croppedShape.Delete();
+                try
+                {
+                    FileDir.DeleteFile(tempFilePath);
+                }
+                catch (Exception)
+                {
+                    // If the file cannot be deleted, we continue without deletion.
```
YuPeiHenry: Is there some information logging that should be done? What is the expected outcome of this block of code?

PowerPointLabs/PowerPointLabs/Models/PowerPointMagnifyingSlide.cs
```diff
@@ -146,9 +145,33 @@ private void AddZoomSlideCroppedPicture(float magnifyRatio = 1.0f)
...
+                }
+
+            }
+            catch (Exception)
+            {
+                // Revert to normal copy and pasting if unable to create file.
+                croppedShape.Cut();
```
YuPeiHenry: Inconsistent SLAP. The above code in the try block sounds like some special/modified routine to copy/paste a shape. Might it be better if the above code is extracted to a function, to match the level of detail from `croppedShape.Cut()`?

PowerPointLabs/PowerPointLabs/Models/PowerPointMagnifyingSlide.cs
ChesterSng: This block of code is needed so that an an exception from this particular exception will not be caught by the catch block below. If the file cannot be deleted due to some permissions issues, we will just leave the file in the user's directory.

leeyh20:
I'll merge this to dev-release so that @damithc can test it.

PowerPointLabs/PowerPointLabs/Models/PowerPointMagnifyingSlide.cs
leeyh20: We do have a logger for pptlabs, you could use it.

PR #1806, opened 2019-03-08
E-learning Lab Version 2 

LiXiaoooowei:
**Description**
This PR fixes some bugs and adds a few new features to e-learning lab.

**Changes in version 2**
- Increase UI text font size
- Change default callout shape format to be rectangular with yellow border
- When user ticks `use shorter callouts`, caption text will be auto-copied to callout text box
- Fix bug on error produced when clicking `speak` button from `AudioPreviewPage`
- Remove exit type PPTL animations if they appear in last click 
- Fix alignment problems on self explanation item
- Change `% completeness` to `Progress: %` on `ProgressStatusForm`
- Do not repeat the same click number 
- Change default callout animations to fade
- Auto resize text box to show all text when there is more than two lines of text
- Enable user to add up to 5 voice preferences : an easy way to try different voices
- Do not re-initialize ELL when slide selection changed unless ELL is open
- Add contextual menus upon right click a self explanation item to add explanations below and above
- Change voice button to be greyed out when disabled   
- Do not re-generate voice if text and speaker are the same during voice preview


blewjy:
Hi @LiXiaoooowei, good job with the PR! Some comments:

**1. When there is a blank item in the pane, the prompt to sync items keeps popping up but it doesn't delete the item for me**. 
The blank item is only deleted when I change slides. If I were to add a blank item and close the ELL pane, or go through the slide show, there will be the sync popup that tells me if I want to sync, but the blank item will still be there even after syncing. 

**2. Radio buttons not disabled, a little bit misleading**
In the ELL settings, under "Voice Preferences", if I do not set all 5 of my preferences to be the 5 azure voices, the "Other azure cloud voices" radio button in "Audio Preview Window" can be selected, even though I am not logged in to azure cloud voices. 
Expected behvaiour: I think the "Other azure cloud voices" button should be disabled as long as I am not logged in.

Also, when I'm not logged in, I think it is better for the radio buttons for the azure voices in the "Audio Preview Window" to be disabled. Currently, the "Speak" buttons for each azure voice is disabled when I'm not logged in, but the radio buttons are enabled, which is a little bit misleading as I can select them and press OK, only to be told that the Azure voices cannot be generated when I try to sync my items.

**3. Voice preferences: I am able to make the rankings have some weird behaviour if I click OK/CANCEL instead of update**
*Steps to reproduce:*
a. Start without any preferences. Rank some voices 1 to 3 (or any other ranking you want, as long as there is a saved preference), and click "Update Preferences".
b. Click 'click to edit preference', then change one of your existing ranking options to the blank option (or any other option actually), then click OK or CANCEL.
c. open up ELL settings again, and the option which you changed the ranking to becomes 0 (or whichever number you selected in step b).

My suggestion: It seems that these preferences are only used inside the "Audio preview window". If that's the only place it is used, I think the ranking which user puts is not very important. maybe if lets say I rank 3 voices as 2, 3, 5, the settings can auto update it for me to 1, 2, 3. Also to solve the problem of the OK button causing some undefined behaviour, maybe allowing the "OK" button to update the preferences as well will be good. if "Cancel" is clicked, preferences should revert to the last confirmed preferences.

LiXiaoooowei:
@blewjy Thanks for your suggestions! I made changes according to some of your suggestions. 

For **point 1 on removal of blank items**, I feel that it is possible if user simply wants to sync custom animations over to elearninglab pane first before making any changes to self explanation item. However, if user changes to the next slide, then I assume that they have finished editing self explanation items, and any blank items will be removed after that.

For **point 2**, I disabled the selection for `other azure voices` if user is not logged in as user cannot choose azure voices from dropdown box in that case. However, disable azure voice radio button selection in general can pose two concerns:
1. Confusion when initializing `Audio Preview Page`. The voice selected as shown on `Audio Preview Page` should be the same as the voice selection that user chose last time. However, if user chose azure voice last time, and logged out subsequently, the radio button for that azure voice should still be selected when `Audio Preview Page` opens. Upon clicking on 'OK' button, the selected voice label should be set for that self explanation item still (now user have selected azure voice even if he is not logged in). To enforce consistency with this user case, I hence enable selection for all voices. 
2. `Audio Preview Page` can be considered as a helper page for user to make audio selection. Therefore, I think it might be ok for users to select azure voice even if they are not logged in. They can still log in afterwards from `elearning lab settings`.

For **point 3**, I fixed it according to your suggestion. Audio rankings will be updated even if `OK` button is clicked instead of `update audio preference` button. Audio rankings will revert back to the previous one if `cancel` button is clicked. However,  I still preserve the ranking of audio items, for example, `2 3 5` is not the same as `1 2 3`.  This is also to be consistent with user's common beliefs as most online survey forms with ranking questions do not automatically change `2 3 5` to `1 2 3`.

ChesterSng:
Hi @LiXiaoooowei, great job on the PR with the many UI changes! Found a few bugs when testing on PowerPoint 2016. 
 
**1. ELL stops functioning when slide gets deleted**
i. Open ELL 
ii. Delete any slide
iii. ELL stops functioning (all existing self-explanation item will disppear from ELL and all ELL buttons become unusable)

*Note: No problem occurs if ELL has not been opened yet. Once ELL is opened once, this bug will occur. If you run ppt from visual studio, an exception will be thrown.*

**2. Audio preview window is a separate "powerpoint" window**
i. Add self-explanation item
ii. Press the Voice Preview Button 
iii. Voice Preview Button will open, click anywhere outside of it (It will be sent to the background) 
iv. Try to click the Voice Preview Button again, there will be no response.

iv. Try to open settings from the ELL ribbon button, there will be an exception shown to the user. 
*Note: This will also happen to the azure login window. Consider the implementation of settings for ELL, the window cannot be sent to the background.*

**3. Caption text will not be copied to callout textbox**
> When user ticks use shorter callouts, caption text will be auto-copied to callout text box

i. Add self-explanation item, tick caption and callout 
ii. Enter something into the caption text box, tick shorter callouts
iii. Enter something into the callout text box **then** delete all content from callout text box
iv. Sync (Callout becomes unchecked)

*This is very minor and I think it might not even need to be addressed.*

**4. No feedback to user when invalid access key is used**
image
Clicking Confirm will just result in no response.

Some suggestions: 
**1. Voice preferences are not persistent**
I felt that voice preferences settings will be more meaningful if it is persistent.

**2. Numbering on the self-explanation item**
The new change to the way the numbering is displayed is rather small and hard on the eyes to see.

LiXiaoooowei:
@ChesterSng Thanks for testing!

**Fixed point 1 and 4. Enlarged label font size.**

For **point 2**, we copy caption text to callout text only when `use shorter callout` is first clicked. Afterwards, we preserve any changes that user did to callout text. If it is empty, then we will uncheck `iscallout` and do not create callout shape to slide. Let's leave the state as it is for now.

For **point 3,** let's discuss in a separate issue. This bug happens because `AudioPreviewWindow` runs on another thread, not from the main thread. Therefore, UI is not blocked. Closing `AudioPreviewWindow` before opening any other window will temporarily fix the exception during runtime. 

**Audio preferences are persistent by implementation**. If I click `OK` button on audio settings window, and then close the presentation and open it again, the same settings will be restored. @blewjy Does the same issue happen on your machine as well?

PowerPointLabs/PowerPointLabs/ELearningLab/Views/ELearningWorkspaceViews/ELearningLabMainPanel.xaml.cs
```diff
@@ -536,6 +592,18 @@ private bool IsAzureVoiceSelected()
             }
             return false;
         }
+        private bool IsSlideExist(int slideId)
```
leeyh20: Does `DoesSlideExist` sound better as the function name?

PowerPointLabs/PowerPointLabs/ThisAddIn.cs
```diff
@@ -733,7 +733,10 @@ private void UpdateELearningPane(int selectedSlidesCount)
             if (selectedSlidesCount > 0)
             {
                 taskpane.eLearningLabMainPanel1.HandleSlideChangedEvent();
-                taskpane.eLearningLabMainPanel1.HandleELearningPaneSlideSelectionChanged();
+                if (elearningLabPane.Visible == true)
```
leeyh20: Should this check also include line 735? Why is line 735  outside of this check?

PowerPointLabs/PowerPointLabs/ELearningLab/Views/ELearningWorkspaceViews/ELearningLabMainPanel.xaml.cs
```diff
@@ -73,11 +93,12 @@ public void ListViewItemPropertyChanged(object sender, PropertyChangedEventArgs
         public void HandleELearningPaneSlideSelectionChanged()
         {
             PowerPointSlide _slide = this.GetCurrentSlide();
-            if (_slide.ID.Equals(slide.ID))
+            if (IsSlideExist(slideId) && _slide.ID.Equals(slide.ID))
             {
                 return;
             }
             slide = _slide;
+            slideId = slide.ID;
```
leeyh20: Better to add some comments here to say that this is to only update the items when slide still exist and current slide is changed (?).

leeyh20: Could be a bit difficult to understand this function if the reader of the code did not follow the github discussion, thus comments might help.

PowerPointLabs/PowerPointLabs/ELearningLab/Views/ELearningWorkspaceViews/ELearningLabMainPanel.xaml.cs
```diff
@@ -536,6 +592,18 @@ private bool IsAzureVoiceSelected()
             }
             return false;
         }
+        private bool IsSlideExist(int slideId)
+        {
+            Slides slides = Globals.ThisAddIn.Application.ActivePresentation.Slides;
+            foreach (Slide slide in slides)
```
leeyh20: For a future version, you could consider optimizing this, rather than checking every slide in the presentation, how about responding to the event of deleting a slide and checking if deleted slide is the elearninglab slide? I'm not super sure either but seems like a good place to optimise later.

PowerPointLabs/PowerPointLabs/ELearningLab/Views/AudioSettingsViews/AudioMainSettingsPage.xaml.cs
```diff
@@ -170,6 +195,58 @@ private void PreviewCheckBox_Unchecked(object sender, RoutedEventArgs e)
...
+        {
+            ComboBox comboBox = sender as ComboBox;
+            if (comboBox.SelectedIndex > 0)
+            {
+                int rank = comboBox.SelectedIndex;
+                if (rankToComboBoxMapping.ContainsKey(rank))
+                {
+                    ComboBox previousComboBox = rankToComboBoxMapping[rank];
+                    if (!comboBox.Equals(previousComboBox))
+                    {
```
leeyh20: Would it be better to add some comments here for the nested if statements? It is hard to understand the code here...

PowerPointLabs/PowerPointLabs/ELearningLab/Converters/ELearningWorkspaceConverters/ListViewClickNoToColorConverter.cs
```diff
@@ -0,0 +1,65 @@
...
+using System.Windows.Media;
+using Microsoft.Office.Tools;
+using PowerPointLabs.ActionFramework.Common.Log;
+using PowerPointLabs.ELearningLab.ELearningWorkspace.Model;
+using PowerPointLabs.ELearningLab.ELearningWorkspace.Views;
+
+namespace PowerPointLabs.ELearningLab.Converters
+{
+#pragma warning disable 0618
+    public class ListViewClickNoToColorConverter : MarkupExtension, IMultiValueConverter
```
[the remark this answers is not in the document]
leeyh20: How about ClickNum?

PowerPointLabs/PowerPointLabs/ThisAddIn.cs
LiXiaoooowei: This check should not include line 735. This is because when slide selection is changed, two things will happen:
1. Check if previous slide contains any unsynced items. If so, we will prompt user to save changes.
2. Load e-learning lab with new items from current slide animations pane.

The first event should happen regardless of whether e-learning lab is visible. For example, if user opens e-learning lab and performs some changes, then closes it. A message should still prompt to user to sync.

PowerPointLabs/PowerPointLabs/ELearningLab/Views/ELearningWorkspaceViews/ELearningLabMainPanel.xaml.cs
LiXiaoooowei: Yes you are right about this. I also have the same concern. Probably this function needs to be removed in a future version. However, I could not seem to find a `slide deleted event` handler for powerpoint, therefore, I tried to iterate through all slides for now.  

Another possible alternative is to add a try-catch block when requesting for `slide.ID` property. However, an exception will still be thrown from visual studio during run time if we are requesting for invalid instance from powerpoint object model. A try-catch block will work fine if we open PPT  directly instead of running from visual studio.

PowerPointLabs/PowerPointLabs/ELearningLab/Views/AudioSettingsViews/AudioMainSettingsPage.xaml.cs
LiXiaoooowei: Comments added!

LiXiaoooowei: This method is trying to handle the case when item A has ranking 1, and subsequently user changes item B to have ranking 1. In that case, the ranking for item A should be removed (set to empty).

PowerPointLabs/PowerPointLabs/ELearningLab/Views/ELearningWorkspaceViews/ELearningLabMainPanel.xaml.cs
LiXiaoooowei: Thanks for the suggestion, comments added.

PowerPointLabs/PowerPointLabs/ELearningLab/Views/ELearningWorkspaceViews/ELearningLabMainPanel.xaml.cs
LiXiaoooowei: Yes it does sound better. Renamed to `DoesSlideExist`.

PowerPointLabs/PowerPointLabs/ELearningLab/Converters/ELearningWorkspaceConverters/ListViewClickNoToColorConverter.cs
```diff
@@ -0,0 +1,65 @@
...
+    {
+        public object Convert(object[] values, Type targetType, object parameter, CultureInfo culture)
+        {
+            bool isDummyItem = (bool)values[0];
```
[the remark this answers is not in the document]
LiXiaoooowei: Yes I didn't think about this. Added try-catch block.

PowerPointLabs/PowerPointLabs/ELearningLab/AudioGenerator/ComputerVoiceGenerator/Model/ComputerVoice.cs
```diff
@@ -9,9 +9,33 @@ namespace PowerPointLabs.ELearningLab.AudioGenerator
     public class ComputerVoice: IVoice
```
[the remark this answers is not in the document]
LiXiaoooowei: Yes you are right about this. An abstract class will be better since we also have common attributes. I have changed interface to abstract class.

PowerPointLabs/PowerPointLabs/ThisAddIn.cs
```diff
@@ -733,7 +733,10 @@ private void UpdateELearningPane(int selectedSlidesCount)
             if (selectedSlidesCount > 0)
             {
                 taskpane.eLearningLabMainPanel1.HandleSlideChangedEvent();
```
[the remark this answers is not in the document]
LiXiaoooowei: Thanks for the reminder. Variable renamed.

PowerPointLabs/PowerPointLabs/ELearningLab/Converters/ELearningWorkspaceConverters/ListViewClickNoToColorConverter.cs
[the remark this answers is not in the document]
LiXiaoooowei: Yes `ClickNum` is more descriptive of the variable. Renamed `ClickNo` to `ClickNum`.

PR #1819, opened 2019-03-19
#1814 Remove textbox from callout in TooltipsLab

### LINKED ISSUE #1814: Tooltips Lab: Remove additional textbox
TooltipsLab is currently adding a textbox to the generated callout shape and grouping them together as one callout. This may unnecessarily complicate things. Let's remove the additional textbox and just use the callout shape's default textbox for any text instead.

blewjy:
Fixes #1814 

Previously, there was a textbox added to the generated callout shape from TooltipsLab which may make things complicated for the user. This textbox has been removed in this PR and for any text, user will use the shape's default textbox instead.


ChesterSng:
@blewjy should we still focus on the textbox of the callout after it is created?

blewjy:
@ChesterSng Personally I don't think it is necessary to focus on the shape's textbox after the callout is created, because it is possible that the user may want to put other things inside the textbox other than text, for e.g. images.

Nevertheless I did give it a try and when trying to select the `TextFrame` of the shape, it seems to result in like a space character that is being highlighted for some reason, instead of being just a blinking cursor as expected... Typing immediately will revert it back to normal, i.e. there is no actual space character there.

#### Actual:


#### Expected:

blewjy:
If we want we can also choose to just select the callout shape

ChesterSng:
The text in the textbox will overflow the callout. I think you should make the callout shape resize when text overflows like in ELL. @blewjy 

Prof @damithc, should there be default text in the callout shape like in here: 
image

blewjy:
> The text in the textbox will overflow the callout. I think you should make the callout shape resize when text overflows like in ELL.

@ChesterSng good point! Fixed! :)

damithc:
> Prof @damithc, should there be default text in the callout shape like in here:

No need.

PR #1820, opened 2019-03-19
#1813 New Icons and Menu Structure

### LINKED ISSUE #1813: Tooltips Lab: New icons and menu structure
TooltipsLab is currently borrowing icons from animation lab.

Let's change to new icons according to this ppt:
[Icons-TooltipsLab.pptx](https://github.com/PowerPointLabs/PowerPointLabs/files/2965026/Icons-TooltipsLab.pptx)

At the same time, TooltipsLab menu will be changed to having just 4 options:
- Create Tooltip
- Create Trigger For Callout
- Create Callout For Trigger
- Convert Shapes To Tooltip

ChesterSng:
Fixes #1813

1.) Updated icons in Tooltips Lab to this: 
[Icons-TooltipsLab.1.pptx](https://github.com/PowerPointLabs/PowerPointLabs/files/2995834/Icons-TooltipsLab.1.pptx)

*The blue represents what the user provides to Tooltips Lab. 
The orange represents what Tooltips Lab generate for users.*

Here is how it looks like in PowerPoint:
ss

2.) Restructure Menu so that there is no more dropdown Create Menu as before. 

3.) Renamed AssignTooltip to ConvertToTooltip for consistency

4.) Removed settings since they are not implemented for now


blewjy:
Hi @ChesterSng, icons and code LGTM. However should we remove the Settings menu for the time being since it's not being implemented yet? Might be a little bit unnecessary for us to have a settings button that actualy doesn't do anything

ChesterSng:
Prof @damithc, are the icons ok?

damithc:
> Prof @damithc, are the icons ok?

Yes, I'm OK.

The icons seem to have assigned some special meaning to orange and blue colors, which is a bit arbitrary though. As a result, while the icons certainly distinguish between the features, they don't help to identify the feature (unless the user remembers the meaning of colors).

PowerPointLabs/PowerPointLabs/ActionFramework/TooltipsLab/CreateCallout/CreateCalloutActionHandler.cs
```diff
@@ -17,27 +18,19 @@ class CreateCalloutActionHandler : ActionHandler
...
-            {
-                return;
-            }
-
             PowerPoint.Selection selection = this.GetCurrentSelection();
 
-            if (selection.Type != PowerPoint.PpSelectionType.ppSelectionShapes)
+            if (currentSlide == null || !ShapeUtil.IsSelectionShape(selection))
             {
-                MessageBox.Show(TooltipsLabText.ErrorNoTriggerShapeSelected);
```
YuPeiHenry: Is there a reason for removing this error message?

PowerPointLabs/PowerPointLabs/ActionFramework/TooltipsLab/CreateTrigger/CreateTriggerActionHandler.cs
```diff
@@ -16,26 +18,18 @@ class CreateTriggerActionHandler : ActionHandler
...
-            {
-                return;
-            }
-
-            PowerPoint.Selection selection = this.GetCurrentSelection();
-
-            if (selection.Type != PowerPoint.PpSelectionType.ppSelectionShapes)
+            if (currentSlide == null || !ShapeUtil.IsSelectionShape(selection))
             {
-                MessageBox.Show(TooltipsLabText.ErrorNoCalloutShapeSelected);
```
YuPeiHenry: Same for this error message.

PowerPointLabs/PowerPointLabs/ActionFramework/TooltipsLab/CreateCallout/CreateCalloutActionHandler.cs
ChesterSng: Hi @YuPeiHenry the reason for removing this error message is because the buttons will no longer be enabled when there is no active shape selection.

PR #1821, opened 2019-03-20
dev-release-4.7.0.0

leeyh20:
dev-release-4.7.0.0

I suspect that the addition of NET Framework 4.5 is because of @LiXiaoooowei 's PR? We need to test it out first.


leeyh20:
Does anyone know why we have 3.5, 4.0 and now also 4.5 in our bootstrapper?

PR #1826, opened 2019-03-21
#1825 Add Progress Bar mode to TimerLab

### LINKED ISSUE #1825: TimerLab Feature Add-on: Progress Bar
**The Original TimerLab uses a bar to move from 0 to n seconds.**
TimerLab Original

**Suggested Add-on to Timerlab to resemble a progress bar.**
TimerLab Suggestion

**Reason:** 
Having an option to change it to the suggested add-on will enable user to make a "progress bar" on their slide quickly.

ChesterSng:
Fixes #1825 
**Outline of Solution**

**The progress bar is implemented the following way:**
1. Create two bars (each of same height and width of the timer body)
2. First bar is filled with colour, second bar is transparent.
3. Concatenate second bar to first bar and group the two bars
4. Align first bar with the timer body. 
5. Apply Shrink to 0 width animation to the progress bar over set duration. 

**Functional Tests are added for Progress Bar:**
1. Tests on the original timer lab are tested again when progress bar is enabled
2. Added test case when there is already a timer on the slide and progress bar is enabled/disabled
- [x] Implementation of Progress Bar
- [x] Adding Functional Test Cases for Progress Bar
- [x] Refactor Code


PowerPointLabs/PowerPointLabs/TimerLab/TimerLabPaneWPF.xaml.cs
```diff
@@ -97,11 +113,21 @@ private int LineMarkerColor()
...
-            AddTimerBody(timerWidth, timerHeight, timerLeft, timerTop, TimerBodyColor());
+            if (isProgressBar)
+            {
+                AddTimerBody(timerWidth, timerHeight, timerLeft, timerTop, SliderColor());
+                AddProgressBar(duration, timerWidth, timerHeight, TimerBodyColor(), SlideWidth());
+            }
+            else
+            {
+                AddTimerBody(timerWidth, timerHeight, timerLeft, timerTop, TimerBodyColor());
+                AddSlider(duration, timerWidth, timerHeight, ProgressBarTimerBodyColor(), SlideWidth());
```
blewjy: Did you mix up `ProgressBarTimerBodyColor()` and `TimerBodyColor()` here? Seems like you are using `ProgressBarTimerBodyColor()` when `isProgressBar` returns false

ChesterSng: The color returns correctly but the naming is wrong. Will change it!

PowerPointLabs/PowerPointLabs/TimerLab/TimerLabPaneWPF.xaml.cs
```diff
@@ -655,13 +786,77 @@ private void HeightTextBox_LostFocus(object sender, RoutedEventArgs e)
...
+                    continue;
+                }
+
+                if (shapesInSlide[i].Equals(invisibleProgressBar))
+                {
+                    indicesOfProgressBars[1] = i;
+                    continue;
+                }
+            }
+            progressBar = (this.GetCurrentSlide().Shapes.Range(indicesOfProgressBars)).Group();
```
blewjy: Instead of looping through all the shapes in the slide, you might want to try the following to group the two shapes:

```c#
string[] barNames = new string[] { visibleProgressBar.Name, invisibleProgressBar.Name };
progressBar = this.GetCurrentSlide().Shapes.Range(barNames ).Group();
```

I believe passing the array of names into the `.Range()` method also works. This way we can shorten the code and also make it more efficient :)

ChesterSng: Thanks @blewjy it works! Previously I tried passing it as a `Shape[]` which did not work out :(

PowerPointLabs/Test/FunctionalTest/TimerLabTest.cs
```diff
@@ -174,6 +212,145 @@ private void TestNonMultipleDenominationDurationWithCountdownTimer(ITimerLabCont
...
+
+            MessageBoxUtil.ExpectMessageBoxWillPopUp("Error",
+               "Only one timer allowed per slide.", timerLab.ClickCreateButton, "Ok");
+            AssertIsSame(PbOriginalSlideNo, PbRecreateBodySlideNo);
+        }
+
+        private void TestEditLineColorAndDeleteTimeMarkerPb(ITimerLabController timerLab)
+        {
+            PpOperations.SelectSlide(ChangeLineColorAndRecreateTimeMarkerSlideNo);
+            int expectedColor = PpOperations.SelectShape(TimerLineMarkerGroup)[1].Line.ForeColor.RGB;
```
YuPeiHenry: It seems this block of code has reuse value. There are some repetitions. I think it would be good to extract the code below into a reusable function, with a comment of what it is for.

```
            int expectedColor = PpOperations.SelectShape(TimerLineMarkerGroup)[1].Line.ForeColor.RGB;

            PpOperations.SelectSlide(PbOriginalSlideNo);
            ShapeRange lineMarkerGroup = PpOperations.SelectShape(TimerLineMarkerGroup);
            lineMarkerGroup.Line.ForeColor.RGB = expectedColor;
            ShapeRange timeMarkerGroup = PpOperations.SelectShape(TimerTimeMarkerGroup);
            timeMarkerGroup.Delete();
```

PowerPointLabs/Test/FunctionalTest/TimerLabTest.cs
```diff
@@ -174,6 +212,145 @@ private void TestNonMultipleDenominationDurationWithCountdownTimer(ITimerLabCont
...
+        {
+            PpOperations.SelectSlide(PbOriginalSlideNo);
+            timerLab.SetDurationTextBoxValue(0.07);
+            AssertIsSame(PbOriginalSlideNo, PbChangeDurationSlideNo);
+        }
+
+        private void TestEditTextColorAndDeleteProgressBarPb(ITimerLabController timerLab)
+        {
+            PpOperations.SelectSlide(ChangeTextColorAndRecreateSliderSlideNo);
+            int expectedColor = PpOperations.SelectShape(TimerTimeMarkerGroup)[1].TextFrame.TextRange.Font.Color.RGB;
```
YuPeiHenry: The same block of code repeated.

PowerPointLabs/Test/FunctionalTest/TimerLabTest.cs
ChesterSng: I think it is not very meaningful to abstract these functions. The level of abstraction seems right as in other functions as wel.

PR #1831, opened 2019-03-28
#1827 Add to common traps doc: Grouping shapes with ShapeRange

### LINKED ISSUE #1827: Add to docs: Grouping Shapes with Shape.Name
When developing new features, we found some difficulty in trying to group a bunch of `Shapes` together when we only have the reference to the `Shape` object. Some of the workarounds include trying to loop through all the shapes in the slide, selecting one by one, which is pretty inefficient and results in lengthy code.

A quick and easy way we found is to extract a `ShapeRange` object from the slide based on the names of the shapes.

```c#
string[] names = new string[] { shape1.Name, shape2.Name };
Shape group = this.GetCurrentSlide().Shapes.Range(names).Group();
```

Will be good to add this into our "Common add-in development traps" document for future contributors to take note!

blewjy:
Fixes #1827 

Just a small addition to our `Common Development Traps` document for grouping of `Shapes` with only a reference to them.

### DIFF
--- a/doc/CommonTraps.md
+++ b/doc/CommonTraps.md
@@ -49,7 +49,7 @@ This document details the common pitfalls when developing for PowerPointLabs, so
 	
 	Step 4. Select Single startup project and choose PowerPointLabs from the drop down list.
 
-1. To group multiple `Shape` objects together when you only have the references to the objects, an easy way is to extract a `ShapeRange` object from the slide based on the names of the shapes, and then call the `Group()` method on that `ShapeRange`
+1. When grouping multiple `Shape` objects together, it is not straightforward to do so with only references to the `Shape` objects. An easy way is to extract a `ShapeRange` object from the slide based on the names of the `Shapes`, and then calling the `Group()` method on that `ShapeRange`.
 ```c#
 string[] names = new string[] { shape1.Name, shape2.Name };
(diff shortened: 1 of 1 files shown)


ChesterSng:
Hi @blewjy, thanks for adding this to common traps! 

Right now `commontraps.md` seems to be an unordered list of tips, tricks and traps. Maybe in the future, we can split tips and tracks to a separate document so that the document name is more intuitive.

doc/CommonTraps.md
```diff
@@ -48,3 +48,9 @@ This document details the common pitfalls when developing for PowerPointLabs, so
 	Step 3. Navigate to Common Properties > Startup Project.
 	
 	Step 4. Select Single startup project and choose PowerPointLabs from the drop down list.
+
+1. To group multiple `Shape` objects together when you only have the references to the objects, an easy way is to extract a `ShapeRange` object from the slide based on the names of the shapes, and then call the `Group()` method on that `ShapeRange`
```
YuPeiHenry: There seems to be 2 main types of points: Tips when implementing functionality, and bugs. The tips seems to mostly start with `When`. Let me know what you think.

blewjy: I think you make a good point about the need to be consistent with the words used. Fixed.

We should also keep this in mind if we do intend to restructure the documents!

PR #1843, opened 2019-04-03
#1836 TooltipsLab: Unit Tests

### LINKED ISSUE #1836: Add Unit Tests for TooltipsLab
Since the functionalities of TooltipsLab is finalised, we can start adding unit tests to TooltipsLab.

blewjy:
Partially fixes #1836 

Added some UTs to Tooltips Lab. 

Features tested:
- Create Tooltip
- Create Trigger from Callout
- Create Callout from Trigger

UTs mainly tested if the tooltip is generated as expected, including when different types of shapes are selected, multiple shapes are selected, and also checking if the generated animations are as expected.

Tested on PPT2013.

**Remaining features to be tested:**
- Convert to Tooltip


ChesterSng:
Hi @blewjy, ran the UTs and they passed. 

Some suggestions for more test cases: 

1. CreateTooltip 

Add a test case where there are already existing tooltips. 
Add a test case where multiple tooltips are being generated.

2. CreateTrigger/CreateCallout

Add test cases where more types of shapes are used. (use lines etc)

blewjy:
@ChesterSng Thanks for you suggestions! I added the test cases you suggested except one.

> Add a test case where multiple tooltips are being generated.

I don't think we have this functionality for "Create Tooltip". "Create Tooltip" should only generate one tooltip.

ChesterSng:
@blewjy Yup, I meant generate successively on the same slide

blewjy:
@ChesterSng, o right, added!

On a side note I realised that when creating multiple tooltip they don't stack sideways like when for e.g. a shape is copy-pasted multiple times. Instead they get created on top of each other. Will open an issue for that

PowerPointLabs/Test/UnitTest/TooltipsLab/CreateCalloutTest.cs
```diff
@@ -0,0 +1,119 @@
...
+        {
+            TestCreateCallout_NormalShape();
+            TestCreateCallout_TriggerShape();
+            TestCreateCallout_CalloutShape();
+            TestCreateCallout_MultipleNormalShapes();
+            TestCreateCallout_MultipleAllTypeShapes();
+            TestCreateCallout_MultipleTypeMoreShapes();
+        }
+
+        private const int CreateCalloutNormalShapeTestSlideNo = 4;
```
ChesterSng: Move these constants to on top all other functions and split the slides and names. Same for the other 2 unit tests!

blewjy: Done!

ChesterSng:
@leeyh20 lgty? 😄

PowerPointLabs/Test/UnitTest/TooltipsLab/CreateTooltipTest.cs
```diff
@@ -0,0 +1,91 @@
...
+        private void TestCreateTooltip_ExistingTooltip()
+        {
+            PpOperations.SelectSlide(CreateTooltipExistingTooltipTestSlideNo);
+            CreateTooltipAndCompare(CreateTooltipExistingTooltipTestSlideNo, CreateTooltipExistingTooltipExpectedSlideNo);
+        }
+
+        private void TestCreateTooltip_MultipleTooltips()
+        {
+            PpOperations.SelectSlide(CreateTooltipMultipleTooltipsTestSlideNo);
+            for (int i = 0; i < 5; i++)
```
leeyh20: the `5` could be in a variable like number of tooltips

blewjy: Done!

PowerPointLabs/Test/UnitTest/TooltipsLab/CreateTriggerTest.cs
```diff
@@ -0,0 +1,121 @@
...
+
+            AssertIsSame(testSlideNo, expectedSlideNo);
+        }
+
+        private void AssertIsSame(int actualSlideIndex, int expectedSlideIndex)
+        {
+            Slide actualSlide = PpOperations.SelectSlide(actualSlideIndex);
+            Slide expectedSlide = PpOperations.SelectSlide(expectedSlideIndex);
+            SlideUtil.IsSameLooking(expectedSlide, actualSlide);
+            SlideUtil.IsSameAnimations(expectedSlide, actualSlide);
```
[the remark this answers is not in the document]
blewjy: The UTs that check for animations are all using `IsSameAnimations`, hence I decided to just follow the naming for consistency. I think the naming is fine, but if we have to rename it we can open a separate issue for that and keep the naming consistent throughout the codebase.

PR #1851, opened 2019-04-06
#1850 Enhance eLL loading process

### LINKED ISSUE #1850: Enhance eLL loading algorithm
Previously eLL loads all items from PPT animation pane. The sequence of items are determined by the ordering on PPT animation pane whereas caption and callout texts for explanation items are grabbed from in-slide storage. 

**Problem with this approach:**
This approach works perfect when users do not manually change the ordering of `PPTL` animations.  For example, based on eLL syncing algorithm, all animations for each explanation item should appear under the same click number. However, if user manually messed up the animation order through PPT animation pane, then eLL will not load explanation items correctly (because eLL does not know how to group animations for each explanation item).
 
**Suggested approach:**
We can modify the eLL in-slide storage to store all information required to restore an explanation item. During eLL loading stage, eLL loads custom items from PPT animations pane, whereas explanation items are loaded from in-slide storage. Custom items and explanation items are combined together later based on their click numbers. 

**Note:**
Suggested approach also means that click numbers shown on eLL panel and PPT animation pane will be the same **ONLY IF** users do not manually re-order explanation item animations through PPT animation pane. Otherwise, the click numbers shown on both sides will be different.

LiXiaoooowei:
Fixes #1850

### DIFF
--- a/PowerPointLabs/PowerPointLabs/ELearningLab/Service/StorageService/ELearningLabTextStorageService.cs
+++ b/PowerPointLabs/PowerPointLabs/ELearningLab/Service/StorageService/ELearningLabTextStorageService.cs
@@ -18,17 +18,17 @@ namespace PowerPointLabs.ELearningLab.Service
 {
     public class ELearningLabTextStorageService
     {
-        public static void StoreSelfExplanationTextToSlide(List<ExplanationItem> selfExplanationClickItems, 
+        public static void StoreSelfExplanationTextToSlide(List<ExplanationItem> selfExplanationClickItems,
             PowerPointSlide slide)
         {
             string shapeName = ELearningLabText.ELearningLabTextStorageShapeName;
             slide.DeleteShapeWithName(shapeName);
             List<Dictionary<string, string>> selfExplanationText =
                 ConvertListToDictionary(selfExplanationClickItems);
-            XElement textInxml = new XElement(ELearningLabText.SelfExplanationTextIdentifier, 
-                selfExplanationText.Select( kv =>
-                 new XElement(ELearningLabText.SelfExplanationItemIdentifier, 
-                from text in kv select new XElement(text.Key, text.Value))));
+            XElement textInxml = new XElement(ELearningLabText.SelfExplanationTextIdentifier,
+                selfExplanationText.Select(kv =>
+                new XElement(ELearningLabText.SelfExplanationItemIdentifier,
+               from text in kv select new XElement(text.Key, text.Value))));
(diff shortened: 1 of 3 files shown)


LiXiaoooowei:
@ChesterSng Thanks for reviewing! ELL does not change or modify normal animations.  During syncing, normal animations will be synced over to eLL pane as well, but their orders are not retained. It is customer's responsibility to re-order normal animations on eLL, otherwise, normal animations will not appear at the desired order due to the insertion of eLL animations in between.

damithc:
> @ChesterSng Thanks for reviewing! ELL does not change or modify normal animations. During syncing, normal animations will be synced over to eLL pane as well, but their orders are not retained. It is customer's responsibility to re-order normal animations on eLL, otherwise, normal animations will not appear at the desired order due to the insertion of eLL animations in between.

eLL _reorders_ the custom animations? That's not good. It's fine for eLL to insert explanations in between, but _reordering_ is a problem. We are basically saying if the user reorders animations in the animation pane, she has to duplicate the same changes again in the eLL. That's unnecessary work for the user.

ChesterSng:
> eLL _reorders_ the custom animations? That's not good. It's fine for eLL to insert explanations in between, but _reordering_ is a problem. We are basically saying if the user reorders animations in the animation pane, she has to duplicate the same changes again in the eLL. That's unnecessary work for the user.

My bad, it doesn't reorder the custom animations, rather it retains the order of the custom animations and inserts self-explanation items in between as prof @damithc said.

blewjy:
Hi @LiXiaoooowei, 

>We are basically saying if the user reorders animations in the animation pane, she has to duplicate the same changes again in the eLL. That's unnecessary work for the user.

I seem to be encountering this situation as mentioned by prof @damithc.

If I were to have a simple callout animation (generated by ELL) and a fly-in animation (generated by myself using PPT), and then I decide to shift the fly-in animation to the end of the animation timeline via the **Animation Pane**, then i click on "Sync" on ELL, my fly-in animation will get pushed back to the start. This will be the behaviour unless I were to manually change the order on ELL as well (double work for the user).

**Before**
ell-loading-before

**After shifting my animation to the end**
ell-loading-after-swap

**After clicking "Sync" on ELL**
ell-loading-after-sync

LiXiaoooowei:
@blewjy Thanks for testing! That's the intended behavior. ELL only updates eLL animations by inserting eLL animations between custom animations. The order of custom animations are retained. 

> We might have to put it somewhere in the user guide that if reordering of ELL-animations is desired, it has to be done via the ELL pane only.

Yes I agree. I will state this behavior in user guide.

blewjy:
Great!

PR #1852, opened 2019-04-07
#1847 eLL: Copy shape for callout/caption across slides

### LINKED ISSUE #1847: Continue caption/callout style across slides
Version 4.9.0.0 (Dev-release)

Current: callout/caption style is copied from the previous callout/caption within the same slide, but not across slides. Example: 

1. add a callout to a slide using eLL
1. change the formatting of callout e.g., change font color
1. add another callout below it. observation: it inherits the custom formatting from the previous callout, as expected.
1. Go to the next slide and add a caption.
1. The callout takes the default formatting, instead of following the same custom formatting

image

Suggested: copy the custom formatting of the callout/caption immediately above it, even if it is in a different slide

LiXiaoooowei:
Fixes #1847 

Previous eLL only copies shape format for callout/caption within slides. This PR exists this behavior to copy shape format across slides.

**Outline of Solution**
Previous eLL only copies shape within slide because `Duplicate()` method only duplicate the shape if the shape exists in the current slide. Instead of duplicating shape, we use `copy` and `paste` instead, i.e. we simply need to copy the shape and paste it into the destination slide. In that way, we can copy shapes across slides.


LiXiaoooowei:
@ChesterSng Thanks for testing! Bug is fixed. However, the template to be copied will replace user clipboard. I am still investigating into better ways to fix this.

blewjy:
Hi @LiXiaoooowei, tested on PPT2013 and found a bug when slide is deleted:

***Steps to reproduce:***
a. Create a self-explanation item with callout/caption and sync.
b. Change the format of the shape(s) and sync again.
c. Create a new slide, then delete the original slide.
d. Create a new self-explanation on this new slide and sync. Exception will be thrown.
```
System.Exception
  HResult=0x80131500
  Message=Error copying shape.
  Source=PowerPointLabs
  StackTrace:
   at PowerPointLabs.ELearningLab.Utility.ShapeUtility.InsertTemplatedShapeToSlide(PowerPointSlide slide, Shape templatedShape, String shapeName, String text) in C:\Users\Bryan\Desktop\PowerPointLabs\PowerPointLabs\PowerPointLabs\ELearningLab\Utility\ShapeUtility.cs:line 98
   at PowerPointLabs.ELearningLab.Service.CaptionService.InsertCaptionShapeToSlide(PowerPointSlide slide, String captionText, Int32 tagNo) in C:\Users\Bryan\Desktop\PowerPointLabs\PowerPointLabs\PowerPointLabs\ELearningLab\Service\FunctionService\CaptionService.cs:line 68
   at PowerPointLabs.ELearningLab.Service.CaptionService.CreateAppearEffectCaptionAnimation(PowerPointSlide slide, String captionText, Int32 clickNo, Int32 tagNo, Boolean isSeperateClick) in C:\Users\Bryan\Desktop\PowerPointLabs\PowerPointLabs\PowerPointLabs\ELearningLab\Service\FunctionService\CaptionService.cs:line 23
   at PowerPointLabs.ELearningLab.Service.ELearningService.CreateAppearEffectAnimation(PowerPointSlide slide, ExplanationItem selfExplanationItem) in C:\Users\Bryan\Desktop\PowerPointLabs\PowerPointLabs\PowerPointLabs\ELearningLab\Service\ELearningService.cs:line 93
   at PowerPointLabs.ELearningLab.Service.ELearningService.SyncAppearEffectAnimationsForSelfExplanationItem(Int32 i) in C:\Users\Bryan\Desktop\PowerPointLabs\PowerPointLabs\PowerPointLabs\ELearningLab\Service\ELearningService.cs:line 46
   at PowerPointLabs.Views.ProcessingStatusForm.Worker_DoWorkForELearningLabService(Object sender, DoWorkEventArgs e) in C:\Users\Bryan\Desktop\PowerPointLabs\PowerPointLabs\PowerPointLabs\Views\ProcessingStatusForm.xaml.cs:line 59
```

PowerPointLabs/PowerPointLabs/ELearningLab/Utility/ShapeUtility.cs
```diff
@@ -88,7 +89,15 @@ public static Shape InsertTemplatedShapeToSlide(PowerPointSlide slide, Shape tem
...
-            Shape copiedShape = templatedShape.Duplicate()[1];
+            try
+            {
+                templatedShape.Copy();               
+            }
+            catch
+            {
+                throw new Exception("Error copying shape.");
+            }
+            Shape copiedShape = slide.Shapes.Paste()[1];
```
ChesterSng: Should the paste code be inside the try-catch block as well? Previously copy-paste caused some problems in `ZoomToArea` and `Crop`

leeyh20: Indeed... generally copy-paste might cause problems in certain machines, and especially will affect the clipboard too.

PowerPointLabs/PowerPointLabs/ELearningLab/Service/FunctionService/CalloutService.cs
```diff
@@ -52,9 +52,18 @@ private static Shape InsertCalloutShapeToSlide(PowerPointSlide slide, string cal
                 templatedShape = shape;
                 return shape;
             }
-            if (templatedShape != null && slide.ContainShape(templatedShape))
+            if (templatedShape != null)
             {
-                return ShapeUtility.InsertTemplatedShapeToSlide(slide, templatedShape, shapeName, calloutText);
+                Shape copiedShape;
```
YuPeiHenry: Is it better if the `if` branches are inverted? Using templatedShape == null as a guard clause with a 1 line return, then have the try catch block afterwards?

LiXiaoooowei:
@blewjy Thanks for testing! Added a try-catch block on that. 

Just a side note, I felt that the Exception Settings should not be set to break for `System.Exception` or `System.Runtime.COMException`. This is because Visual Studio will break whenever the exception is incurred, regardless or whether it is handled or not, see [here](https://docs.microsoft.com/en-us/visualstudio/debugger/managing-exceptions-with-the-debugger?view=vs-2019).

PowerPointLabs/PowerPointLabs/ELearningLab/Service/FunctionService/CalloutService.cs
LiXiaoooowei: You're right! That's much better. I checked for `templatedShape == null` in `ShapeUtility.InsertTemplatedShapeToSlide` method instead, and added try-catch block for that method when it was called.

PowerPointLabs/PowerPointLabs/ELearningLab/Utility/ShapeUtility.cs
LiXiaoooowei: @ChesterSng @leeyh20 Thanks for reminding! I enclosed the `paste` part to try-catch block as well.

PR #1864, opened 2019-04-11
#1863 Strikethrough voice label if the voice does not exist on local machine

### LINKED ISSUE #1863: e-Learning Lab: strikethrough voice label on explanation item if the voice label does not exist
**Problem**
When eLL loads an explanation item from slide, voice label for the item will be displayed on the UI. During syncing, eLL will fetch the voice from the list of available voices and generate audio using the voice fetched. However, we should also consider the case when the voice does not exist on user's machine. This could happen if user chose in-computer voice for an explanation item and subsequently opened the slide deck from another machine which did not contain the particular in-computer voice that the user chose previously for the item.

**Current eLL behaviour:**
- The invalid voice label is still loaded to eLL explanation item UI, i.e. 
- Audio will not be generated during syncing. 
- When audio preview window is opened from explanation item, `computer voice` will be checked with an empty value, i.e.

**Suggested behaviour:**
- Strikethrough the voice label on eLL explanation item if the voice does not exist on user's machine, i.e. 
- During syncing, automatically use default voice. 
- When audio preview window is opened from the explanation item, automatically select default voice instead, i.e.

LiXiaoooowei:
Fixes #1863


LiXiaoooowei:
To test this PR, one can open `selection pane` from `Home`, and click on a shape named `PPTL_Storage`. The in-slide storage will be visible. Next edit the value in `<VoiceLabel>` tag to any random name, and reload eLL.
`

ChesterSng:
Hi @LiXiaoooowei, tested and it works like you mentioned in this issue. 

I'm a bit curious to why there needs to be fix to this behaviour. Is the only reason for fixing this for the possiblity that a user will change the value of `VoiceLabel` in the in slide storage?

PowerPointLabs/PowerPointLabs/ELearningLab/ELearningWorkspace/Model/ClickItem/ExplanationItem.cs
```diff
@@ -113,6 +114,8 @@ public string VoiceLabel
             {
                 voiceLabel = value;
                 NotifyPropertyChanged("VoiceLabel");
+                NotifyPropertyChanged("IsTriggerTypeComboBoxEnabled");
```
ChesterSng: Should the strings here be refactored into the constants file for elearningLab?

PowerPointLabs/PowerPointLabs/ELearningLab/Service/ELearningService.cs
```diff
@@ -70,8 +70,19 @@ private void CreateAppearEffectAnimation(PowerPointSlide slide, ExplanationItem
...
             if (selfExplanationItem.IsVoice)
             {
-                Effect effect = AudioService.CreateAppearEffectAudioAnimation(slide, selfExplanationItem.CaptionText, selfExplanationItem.VoiceLabel,
-                    selfExplanationItem.ClickNo, selfExplanationItem.tagNo, isSeparateClick);
+                Effect effect;
+                if (selfExplanationItem.IsVoiceLabelInvalid)
+                {
+                    string defaultVoiceLabel = string.Format(ELearningLabText.AudioDefaultLabelFormat, AudioSettingService.selectedVoice.VoiceName);
+                    effect = AudioService.CreateAppearEffectAudioAnimation(slide, selfExplanationItem.CaptionText, defaultVoiceLabel,
+                     selfExplanationItem.ClickNo, selfExplanationItem.tagNo, isSeparateClick);
```
ChesterSng: the indentation looks a bit weird here (1 space only)

PowerPointLabs/PowerPointLabs/ELearningLab/Service/ELearningService.cs
```diff
@@ -70,8 +70,19 @@ private void CreateAppearEffectAnimation(PowerPointSlide slide, ExplanationItem
...
+                    selfExplanationItem.VoiceLabel = defaultVoiceLabel;
+                }
+                else
+                {
+                    effect = AudioService.CreateAppearEffectAudioAnimation(slide, selfExplanationItem.CaptionText, selfExplanationItem.VoiceLabel,
+                      selfExplanationItem.ClickNo, selfExplanationItem.tagNo, isSeparateClick);
```
ChesterSng: same here

PowerPointLabs/PowerPointLabs/ELearningLab/ELearningWorkspace/Model/ClickItem/ExplanationItem.cs
LiXiaoooowei: Thanks for the suggestion! Refactored!

PowerPointLabs/PowerPointLabs/ELearningLab/Service/ELearningService.cs
LiXiaoooowei: Thanks for noticing! Indented properly.

PowerPointLabs/PowerPointLabs/ELearningLab/Service/ELearningService.cs
LiXiaoooowei: Indented

PR #1866, opened 2019-04-11
#1835 Use ItemSource to reduce Code Duplication

### LINKED ISSUE #1835: Use ItemSource to reduce code duplication
## Code quality
Use BindingSource to reduce code duplication

#### Description of Feature
Currently code is duplicated for each color palette in ColorsLab. See [link](https://github.com/PowerPointLabs/PowerPointLabs/blob/1a9df66592611eca484b81c9fbda46ed7c0f009e/PowerPointLabs/PowerPointLabs/ColorsLab/ColorsLabPaneWPF.xaml#L348)

#### Proposal
Use `.ItemSource` to populate and auto-update each item. 

Example of an application of `.ItemSource` for similar purposes: [example](https://www.wpf-tutorial.com/listview-control/listview-data-binding-item-template/)

YuPeiHenry:
Fixes #1835

Ready for Review. Ran ColorsLab test on PPT 2016.
**Outline of Solution**
The current implementation uses `ObservableCollection` to store the colors. It is possible to further reduce duplication in the xaml file (using ItemSource), but maintainers won't be able to see the actual panel being reflected in Visual Studio.


PowerPointLabs/PowerPointLabs/DataSources/ColorDataSource.cs
```diff
@@ -1,515 +1,183 @@
...
-        private HSLColor themeColorSixValue;
 
-        public HSLColor ThemeColorSix
-        {
-            get
-            {
-                return themeColorSixValue;
-            }
-            set
+            for (int i = index - 1; i > 0; i--)
```
blewjy: This line of code is causing the recent colors not to update properly. Should start from `index` instead of `index - 1`

PowerPointLabs/PowerPointLabs/ColorsLab/ColorsLabPaneWPF.xaml.cs
```diff
@@ -1265,57 +1237,17 @@ private void EmptyThemePanel()
...
+            dataSource.SetFavoriteColor(column, color);
         }
 
         /// <summary>
         /// Save current panel as default theme color.
         /// </summary>
         private void SaveDefaultColorPaneThemeColors()
         {
-            dataSource.SaveThemeColorsInFile(_defaultThemeColorDirectory);
+            dataSource.SaveFavoriteColorsInFile(_defaultThemeColorDirectory);
```
leeyh20: Default color pane theme colours are related to favourite colors now?

YuPeiHenry: @leeyh20 In the code it seems that they are mean the same thing. Did I misinterpret it?

YuPeiHenry: The change here is because I renamed the other function. Should I rename this as well in this PR?

YuPeiHenry:
@blewjy It seems 2 of the bugs are caused by saving issues and are fixed (serialize a List rather than ObservableList so that it is compatible with old files). The last one is caused by `index - 1` typo, which is also fixed.

PowerPointLabs/PowerPointLabs/ColorsLab/ColorsLabPaneWPF.xaml.cs
blewjy: @leeyh20 @YuPeiHenry Yes, i believe they mean the same thing. When I was porting over i think I may have messed up some of the naming and it became inconsistent. Will be good if you can standardise this, Henry.

blewjy: Let's use `Favourite Colors` since that's the label on the pane

PR #1909, opened 2019-05-13
#1902 ui problem using dark theme

### LINKED ISSUE #1902: UI problem when using the the dark theme
A user sent me this screenshot. 

image

Any idea how to reproduce, and solve this problem?

yuhongherald:
Fixes #1902 #1924 
- [x] Event to detect theme change
- [x] Get color scheme of application theme
- [x] Change PPL color on theme change.

- Some elements do not change properly (textbox placeholders, winform)
- Write tests for adaptive colors for support components. (Future TODO)

- Traverse elements using their Visual trees instead of Logical Trees.
- ListBox requires watching `ItemContainerGeneratorChanged` for new items.
- Frame requires watching `ContentRendered` before it is populated with elements.
- Refreshing visuals (#1924) requires watching `Load` to force a repaint using `invalidateVisual`.
- The styling is now unchanged if the application starts with a default (colorful) theme and only changes if it is changed or starts in another theme.

Caveats:
Previously I watched `CollectionChanged` for `INotifyCollectionChanged(listBox.Items)` but this didn't work for E-Learning Lab as it uses templates for items.

Using the ItemContainerGenerator to retrieve a newly created item returns null using this method, as the handler is notified too early. ItemContainerGeneratorChanged is notified more frequently, and can wait for `.Status` to be `.ContainersGenerated` for it to be finalized.

Using the newly created items from the arguments only retrieves the first element (strangely) so I used index instead.

- As Visual Trees tend to be large, optimized my code and added a IsUpdated check to skip traversal.

Old log:

Method used to achieve each step:

1. Detecting color theme change:

This was done following the instructions from [here](https://princetonits.com/detecting-office-theme-change-event-vsto-addin-registry-value-change-event-c/), with the logic extracted into `RegistryWatcher`.

A brief summary is that the RegistryWatcher watches a specific key, in this case, the UI Theme for Powerpoint, and notifies all its subscribers when the key has changed.

2. Get color scheme of application theme:

I have not found a good method for getting the colors for the application theme. Using the registry values for the different themes from the previous step, `RegistryWatcher` notifies `ThemeManager` to update the color scheme using a method `UpdateColorTheme` before notifying other subscribers.

3. Change PPL color on theme change.

Current proposed solution:
Adding helper methods `AddTaskPane` and `RemoveTaskPane` in `ThisAddIn` so that the event handler registration is tied with the pane registration.

Use of interface `IWPFContainer` and extension of `WIndows.Controls.Control`:
+ Makes it non-compulsory, it is also hard to miss out when implementing.
+ Duplicate logic is localized to a single area.

A theme change is invoked immediately on registration so that the `UpdateColor` code can be reused for initialization.

The updating of colors is done manually by element type.

Alternative solution:
Add and remove event handlers in the constructor and destructor method.
Pros: It looks cleaner.
Cons: Inherited parent constructors will run first, which means that the adding and removing of event handlers need to be done at the end of the construction. New classes need to be programmed with this in mind.

Problems encountered:
- The assigned `SolidColorBrush` for the colors need to be constructed on the same thread which the `Windows.Controls.Control` class is constructed on. One way to do this is:
```
if (!element.Dispatcher.CheckAccess())
{
    element.Dispatcher.Invoke(() => element.UpdateColors(sender, e));
    return;
}
```
- As the updating of colors is done manually by element type, I have also not added support for ListView elements.


blewjy:
Hey @yuhongherald, I've not tested this yet (I currently do not have access to my windows comp), but it looks like you only changed the background color for `ResizeLab`. Is only `ResizeLab` affected for this dark theme issue? What about other labs with panes?

Alternatively, we can also explore trying to detect the current active theme in the PowerPoint application? Although it might be a little difficult... I couldn't find anything related to the application theme in the microsoft docs, however I did find this [link](https://princetonits.com/detecting-office-theme-change-event-vsto-addin-registry-value-change-event-c/) that involves checking the registry keys. Not sure how difficult this will be to implement, but if we can detect the theme, we can then set the font colors accordingly.

yuhongherald:
@blewjy I have added adapative color scheme support. Is it ok if I don't write tests for detecting UI color change?

blewjy:
@yuhongherald Great job! Tested on PPT 365 and looks great! The colors are updating as expected on the different themes. I think we can leave it now and do more manual testing first before we write some test cases for this. I'm not very sure if we are able to change PPT theme colors programmatically, but we can explore that some other time. 

A couple of things before we merge this:
1. The colors are updating as expected except for e-Learning Lab and "Record Audio Manually" for Narrations Lab. Opening the manual recording pane actually throws and exception and the pane is not opened:
```
System.NullReferenceException
Object reference not set to an instance of an object.
Stack Trace:
   at PowerPointLabs.ColorThemes.Extensions.ThemeExtensions.UpdateColors(Control element, Object sender, ColorTheme e) in C:\Users\bryan\Desktop\PowerPointLabs\PowerPointLabs\PowerPointLabs\ColorTheme\ThemeExtensions.cs:line 21
   at PowerPointLabs.ColorThemes.ThemeManager.add_ColorThemeChanged(EventHandler`1 value) in C:\Users\bryan\Desktop\PowerPointLabs\PowerPointLabs\PowerPointLabs\ColorTheme\ThemeManager.cs:line 48
   at PowerPointLabs.ThisAddIn.AddTaskPane(UserControl control, Control wpfControl, String title, DocumentWindow wnd) in C:\Users\bryan\Desktop\PowerPointLabs\PowerPointLabs\PowerPointLabs\ThisAddIn.cs:line 511
   at PowerPointLabs.ThisAddIn.RegisterTaskPane(UserControl control, String title, DocumentWindow wnd, Control wpfControl, EventHandler visibleChangeEventHandler, EventHandler dockPositionChangeEventHandler) in C:\Users\bryan\Desktop\PowerPointLabs\PowerPointLabs\PowerPointLabs\ThisAddIn.cs:line 396
   at PowerPointLabs.ThisAddIn.RegisterRecorderPane(DocumentWindow activeWindow, String tempFullPath) in C:\Users\bryan\Desktop\PowerPointLabs\PowerPointLabs\PowerPointLabs\ThisAddIn.cs:line 308
   at PowerPointLabs.ActionFramework.NarrationsLab.RecordNarrationsActionHandler.ExecuteAction(String ribbonId) in C:\Users\bryan\Desktop\PowerPointLabs\PowerPointLabs\PowerPointLabs\ActionFramework\NarrationsLab\RecordNarrations\RecordNarrationsActionHandler.cs:line 29
   at PowerPointLabs.ActionFramework.Common.Interface.ActionHandler.Execute(String ribbonId) in C:\Users\bryan\Desktop\PowerPointLabs\PowerPointLabs\PowerPointLabs\ActionFramework\Common\Interface\ActionHandler.cs:line 14

```

2. For e-Learning Lab, when the slide has no self-explanation items, the pane has a large white box (for the dark themes). It would be ideal if this could be made to change according to the theme as well. I'm not too sure how feasible this is because i believe it is part of one of the winforms elements.

yuhongherald:
ok noted. The large white box is a ListView, will add support for it next week. I will take a closer look at the null reference exception.

yuhongherald:
@blewjy  Narrations Lab still uses WinForms, so my current implementation is unable to give it adaptive colors.

yuhongherald:
I have some difficulty in improving the performance of color theme changing. Currently, if I optimize the performance, the colors for e-Learning Lab is a bit weird.

1. The borders used for the circle is colored like a normal border, so it is no longer visible.
2. The textbox for "use a different text for the callout" does not appear correctly, the most newly created entry always has a placeholder text and a different background color.
image
Will be looking into it. Sorry for the long delay.

yuhongherald:
@blewjy

Some difficulties I had when implementing adaptive colors:

I needed to change the StyleCop library to allow type switch case. If there's a need to revert, please let me know.
The borders used for the circle is colored like a normal border, so it is no longer visible.
The textbox placeholder text disappears.
Tooltips lab setting panel doesn't look right.
I have to comment out the resource dictionary used for styling for e-Learning Lab settings.
If it is possible, we can open these as new issues.

PowerPointLabs/PowerPointLabs/ActionFramework/Common/Extension/ActionFrameworkExtensions.cs
```diff
@@ -105,6 +106,7 @@ public static CustomTaskPane GetTaskPane(Type taskPaneType)
                 DocumentWindow activeWindow = Globals.ThisAddIn.Application.ActiveWindow;
 
                 return Globals.ThisAddIn.RegisterTaskPane(taskPaneControl, taskPaneTitle, activeWindow,
+                    (control is IWPFContainer) ? ((IWPFContainer)control).WpfControl : null,
```
YuPeiHenry: May want to extract this to a variable.

PowerPointLabs/PowerPointLabs/ActionFramework/NarrationsLab/RecordNarrations/RecordNarrationsActionHandler.cs
```diff
@@ -14,19 +15,14 @@ class RecordNarrationsActionHandler : ActionHandler
     {
         protected override void ExecuteAction(string ribbonId)
         {
-            //TODO: This needs to improved to stop using global variables
```
YuPeiHenry: Any reason for removing this? The `this.GetCurrentPresentation().Presentation` below looks like accessing a global variable, and its still required.

PowerPointLabs/PowerPointLabs/ActionFramework/NarrationsLab/RecordNarrations/RecordNarrationsActionHandler.cs
```diff
@@ -41,5 +37,26 @@ protected override void ExecuteAction(string ribbonId)
...
+
+            if (recorderPane == null)
+            {
+                return;
+            }
+
+            RecorderTaskPane recorder = recorderPane.Control as RecorderTaskPane;
+
+            // trigger close form event when closing hide the pane
+            if (!recorder?.Visible ?? false)
```
YuPeiHenry: I think a `null` check here is more direct.

PowerPointLabs/PowerPointLabs/ActionFramework/NarrationsLab/RecordNarrations/RecordNarrationsActionHandler.cs
```diff
@@ -41,5 +37,26 @@ protected override void ExecuteAction(string ribbonId)
...
+            {
+                recorder.RecorderPaneClosing();
+                // remove recorder pane and force it to reload when next time open
+                // TODO: Callback to remove task pane from thisaddin, register event.
```
YuPeiHenry: Can summarize the list of TODOs in the PR?

PowerPointLabs/PowerPointLabs/ActionFramework/NarrationsLab/TempStorage.cs
```diff
@@ -0,0 +1,193 @@
...
+
+        private Presentation _currentPresentation;
+        private string _tempPath;
+
+        public string TempPath
+        {
+            get
+            {
+                CheckPresentationChanged();
+                if (_tempPath == null) { throw new NullReferenceException("TempStorage has not been initialized! Call Setup before using."); }
```
YuPeiHenry: Please move the thrown exception to a new line.

PowerPointLabs/PowerPointLabs/ActionFramework/NarrationsLab/TempStorage.cs
```diff
@@ -0,0 +1,193 @@
...
+                return _tempPath;
+            }
+        }
+
+        private void CheckPresentationChanged()
```
YuPeiHenry: Consider renaming the function. It does not return a boolean reflecting on the state of the presentation changing. It seems that it performs some actions should the presentation get changed.

PowerPointLabs/PowerPointLabs/ActionFramework/NarrationsLab/TempStorage.cs
```diff
@@ -0,0 +1,193 @@
...
+            string presFullName = pres.FullName;
+            string presName = pres.Name;
+            string zipFullPath = tempPath + tempZipName;
+
+            // in case of embedded slides, we need to regulate the file name and full name
+            RegulatePresentationName(pres, tempPath, ref presName, ref presFullName);
+
+            try
+            {
+                if (IsEmptyFile(presFullName))
```
YuPeiHenry: Can I check if `IsEmptyFile` has a possibility of throwing an exception?

PowerPointLabs/PowerPointLabs/ActionFramework/NarrationsLab/TempStorage.cs
```diff
@@ -0,0 +1,193 @@
...
+        {
+            // this function is used to handle "embed on other application" issue. In this case,
+            // all of presentation name, path and full name do not match the usual rule: name is
+            // "Untitled", path is empty string and full name is "slide in XX application". We need
+            // to regulate these fields properly.
+
+            presName = presName.AppendIfAbsent(fileSuffix);
+
+            if (tempPath != null)
+            {
```
YuPeiHenry: Change it to a guard clause to reduce nesting. `if (tempPath == null) return`

PowerPointLabs/PowerPointLabs/ActionFramework/NarrationsLab/TempStorage.cs
YuPeiHenry: Can ignore this actually, since Narrations Lab will be discontinued.

PowerPointLabs/PowerPointLabs/ActionFramework/NarrationsLab/RecordNarrations/RecordNarrationsActionHandler.cs
YuPeiHenry: Can ignore this, narrations lab will be discontinued.

PowerPointLabs/PowerPointLabs/ActionFramework/NarrationsLab/TempStorage.cs
YuPeiHenry: Can ignore this, narrations lab will be discontinued.

PowerPointLabs/PowerPointLabs/ActionFramework/NarrationsLab/TempStorage.cs
YuPeiHenry: Can ignore this, narrations lab will be discontinued.

PowerPointLabs/PowerPointLabs/ActionFramework/NarrationsLab/TempStorage.cs
YuPeiHenry: Can ignore this, narrations lab will be discontinued.

PowerPointLabs/PowerPointLabs/ActionFramework/NarrationsLab/TempStorage.cs
```diff
@@ -0,0 +1,193 @@
...
+                foreach (ZipStorer.ZipFileEntry entry in dir)
+                {
+                    string name = Path.GetFileName(entry.FilenameInZip);
+
+                    if (name == null)
+                    {
+                        continue;
+                    }
+
+                    if (name.Contains(".wav") ||
```
YuPeiHenry: Can use optional chaining `name?.Contains`. Can ignore, narrations lab will be discontinued.

PowerPointLabs/PowerPointLabs/Utils/RegistryWatcher.cs
```diff
@@ -0,0 +1,99 @@
...
+                T key = GetKeyAndUpdateKeyStatus();
+                if (IsDefaultKey)
+                {
+                    return;
+                }
+                ValueChanged?.Invoke(this, key);
+            }
+            catch (Exception)
+            {
+
```
YuPeiHenry: Empty catch clause. Should anything be done on an exception?

PowerPointLabs/PowerPointLabs/ActionFramework/NarrationsLab/RecordNarrations/RecordNarrationsActionHandler.cs
yuhongherald: It is an extension wrapper that hides the global implementation on another level. I can add it back.

PowerPointLabs/PowerPointLabs/ActionFramework/NarrationsLab/RecordNarrations/RecordNarrationsActionHandler.cs
yuhongherald: I will be removing this TODO in favor of the old one in this file.

yuhongherald:
@YuPeiHenry I did some corrections and removed some redundant code, any other changes required?

yuhongherald:
I have added a check for the Powerpoint version and also a default color for unknown themes. @ChesterSng it should work now.

yuhongherald:
@YuPeiHenry I have added a catch to the statement in the constructor of RegistryWatcher, as the key might not have an entry initially. Let me know if it still gives problems.

yuhongherald:
Tested and working on PPT2016.

blewjy:
@yuhongherald, a couple of UI bugs:

1. The borders in Resize lab is not showing, whereas in PositionsLab is ok.

Capture

2. Adding new shapes in ShapesLab causes the newest shape label to have a white background, which is not dismissable.

Capture

3. It seems that the header background color for the different groups in each pane are now the same color as the rest of the pane. Will it be possible to retain the color, but change the text color?

For e.g. in the screenshot below, the background is black, but just change the font color to white when this theme is detected

Capture

yuhongherald:
I have made changes based on your feedback:
1. The theme does not get affected if PPT starts in the default color theme.
2. Backgrounds are now consistent in color, and borders are made transparent (to avoid inconsistencies).

PR #1912, opened 2019-05-18
#1911 Explanation items does not load correctly

### LINKED ISSUE #1911: eLearningLab: Click number and trigger type of explanation items do not load correctly
## Bug Report  
Currently, eLearningLab does not load click number and trigger type of explanation items correctly if there are no more custom animation items following after explanation items. In that case, the explanation item that follows will always be set to trigger with previous. 
PowerPoint 2016
#### Steps to replicate
1. Create an explanation item and choose either callout/caption/voice, and change the item to start on click. 
2. Click sync to sync explanation item to the slide.
3. Switch to some other slide, eLL will reload.
4. Switch back to the previous slide where explanation item was created. Notice that the explanation item is set to trigger with previous and click number 0.  
#### Expected behavior
The same explanation item should be loaded, which was set to trigger on click and a click number of 1
#### Actual behavior
The explanation item was loaded to trigger with previous and click number 0

LiXiaoooowei:
Fixes #1911 

The problem is fixed by setting trigger type and click number of explanation items based on the information retrieved from in-slide storage. The code for creating an explanation item from in-slide storage is further refactored into a separate method.

### DIFF
--- a/PowerPointLabs/PowerPointLabs/ELearningLab/Views/ELearningWorkspaceViews/ELearningLabMainPanel.xaml.cs
+++ b/PowerPointLabs/PowerPointLabs/ELearningLab/Views/ELearningWorkspaceViews/ELearningLabMainPanel.xaml.cs
@@ -254,7 +254,6 @@ namespace PowerPointLabs.ELearningLab.Views
         }
         private ObservableCollection<ClickItem> LoadItems(DoWorkEventArgs e)
         {
-            DateTime start = DateTime.Now;
             SelfExplanationTagService.Clear();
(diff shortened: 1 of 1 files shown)


YuPeiHenry:
@LiXiaoooowei I have tested your fix and it works. Can you change the name of the branch to include the issue number at the front? Like `1911-eLearningLab-explanations-always-start-with-1`. Thanks you.

LiXiaoooowei:
@YuPeiHenry Thanks for testing! I will close this PR and re-open another PR #1920 with branch name prefixed with the issue number.

Issue #1525, opened 2017-07-20
Testing: Implement a way to get WPF control handles


initialshl:
We currently have ways to get WinForm handles (Test.Util.MessageBoxUtil.ExpectMessageBoxWillPopUp()), but our UI have been converted into WPF. 

We need a way to get the WPF control handles so that we can send message (such as clicking the OK button) to Settings Dialogs and other WPF dialogs.

leeyh20:
@initialshl  I saw this existing code in SyncLabController,
```
UIThreadExecutor.Execute(() =>
                {
                    _pane.SyncPaneWPF1.Dialog.okButton.RaiseEvent(new RoutedEventArgs(ButtonBase.ClickEvent));
                });
```
Does this already solve the problem of getting the WPF control handles?

initialshl:
Hm maybe @IanTeo can answer this

yuhongherald:
I noted that all windows created are derived from System.Windows.Window from my PR on #1902 . Will be looking deeper into the code.

PR #1915, opened 2019-05-23
#1913 Error in Zoomlab step back

### LINKED ISSUE #1913: Error in Zoom Lab: Step Back
Sent by a user. Error details:

```
System.Runtime.InteropServices.COMException
Shapes (unknown member) : Invalid request.  Clipboard is empty or contains data which may not be pasted here.
Stack Trace:
   at Microsoft.Office.Interop.PowerPoint.Shapes.Paste()
   at PowerPointLabs.ZoomLab.AutoZoom.GetStepBackWithBackgroundShapeToZoom(PowerPointSlide currentSlide, PowerPointSlide addedSlide, Shape previousSlidePicture, Shape& backgroundShape)
   at PowerPointLabs.ZoomLab.AutoZoom.AddStepBackAnimation(Shape selectedShape, PowerPointSlide currentSlide, PowerPointStepBackSlide& addedSlide, Boolean includeAckSlide, Boolean deletePreviouslyAdded)
```

yuhongherald:
Fixes #1913

- [x] Investigate the bug
- [X] Provide fix for placeholder text copy (Invalid Access Operation)
- [x] Apply placeholder text copy to other parts of PPL.
- [x] Run UTs & FTs
- [x] Write wrappers for Shapes.AddPicture and AddPicture2, AddMediaObject2

UPDATE on wrappers:
It seems to be wrongly assumed that the adding of audio shape throws an exception becuase the text
is not in English, but rather, it is due to `AccessViolationException` caused by `Shapes.AddMediaObject2`
when there is no read permission.

The clipboard error is likely to be due to interference from other programs. To combat this problem, a possible solution is to wait for the clipboard to be vacant, lock the clipboard and release it after the operation.

For the fixing of placeholder text, I wrote a method `SafeCopy` in `PowerpointExtensions`.
- If the shape is a placeholder and has text, a textbox will be created, with the placeholder created using the existing ShapeUtil

Some things to note:
- The following classes seem to require the clipboard for long periods of time, so they are unable to use the new implementation. Errors regarding the use of clipboard may persist through them:
ClipboardUtil
PowerPointShapeGalleryPresentation

- PowerpointShapeGalleryPresentation has a huge reliance on data on the clipboard.
- EffectsLab test threshold needs to be lowered for it to pass
- Autonarration test requires waiting for a while before checking that the number of animations is the same so that the animations will be added.
- There was an error due to selecting shapes not on an active slide after switching to `SafeCopy`. A future `SafeSelect` KIV.


yuhongherald:
It seems that upgrading beyond .NET 3.5 have caused AccessViolationException to become uncatchable. I will be extracting the try catch AccessViolationException into a method for easy editing in the future

yuhongherald:
@YuPeiHenry can you check if the AccessViolationException persists and if there is any other bugs?

yuhongherald:
The clipboard bug was due to the test running on a different process but trying to use the Powerpoint process hwnd to lock the clipboard.

There is still a small issue of ZoomToArea test being off by 0.8 pixels, which is more than the threshold of 0.5 pixels. I'm not sure how it happened as I don't seem to have modified any relevant code.

PowerPointLabs/PowerPointLabs/ActionFramework/Common/Extension/PPLClipboard.cs
```diff
@@ -0,0 +1,204 @@
...
+            /*
+            foreach (string lFormat in lFormats)
+            {
+                if (lDataObject.GetDataPresent(lFormat))
+                {
+                    lBackup.Add(lFormat, lDataObject.GetData(lFormat, false));
+                }
+            }
+            */
+            //Clipboard.Clear(); // to prevent any weird data from showing up
```
YuPeiHenry: Any reason for the commented code?

PowerPointLabs/PowerPointLabs/ActionFramework/Common/Extension/PPLClipboard.cs
```diff
@@ -0,0 +1,204 @@
...
+        // referred to https://stackoverflow.com/questions/6262454/c-sharp-backing-up-and-restoring-clipboard
+        // not working stably
+        public void RestoreClipboard()
+        {
+            /*
+            foreach (KeyValuePair<string, object> pair in lBackup)
+            {
+                lDataObject.SetData(pair.Key, pair.Value, false);
+            }
+            */
```
YuPeiHenry: Any reason for the commented code?

PowerPointLabs/PowerPointLabs/ActionFramework/Common/Extension/PowerpointExtension.cs
```diff
@@ -0,0 +1,192 @@
...
+
+        private static void TryAndCatch(Action a)
+        {
+            try
+            {
+                a();
+            }
+            catch
+            {
+                // do nothing
```
YuPeiHenry: Any reason for catching without any logging?

PowerPointLabs/PowerPointLabs/ActionFramework/Common/Extension/PowerpointExtension.cs
```diff
@@ -0,0 +1,192 @@
...
+            {
+                TryAndCatch(() => i.SetValue(target, i.GetValue(source)));
+            }
+        }
+
+        private static void TryAndCatch(Action a)
+        {
+            try
+            {
+                a();
```
YuPeiHenry: Will giving the variable name `action` be better?

PowerPointLabs/PowerPointLabs/Utils/FileDir.cs
```diff
@@ -205,6 +206,29 @@ public static void DeleteFile(string filePath)
...
+        /// </summary>
+        public static void TryDeleteFile(string filePath)
+        {
+            try
+            {
+                DeleteFile(filePath);
+            }
+            catch (Exception)
+            {
+            }
```
YuPeiHenry: Any reason why the catch clause is empty?

PowerPointLabs/PowerPointLabs/ActionFramework/Common/Extension/PPLClipboard.cs
yuhongherald: the code is suggested by an online post but seems to break the clipboard

PowerPointLabs/PowerPointLabs/ActionFramework/Common/Extension/PowerpointExtension.cs
yuhongherald: When doing reflection, exceptions can occur quite often. I will add the logging

PowerPointLabs/PowerPointLabs/Utils/FileDir.cs
yuhongherald: added logging

yuhongherald:
The unit tests are a bit unstable currently because of how PPLClipboard is initialized/teardown.

yuhongherald:
`SyncLabTest` fails sometimes because the WPF UI has not fully initialized due to the race condition introduce here:
```
            new Task(() =>
            {
                ThreadUtil.WaitFor(1000);
                syncLab.DialogClickOk();
            }).Start();
            syncLab.Copy();
```
This issue is addressed in #1525. Other than that, the tests work fine for PPT2016

blewjy:
No problems when testing with 2016. @YuPeiHenry @ChesterSng Possible to test on 2010 and 2013?

@yuhongherald There are some merge conflicts

blewjy:
Tested on PPT2013. Not able to reproduce any of the issues listed in #1913, so that's good. Will merge this and any further issues to address as a separate issue.

PR #1925, opened 2019-06-10
#1525 Testing implement wpf control handles

### LINKED ISSUE #1525: Testing: Implement a way to get WPF control handles

yuhongherald:
Fixes #1525

**PR Checklist**
- [x] Add method to get window handles
- [x] Improve tests to fail when exception windows appear.
- [x] Integrate with existing tests
- [Separate PR] Replace Winform elements with WPF elements

Getting window handles:
- HWND: Using `AutomationEventHandler`, the test will be notified on any new windows that are opened.
- Window identity: The identity of the window is verified using its `Title` and `processId`.
- Notifying main test thread: A `ManualResetEventSlim` is used to block the main thread until a suitable window is detected.
- The HWND is used to retrieve the `Window` on the application side using `HWNDSource.FromHwnd`, which is then passed back with Object Marshaling.

Compared to the old method, this is event-driven and does not employ busy waiting, which reduces test time and gives more control.

Automatic Window closing:
- Using a WindowWatcher, any windows that are not whitelisted will be closed, preventing error windows from freezing the test. Windows that will throw an exception will propagate the exception to the test.

About Object Marshaling:
It is possible to retrieve the window handle, but operations specific to the wpf window must be done on the application side (in the Powerpoint project) due to object Marshaling restrictions. Most wpf windows do not support `Serialization`. `MarshalWindow` has been written as a wrapper class to support generic operations using C# Reflection methods.

Compared to using user32 Window APIs, this method is more extensible, as a constants table can be declared using `nameof`.

There are some Winform elements that have yet to be replaced due to their convenient templates:
- MessageBox
- ColorDialog


blewjy:
@yuhongherald Code looks good, however the tests don't seem to be working well.

From the current dev-release (5.1.0.1), I am able to run all the FTs and UTs together -- meaning just right-click FTs/UTs header, then `Run selected tests`, allows me to run all of them at once without stopping (although there are a couple of tests failling).

However on your PR, I am unable to do so for the UTs. The tests get stuck at a couple of different UTs (`TestSyncLineDashType` is one of them), although running those tests that get stuck individually is fine. 

Is this happening for you as well? Do you know what's causing this? :o

PowerPointLabs/Test/Util/WPFWindowUtil.cs
```diff
@@ -0,0 +1,85 @@
...
+using TestInterface.Windows;
+
+namespace Test.Util
+{
+    static class WPFWindowUtil
+    {
+        public static string After(this string original, string searchTerm)
+        {
+            int index;
+            return (index = original.LastIndexOf(searchTerm)) != -1 ? original.Substring(index + searchTerm.Length) : "";
```
YuPeiHenry: Might it be better if the assignment to index is done on the previous line? `int index = original.LastIndexOf(searchTerm))`

yuhongherald:
I'm able to run all my unit tests at once, but ConvertShapesToTooltip fails because there is a dialog box that is expected in it, but `WindowWatcher` is not initialized.

PowerPointLabs/Test/Util/WindowWatcher.cs
```diff
@@ -0,0 +1,116 @@
...
+                AutomationElement element = sender as AutomationElement;
+                if (element.Current.ProcessId != processId &&
+                Process.GetProcessById(element.Current.ProcessId).ProcessName != "POWERPNT") { return; }
+
+                IntPtr handle = new IntPtr(element.Current.NativeWindowHandle);
+                string windowName = WindowUtil.GetWindowTitle(handle);
+                if (windowName == "")
+                {
+                    // Can't be sure what this is
+                    return;
```
YuPeiHenry: If you can't be sure what this is, might logging/throwing an exception be useful for troubleshooting?

PowerPointLabs/PowerPointLabs.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 14
-VisualStudioVersion = 14.0.24720.0
+# Visual Studio 15
```
YuPeiHenry: Any idea why the solution file is changed? Is this required?

yuhongherald:
@blewjy I have added a hotfix in unit tests to close windows with matching titles, just a note that it is _not the POWERPNT_ process.

PowerPointLabs/PowerPointLabs.sln
yuhongherald: nope, it's not required

PR #1936, opened 2019-06-24
#1935 create self explanation test passes despite showing different results

### LINKED ISSUE #1935: E-Learning Lab: CreateSelfExplanationTest passes despite showing different results
## Bug Report 
Actual:
image
Expected:
image

Reason for difference:
The shorter callout explanation does not show in place of the original text.

This error can also be observed in normal use:

## Steps to replicate
1. Open E-Learning Lab
2. In the first entry on the right, check callout and type "a" into the textbox.
3. Press the Sync button.
4. Check the use a different text for the callout. The second textbox should show "a".
5. Press the Sync button.
6. Replace the second textbox's content with "aa".
7. Press Sync.

Expected:
image

Actual:
image

The reason for the difference is likely because after Syncing, the SelfExplanationItem does not remember that the "Use a different text for a callout" is checked.

yuhongherald:
Fixes #1935

- [x] Find the reason for the textboxes stacking and fix it
- [x] Modify the test to catch the old bug

Investigation:
- The reason for abnormal behavior in the test is because HasShortVersion (now IsShortVersionIndicated) is not set to true.
- The positioning of the textboxes was wrong due to the order of positioning before sizing the textbox to fit the text.

Changes:
- The HasShortVersion is now a computed property to keep data consistent, and IsShortVersionIndicated will now notify UI on change (if it has been accidentally modified it will still be correct).
- The position of textboxes have been reordered

Not sure why the callout boxes are still slightly off though.


PowerPointLabs/PowerPointLabs/ELearningLab/Utility/ShapeUtility.cs
```diff
@@ -32,10 +32,12 @@ public static Shape InsertDefaultCalloutBoxToSlide(PowerPointSlide slide, string
             Shape calloutBox = slide.Shapes.AddShape(MsoAutoShapeType.msoShapeRoundedRectangularCallout, 10, 10, 100, 10);
             calloutBox.Name = shapeName;
             calloutBox.TextFrame.TextRange.Text = calloutText;
-            calloutBox.Left = 10;
-            calloutBox.Top = 10;
             calloutBox.TextFrame.AutoSize = PpAutoSize.ppAutoSizeShapeToFitText;
             calloutBox.TextEffect.Alignment = MsoTextEffectAlignment.msoTextEffectAlignmentLeft;
+
+            calloutBox.Left = 10;
+            calloutBox.Top = 10;
```
YuPeiHenry: So moving these 2 lines of code fixes the issue.

PowerPointLabs/PowerPointLabs/ELearningLab/Utility/ShapeUtility.cs
```diff
@@ -107,6 +109,12 @@ public static Shape InsertTemplatedShapeToSlide(PowerPointSlide slide, Shape tem
                 throw new Exception("Copied shape is null");
             }
             copiedShape.Name = shapeName;
+
+            copiedShape.TextFrame.TextRange.Text = text;
+            copiedShape.TextFrame.WordWrap = MsoTriState.msoTrue;
+
+            copiedShape.TextFrame.AutoSize = PpAutoSize.ppAutoSizeShapeToFitText;
```
YuPeiHenry: Do these lines need to be moved?

yuhongherald:
Tested on 2016, seems to be working, though the result is _slightly_ different from the old expected result. Should I remake the slides for expected @blewjy ?

blewjy:
@yuhongherald Which test are you referring to?

yuhongherald:
@blewjy CreateSelfExplanationTest

PR #1940, opened 2019-06-26
#1900 colors lab eyedropper timer not updating

### LINKED ISSUE #1900: Cannot pick a color using ColorsLab
PowerPointLabs Version 5.0.0.0 (Dev-release)
Win10, PPT2013

When trying to pick a color, the the button cannot be dragged. The preview circle stops moving almost immediately after I start dragging the button.
image

I have this problem in my home computer (PPT 2013), but not in my office computer (PPT 2016)

yuhongherald:
Fixes #1900

The main change is that the `DispatcherTimer` is now constructed using the same `Dispatcher` as the `Control` that owns it. Some other modifications include:

- Wrapping brush modification with `.Invoke`, as it can only be modified on the same `Dispatcher` thread. This is to prevent modification from failing if it had been accessed from a different thread.
- PositionsLab timer is now object level instead of class level


PowerPointLabs/PowerPointLabs/ThisAddIn.cs
```diff
@@ -1167,7 +1167,8 @@ private void ThisAddInShutdown(object sender, EventArgs e)
             PPMouse.StopHook();
             PPKeyboard.StopHook();
             PPCopy.StopHook();
-            PositionsPaneWpf.ClearAllEventHandlers();
+            //PositionsPaneWpf.ClearAllEventHandlers();
```
YuPeiHenry: @yuhongherald  Is there a reason for leaving this line as a comment?

yuhongherald: The clearing of event handlers have been moved into the destructor after changing the events from class level to object level.

yuhongherald:
Tested on 2016, works as it does previously

PR #1943, opened 2019-06-27
#1854 create tooltip stack instead of cascade

### LINKED ISSUE #1854: TooltipsLab: "Create Tooltip" will always create in the same exact position
**On TooltipsLab:**

If user clicks on "Create Tooltip" multiple times to create multiple tootlips, the generated tooltips just get stacked on top of each other instead of stack sideways similar to how copy and paste works:
tooltips-stafcking

We should make it stack sideways in the event that user generates multiple tooltips in a row so that it is easier to grab the generated shapes and move them around.

yuhongherald:
Fixes #1854

**PR Checklist**
- [x] Displace tooltip if it stacks
- [x] Write (modify existing) tests to ensure it works (UT)

The newly created trigger is displaced if there is another shape which is not itself, the same type, same size and same position. This is done until there is no such other shape, or until a maximum number of tries.
tooltip-cascade


yuhongherald:
Testing on 2016 and is working as expected

blewjy:
Nice! Very cool! Works as expected on PPT2016. Neat indeed. Tests are passing and feature works well. Code LGTM.

PR #1946, opened 2019-07-02
#1856-ResizeLab changes orientation of text in shape

### LINKED ISSUE #1856: ResizeLab changes orientation of text in shape
## Bug Report  
 
Before using ResizeLab:
image
After using ResizeLab:
image

PowerPoint 2016, Windows 10
#### Steps to replicate
Use any function in `ResizeLab` and this will change the orientation of the text to horizontal.

yuhongherald:
Fixes #1856

- [x] Add a fix to keep text rotation
- [x] Add tests to check that fix is working

Investigation:
I observed that the shape loses its rotation when it goes through the `PPShape` constructor.

Changes:
To preserve the text rotation, the original rotation is added to the text's Z rotation.


PowerPointLabs/Test/UnitTest/PPShape/PPShapeTest.cs
```diff
@@ -0,0 +1,67 @@
+﻿using Microsoft.Office.Interop.PowerPoint;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+using PowerPointLabs.Models;
+using PowerPointLabs.TooltipsLab;
+using PowerPointLabs.Utils;
+using Test.Util;
+
+namespace Test.UnitTest.TooltipsLab
```
blewjy: Wrong namespace? This test shouldnt be part of tooltipsLab

yuhongherald: ok noted

PR #1949, opened 2019-07-03
#1855 step back does not account for fading animations

### LINKED ISSUE #1855: Zoom Lab: Step back does not account for fading animations
#### Description of feature
Currently Zoom Lab creates an image of a slide without accounting for fade animations when using step back. Drill Down does some accounting for animations.

Slide with fade animations:
animation

Slide with step back:
animation2

#### Justification
The slide transition will not be smooth if extra effort is not placed in.

yuhongherald:
Fixes #1855

- [x] Add a method to create a slide without entry animation shapes

To account for fading animations, I have added a method that deletes all shapes with entry animations before converting the slide to a shape.

Limitations of current implementation:
- The current method checks against a dictionary of animations to check if it is an entry animation.
- Shapes containing entry text will be missing in step back transition as there is no way to set the fill of the text to no fill.
- The current test methods compare the slides visually, which will not detect if it is actually done correctly.

I am also not sure why is there a .vs file deleted


blewjy:
@yuhongherald  Hmm... Any idea what exactly the .vs config file is for? It looks kinda important from a quick google. Will it better if you close this PR + branch and reopen another with the same changes? Or does making the change always delete this file?

yuhongherald:
I have done another PR #1960. The file was deleted because I had to reconfigure my PPL project (and deleting the .vs file is part of the solution).

PR #1966, opened 2019-07-09
#1531 convert to picture for shapes in a group

### LINKED ISSUE #1531: Convert to picture for shapes in a group
Currently, we a shape in a group is selected for convert to picture, the entire group is converted to picture. We should convert the selected shape(s) only, instead of the entire group.

We can use EditName as a reference for this implementation. It might be difficult as we have to regroup the items

yuhongherald:
Fixes #1531

- [x] Allow user to convert single shape in group to image
- [x] Write tests to ensure fix is working

When a shape in the group is selected, it will be reflected as the child shaperange. However, when I do so, the shape detaches from the group.
demo


blewjy:
@yuhongherald tested on PPT2013, working as expected but only for a **single** shape. From your gif above, try selecting 2 or more shapes within that group, then convert --> exception will be thrown:

```
System.ArgumentException
The specified value is out of range.
Stack Trace:
   at Microsoft.Office.Interop.PowerPoint.ShapeRange.Regroup()
   at PowerPointLabs.ShortcutsLab.ConvertToPicture.GetShapeFromShapeRange(ShapeRange shapeRange) in C:\Users\ppt2013\Desktop\PowerPointLabs\PowerPointLabs\PowerPointLabs\ShortcutsLab\ConvertToPicture.cs:line 125
   at PowerPointLabs.ShortcutsLab.ConvertToPicture.GetShapeFromSelection(Selection selection) in C:\Users\ppt2013\Desktop\PowerPointLabs\PowerPointLabs\PowerPointLabs\ShortcutsLab\ConvertToPicture.cs:line 102
   at PowerPointLabs.ShortcutsLab.ConvertToPicture.Convert(PowerPointPresentation pres, PowerPointSlide slide, Selection selection) in C:\Users\ppt2013\Desktop\PowerPointLabs\PowerPointLabs\PowerPointLabs\ShortcutsLab\ConvertToPicture.cs:line 25
   at PowerPointLabs.ActionFramework.ShortcutsLab.ConvertToPictureActionHandler.ExecuteAction(String ribbonId) in C:\Users\ppt2013\Desktop\PowerPointLabs\PowerPointLabs\PowerPointLabs\ActionFramework\ShortcutsLab\ConvertToPicture\ConvertToPictureActionHandler.cs:line 21
   at PowerPointLabs.ActionFramework.Common.Interface.ActionHandler.Execute(String ribbonId) in C:\Users\ppt2013\Desktop\PowerPointLabs\PowerPointLabs\PowerPointLabs\ActionFramework\Common\Interface\ActionHandler.cs:line 14

```

Also:

1. So the expected behaviour is that once the selected shape(s) is/are converted, it will be removed from the group completely? I'm just thinking if it would better to keep them in the original group, but make a duplicate of the shapes and then convert that duplicate to picture. Or I could just be over-engineering this, haha. Let me know what you think @yuhongherald 

2. Whether we do support point 1 or not, we should support the multiple selected shapes thing :)

yuhongherald:
@blewjy thanks for testing! I have rewritten the implementation to duplicate the shapes if it has a parent shape (grouped) and also fixed the exception by using a different method to regroup. Hopefully it works now.

yuhongherald:
The implementation is changed to use the selection after ungrouping, since the childshaperange cannot be found on 2010.

PR #1968, opened 2019-07-10
#1918 text overflow cannot scroll

### LINKED ISSUE #1918: UI bug in E-Learning Lab: Text overflow cannot scroll
image

PPT2016, Win10

The name for the audio does not display fully and cannot be scrolled.

Suggested fix:
Wrap the text when the name overflows.

yuhongherald:
Fixes #1918

My proposed implementation to solve the problem is to truncate the name of the item if it overflows, so that the item sizes can be kept the same.

The implementation is similar to #1384, but the size of the parent element is also variable (the ListView), so wrapping the ListView with a DockPanel similarly seem to work

image


yuhongherald:
It does not seem to work on PowerPoint 2010, and the width does not update when the size of the pane changes.

PR #1993, opened 2019-08-01
#1992 ft support multiple versions

### LINKED ISSUE #1992: Test PowerPointLabs using specific PowerPoint version
On machines with multiple PowerPoint versions, the latest version is always used. An [article](https://social.msdn.microsoft.com/Forums/sqlserver/en-US/873234b1-b96b-4a70-8d3f-9690a78b6934/powerpoint-cannot-be-opened-by-specifying-the-version-c?forum=windowsgeneraldevelopmentissues) describes why.

To support testing PowerPoint on older versions, we should add a method to specify the PowerPoint version for:
- UTs
- FTs

yuhongherald:
Fixes #1992

Outline of solution:
To select a PowerPoint version, go to Test>Properties>Settings>PowerPoint_path.
To use newest PowerPoint version, leave PowerPoint_path blank. If you want to use a specific version, use the full executable path.

Other bugs fixed:
- Application is busy(2013): Seems to be caused by rapid closing of windows. A solution is to wait for the task that caused the window to open to complete, and to wait for the process to idle.


yuhongherald:
Tested FTs on 2013 (on a machine with both 2013 and 2016) and is working.

PR #1998, opened 2019-12-15
#1918 UI bug in E-Learning Lab: Text overflow cannot scroll

### LINKED ISSUE #1918: UI bug in E-Learning Lab: Text overflow cannot scroll

daekoon:
Fixes #1918 

**Outline of Solution**
Fixed the issue by making the name wrap if it overflows, as suggested in the issue tracker.
One problem faced while implementing this solution was that in order for the TextBlock to wrap, its width cannot be set to automatic. To circumvent this, I made the grid's column inherit the width of the ListView, and created a simple converter that subtracts 45 pixels from the value given. (45 pixels is the size of the icon on the left)

### DIFF
--- a/PowerPointLabs/PowerPointLabs/ELearningLab/Converters/ShapeNameWidthConverter.cs
+++ b/PowerPointLabs/PowerPointLabs/ELearningLab/Converters/ShapeNameWidthConverter.cs
@@ -1,0 +1,28 @@
+﻿using System;
+using System.Text.RegularExpressions;
+using System.Windows;
+using System.Windows.Data;
+
+namespace PowerPointLabs.ELearningLab.Converters
+{
+    public class ShapeNameWidthConverter : IValueConverter
+    {
+        public object Convert(object value, Type targetType, object parameter, System.Globalization.CultureInfo culture)
+        {
+            if (value is double)
+            {
+                return (double)value - 45;
+            }
+            return null;
+        }
+
+        public object ConvertBack(object value, Type targetType, object parameter, System.Globalization.CultureInfo culture)
+        {
+            if (value is double)
+            {
+                return (double)value + 45;
+            }
+            return null;
+        }
+    }
+}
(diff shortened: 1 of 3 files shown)


daekoon:
There seems to be a problem with the SonarQube during integration.

yuhongherald:
Noted the problem with SonarQube, will be trying to diagnose the problem.

PowerPointLabs/PowerPointLabs/ELearningLab/Converters/ShapeNameWidthConverter.cs
```diff
@@ -0,0 +1,28 @@
...
+
+namespace PowerPointLabs.ELearningLab.Converters
+{
+    public class ShapeNameWidthConverter : IValueConverter
+    {
+        public object Convert(object value, Type targetType, object parameter, System.Globalization.CultureInfo culture)
+        {
+            if (value is double)
+            {
+                return (double)value - 45;
```
yuhongherald: 45 is a constant that is used multiple times, you can declare a constant locally.

PowerPointLabs/PowerPointLabs/ELearningLab/Converters/ShapeNameWidthConverter.cs
```diff
@@ -0,0 +1,28 @@
+﻿using System;
+using System.Text.RegularExpressions;
```
yuhongherald: Unused import directives should be removed.

PowerPointLabs/PowerPointLabs/ELearningLab/Converters/ShapeNameWidthConverter.cs
daekoon: Replaced constant with a local constant

PowerPointLabs/PowerPointLabs/ELearningLab/Converters/ShapeNameWidthConverter.cs
daekoon: Removed unused directive

yuhongherald:
Great! Tested on PPT2013 and it works.

PR #2006, opened 2020-02-08
#1947 HighlightLab bug: Background & Point highlight overwrite each other

### LINKED ISSUE #1947: Highlight Lab: Background highlight disappears after adding point highlight
#### Environment
PPT2016 Win10 dev-release 5.1.0.1
#### Steps to replicate
1. Create a new slide, add a point
2. Select all in the text box with the point
3. Highlight background
4. Highlight point
The background highlight is observed to not appear at all during slide show.

This is not observed when highlight background is replaced with highlight text.

ccyccyccy:
Fixes #1947 

**Outline of Solution**
Removed lines which implemented deletion of other currently existing highlight animations.

Now it looks like this if background highlight is used after point highlight and vice versa:
ppt

### DIFF
--- a/PowerPointLabs/PowerPointLabs/HighlightLab/HighlightBulletsBackground.cs
+++ b/PowerPointLabs/PowerPointLabs/HighlightLab/HighlightBulletsBackground.cs
@@ -43,7 +43,6 @@ namespace PowerPointLabs.HighlightLab
                         break;
                     case HighlightBackgroundSelection.kNoneSelected:
                         currentSlide.DeleteIndicator();
-                        currentSlide.DeleteShapesWithPrefix("PPTLabsHighlightBackgroundShape");
                         shapesToUse = GetAllUsableShapesInSlide(currentSlide);
                         break;
                     default:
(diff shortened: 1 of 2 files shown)


yuhongherald:
Tested on PowerPoint 2013 and works as expected.

ChesterSng:
> ppt

If the previous highlight/text is not removed, then it yields a weird situation where the highlight on the first **bullet point background** appears first while the highlight on the **bullet points** iterate.

If we want to allow highlight background and highlight point to be used together, I suggest that we change highlight background's behaviour such that the highlight on the first bullet point's background does not appear until the first click. (Same as highlight point's behaviour)

Prof @damithc what do you think?

damithc:
> Prof @damithc what do you think?

Perhaps we can assume it is rare for the two to be used together and leave things be? No point complicating the code just to cater for a rare case.

ccyccyccy:
> Otherwise, imo, there is little use case for using both at the same time and we can retain the current behaviour but fix interaction between `highlight background` and `highlight text` instead.

I'm not too sure on what interaction you mean? They seem to work as expected. Should they remove each other's animation?

PR #2010, opened 2020-02-26
Add preview for background highlight points

### LINKED ISSUE #2009: Highlight background does not preview animation
Currently in the highlight lab, only `Highlight Points` display an animation preview after adding the effect, but `Highlight Background` does not, even though they serve a similar function.

We should add a preview for `Highlight Background` as well.

ccyccyccy:
Fixes #2009 

### Outline of Solution
Fixed the issue by checking if animation is successfully added to the slide, and then playing the animation preview.

### DIFF
--- a/PowerPointLabs/PowerPointLabs/HighlightLab/HighlightBulletsBackground.cs
+++ b/PowerPointLabs/PowerPointLabs/HighlightLab/HighlightBulletsBackground.cs
@@ -69,6 +69,12 @@ namespace PowerPointLabs.HighlightLab
                     AnimationLabSettings.IsUseFrameAnimation = oldValue;
                     PowerPointPresentation.Current.AddAckSlide();
                 }
+
+                if (currentSlide.HasAnimationForClick(clickNumber: 1))
+                {
+                    Globals.ThisAddIn.Application.CommandBars.ExecuteMso("AnimationPreview");
+                }
+
                 Globals.ThisAddIn.Application.ActiveWindow.Selection.Unselect();
             }
             catch (Exception e)


iskandarzulkarnaien:
Tested to be working on 2016, have you run the test suite?

ccyccyccy:
> 
> 
> Tested to be working on 2016, have you run the test suite?

Yep ran the test suite.

iskandarzulkarnaien:
> 
> 
> > Tested to be working on 2016, have you run the test suite?
> 
> Yep ran the test suite.

Aiight nice work!

xantho09:
Seems to be working on Office 365 (PPT 2019).

The change seems to preview all the animations in the slide (as opposed to only the animations added by Highlight Background). I realise the added lines are the same as those in `AddHighlightBulletsText()`. but is there a way to have only the newly added animations preview (e.g. when adding an animation via the standard Animations pane)?

xantho09:
> _Is there a way to have only the newly added animations preview (e.g. when adding an animation via the standard Animations pane)?_

I suppose this can be made into another issue instead of trying to remedy it here. This PR looks good to me.

ccyccyccy:
> > _Is there a way to have only the newly added animations preview (e.g. when adding an animation via the standard Animations pane)?_
> 
> I suppose this can be made into another issue instead of trying to remedy it here. This PR looks good to me.

Yeah I think so too. This behaviour is the same as the current one for point highlight

ccyccyccy:
Sorry for the late reply. How should I change the branch name?
According to https://stackoverflow.com/questions/20007578/renaming-a-branch-while-on-pull-request, I should create a new pull request?

YuPeiHenry:
Yeah it seems that the only way would be to open a new PR. You can reference this one for the previous reviews that are done.

ccyccyccy:
Closing due to opening another PR for this.

Issue #2002, opened 2020-01-14
'Replace with clipboard' feature seem to lose image quality


damithc:
Our right-click menu has a feature 'Replace with clipboard' that can be used to replaces an image on a slide with another image currently copied into the clipboard.
However, this feature seems to degrade the image quality of the pasted image i.e., compared to the quality you get when pasting directly

iskandarzulkarnaien:
I've discovered that the source of the bug is `ShapeUtil#ApplyAllPossibleFormats` which is being called inside `ReplaceWithClipboard#Execute`. 

@blewjy I don't quite understand the need to call `ApplyAllPossibleFormats` here since from my own testing it seems like its not needed, could you advise on this?

iskandarzulkarnaien:
Also it seems that `ReplaceWithClipboard` is doesn't work quite as intended on shapes with many different formats applied:

![image](https://user-images.githubusercontent.com/43085796/76416298-c4d99880-63d5-11ea-949c-c6da00495fab.png)

blewjy:
@iskandarzulkarnaien I just tested this and I'm encountering the same issues. I was able to reproduce Target 1 and Target 2 that you showed as well.

Commenting out `ApplyAllPossibleFormats` indeed seemed to fix the image quality issue, so something in there must be messing with the image quality. However I believe one of the original intention of `ApplyAllPossibleFormats` was to maintain the original formatting of the target shape/image whilst changing the content.

Try this for example:

Target image on the left, copied image on the right
![capt1](https://user-images.githubusercontent.com/39292809/76422087-9c569c00-63df-11ea-8e09-384c6b93e8d7.png)

Result:
![capt2](https://user-images.githubusercontent.com/39292809/76422125-ae383f00-63df-11ea-9d63-f76e576a9f64.png)

Without `ApplyAllPossibleFormats`, the green border is lost. The point of `Replace with Clipboard` in the first place is to maintain the formatting of the original shape/image, as well the animations, otherwise user can just simple Ctrl+C Ctrl+V, so this is a feature that we should keep. 

So although it seems a little unintuitive at first, actually the results that you showed in your last comment is actually intended. The image in Target 1 has no formatting, thats why only the text was kept. And in target 2 we kept the formatting (blue fill and nothing else) and replace the text.

My suggestion would be to see if we can fix the image quality issue from within the `ApplyAllPossibleFormats` call without changing the intended behaviour.. But open for discussion of course

[...]

PR #2013, opened 2020-03-16
#2002 replace with clipboard lose image quality

iskandarzulkarnaien:
Fixes issue #2002 . Source of picture degradation are the following formats being applied to pictures:
```
ContourColorFormat
ContourWidthFormat
DepthSizeFormat
LightingAngleFormat
MaterialEffectFormat
ThreeDRotationEffectFormat
```

Issue is fixed by introducing a check to prevent pictures from having such formats applied to them.

A picture degradation test has been added to PasteLabFT as well.

### DIFF
--- a/PowerPointLabs/PowerPointLabs/PasteLab/ReplaceWithClipboard.cs
+++ b/PowerPointLabs/PowerPointLabs/PasteLab/ReplaceWithClipboard.cs
@@ -15,7 +15,14 @@ namespace PowerPointLabs.PasteLab
         {
             // ignore height & width, it doesn't always make sense to sync the height & width especially for circles, squares
             List<Format> formatsToIgnore = new List<Format> {new PositionHeightFormat(), new PositionWidthFormat()};
-            
+
+            // the following effects are not useful for pictures and cause degradation of picture quality
+            List<Format> formatsToIgnoreIfPicture = new List<Format>
+            {
+                new ContourColorFormat(), new ContourWidthFormat(), new DepthSizeFormat(), new LightingAngleFormat(),
+                new MaterialEffectFormat(), new PositionHeightFormat(), new PositionWidthFormat(), new ThreeDRotationEffectFormat()
+            };
+
             // Replacing individual shape
             if (selectedChildShapes.Count == 0)
             {
@@ -32,7 +39,16 @@ namespace PowerPointLabs.PasteLab
 
                 slide.DeleteShapeAnimations(pastingShape);
                 slide.TransferAnimation(selectedShape, pastingShape);
-                ShapeUtil.ApplyAllPossibleFormats(selectedShape, pastingShape, formatsToIgnore);
+
+                if (pastingShape.IsPicture())
+                {
+                    ShapeUtil.ApplyAllPossibleFormats(selectedShape, pastingShape, formatsToIgnoreIfPicture);
+                }
+                else
+                {
+                    ShapeUtil.ApplyAllPossibleFormats(selectedShape, pastingShape, formatsToIgnore);
+                }
+
                 selectedShape.SafeDelete();
 
                 return slide.ToShapeRange(pastingShape);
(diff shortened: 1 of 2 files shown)


PowerPointLabs/Test/FunctionalTest/PasteLabTest.cs
```diff
@@ -180,15 +187,46 @@ private void IsClipboardRestoredPasteIntoGroup(int originalSlideNo, int expSlide
...
+            }
+            catch (AssertFailedException e)
+            {
+                if (e.Message.Equals("AssertNotSame failed. The slides look similar."))
+                {
+                    throw e;
+                }
+            }
+        }
+
```
iskandarzulkarnaien: I am aware that this is not the most elegant way to implement this method, but I wanted to avoid having to introduce lots of duplicate code for a method that is essentially the opposite of `AssertIsSame`. Any suggested improvements you guys have would be well appreciated.

xantho09: A possible slightly-more-elegant-but-still-kinda-not-really-that-elegant way to do it can be as such:

```csharp
try
{
    AssertIsSame(actualSlideNo, expectedSlideNo, similarityTolerance);
}
catch (AssertFailedException e)
{
    return;
}

throw new AssertFailedException("AssertNotSame failed. The slides look similar.");
```

xantho09:
Looks good on PPT 2019.

PowerPointLabs/Test/FunctionalTest/PasteLabTest.cs
iskandarzulkarnaien: Aiight thanks! This seems much cleaner

PowerPointLabs/PowerPointLabs/PasteLab/ReplaceWithClipboard.cs
```diff
@@ -74,10 +90,17 @@ static internal class ReplaceWithClipboard
...
                 {
-                    ShapeUtil.ApplyAllPossibleFormats(selectedChildShape, pastingShapes[i], formatsToIgnore);
+                    if (pastingShapes[i].IsPicture())
+                    {
+                        ShapeUtil.ApplyAllPossibleFormats(selectedChildShape, pastingShapes[i], formatsToIgnoreIfPicture);
+                    }
+                    else
+                    {
+                        ShapeUtil.ApplyAllPossibleFormats(selectedChildShape, pastingShapes[i], formatsToIgnore);
+                    }
```
YuPeiHenry: This logic seems to have appeared twice, but the duplicated code is short and only for 2 occurrences. Refactoring it into a separate function might make it a little less convenient to read though.

blewjy:
I encountered the same issue with the image resizing itself following the steps shown by @YuPeiHenry. I was using PPT2013 and this was the image i tested with https://wallpaperaccess.com/full/493303.jpg

iskandarzulkarnaien:
@YuPeiHenry, I suspect the exception thrown is related to #2017. I'm working on a fix for it, lets see whether it solves the problem.

@YuPeiHenry, @blewjy, does the resize problem occur on `dev-release`? I have no problems on 2016. Something to note is that it seems that 2016 does its own resizing for images when you ctrl-v them in, my images never seem to exceed the slide boundaries even when I use a 21:9 HD picture.

ReplaceWithClipboard seems to respect this automatic resizing if the source of the image is outside of PowerPoint.

Scenario 1:
1. Copy HD image from browser (I used @blewjy 's https://wallpaperaccess.com/full/493303.jpg)
2. Ctrl-V to paste into PowerPoint (notice that the image has resized itself to fit perfectly into the slide)
3. Repeat step 1
4. ReplaceWithClipboard on any shape. The pasted image will be the same size as the one in step 2.

Scenario 2:
1. Copy HD image from browser (used Bryan's one again)
2. Ctrl-V to paste into PowerPoint (again, image has resized itself)
3. Manually enlarge the image by dragging the anchor points
4. Ctrl-C to copy the image
5. ReplaceWithClipboard on any shape. The pasted image will be the same as in step 3.

blewjy:
> does the resize problem occur on `dev-release`?

Yes. Here's something interesting on PPT 2013: When i Ctrl+V that image on to the slide, it's larger than the slide and it doesn't auto resize itself. But after I copy the same image from the browser again and do Replace With Clipboard, the image becomes smaller, but it actually becomes exactly the size of the slide. So Replace With Clipboard in this case is sorta doing the "automatic resizing" for us i guess...?

iskandarzulkarnaien:
Hmm seems like a version-specific issue. What's your advice on how to proceed from here? In my opinion this issue would only annoy the user if they intended for the image to exceed the bounds of the slide, but I wouldn't think this applies to most people

blewjy:
@iskandarzulkarnaien let's keep this as it is for now since it only occurs on 2010 and 2013. 

@YuPeiHenry anything else? Otherwise we can merge this. The `System.AccessViolationException` exception I'm guessing is unrelated to this PR since it occurs on `dev-release` as well. Let's handle it in another PR.

YuPeiHenry:
@blewjy Yeah there are no other issues with the PR.

PR #2015, opened 2020-03-18
#1370 paste into group exception after undo

### LINKED ISSUE #1370: Paste Into Group has exception after undo
Paste Into Group throws an uncaught exception if I paste, undo and paste again

iskandarzulkarnaien:
Fixes #1370 by starting an undo entry within `PasteIntoGroup`.

### DIFF
--- a/PowerPointLabs/PowerPointLabs/ActionFramework/PasteLab/PasteIntoGroup/PasteIntoGroupActionHandler.cs
+++ b/PowerPointLabs/PowerPointLabs/ActionFramework/PasteLab/PasteIntoGroup/PasteIntoGroupActionHandler.cs
@@ -3,6 +3,7 @@
 using Microsoft.Office.Interop.PowerPoint;
 
 using PowerPointLabs.ActionFramework.Common.Attribute;
+using PowerPointLabs.ActionFramework.Common.Extension;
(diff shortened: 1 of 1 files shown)


blewjy:
Hm... I'm unable to reproduce the original bug. Could you list out some steps to reproduce? This is what I did:
1. Create 3 shapes, group the first 2.
2. Copy the third shape, right click on the group and do Paste Into Group. Success.
3. Ctrl-Z to undo. Third shape is removed from group.
4. Right click on the group and do Paste Into Group again. Success.

iskandarzulkarnaien:
@blewjy Hmm...that's strange. This bug is present in the Tutorial.pptx that comes with PowerPointLabs.

1. Copy the leaf in slide 53
2. Paste Into Group in slide 54. Success.
3. Ctrl-Z to undo. Leaf is removed from group.
4. Right click and Paste Into Group again. Exception Thrown.

blewjy:
Hmm interesting. I tried again on PPT2016 and I could reproduce it. No problems on PPT2013. Either way, this proposed fix works on PPT2016 and does not affect PPT2013, so it's good. Thanks!

blewjy:
I'll just merge this since it's a pretty trivial fix.

PR #2016, opened 2020-03-19
#2008 Improve UI Dark Theme

### LINKED ISSUE #2008: Improve UI appearance when using Dark Theme
## Bug Report  

While most of PowerPointLabs is usable when used with the Dark Theme in PowerPoint, some parts are still hard to use such as:

- When typing in a TextBox, the Caret is still Black

image

- The Checks in Checkboxes are hard to see. In SyncLab, this the Checks are not visible at all unless the mouse is hovering over them.

image

image

- The image previews in SyncLab are still in Black. (Refer to previous screenshot)
- When hovering over a Button, the text becomes hard to see.

image

 
Windows 10, Office 365

xantho09:
Fixes #2008  

**Outline of Solution**
1. Caret of `TextBox`

   This was fixed by setting the `CaretBrush` property to `theme.foreground` in `ThemeExtensions.cs`

1. Checks in `CheckBox`

   The check colour cannot be easily set; there is no simple `CheckColor` property in `CheckBox`. This requires modifying the Control Template of the `CheckBox`.

   Instead, the fix simply maintain the background colour of the `CheckBox` (as White). This way, the Check mark can always be seen.

1. Text of images in SyncLab

   Setting the image text's foreground to match the theme is not easy given the current method of applying themes to the `Control`s.

   Instead, the fix is to place the image around a border whose background remains White regardless of the theme. A special case had to be added in `ThemeExtensions.cs` to ensure that this border's background is not changed upon the theme changing.

1. Button colours

   Like the `CheckBox`, a `Button`'s colours upon mouse-over or click is not easily settable; the `Button`'s Control Template has to be rewritten. So, that's what is done.

   Firstly, an additional property called `ButtonTheme` is added to the `ColorTheme` struct. This contains all the colours of the background, foreground and border of the button when idle, upon mouse-over, when clicked, and when disabled.

   A copy of a default `Button`'s Control Template, excluding the colours set upon mouse-over, etc., is placed in a Resource Dictionary called `GeneralResourceDictionary.xaml`, placed within the Resources/ResourceDictionaries folder. When setting the theme, this base Control Template is retrieved and a new `Style` object is constructed using the `ButtonTheme` property of the `ColorTheme`. 

**Additional Notes**
The current method of applying the themes still feels a little clunky. 

The code in `ThemeExtensions.cs` will apply the same change for any element of a certain type. For example, any `Border` object will have its background set to `colorTheme.background`, regardless of its initial colour. If two `Border` objects have different backgrounds when designing within the .xaml file, this difference will no longer be reflected. This can be rectified by including an additional check within `ThemeExtensions.cs`, but it's not very convenient.

I plan to experiment with different ways of applying the themes in the hopes that I can find a more modular approach, with the end goal of ensuring the views' appearances in other themes are as good as those in the default "Colorful" theme.

There are some articles on the Internet about applying themes, but those are for `Application`s, which in turn, have an `App.xaml` file. PPT Labs does not have such a file, so those solutions are not immediately applicable here.


xantho09:
I just realised that there is a change to `doc/TechnicalFAQs.md` that I don't remember making and a few files to `doc/images/TechnicalFAQ` that I don't remember adding.

Should I delete those files and that block of text, or is there another way this should be remedied?

Also, it says there is a change to `PowerPointLabs.sln` despite the deleted and inserted text being exactly the same. What should be done for this?

yuhongherald:
For the sln file, it might have something to do with the line endings related to PR #1977. Not sure why it happens though.

For the TechnicalFAQ, those changes are from PR #2003, which is merged into master. Not sure why is it showing up here. @YuPeiHenry

YuPeiHenry:
I think you can keep the changes for `TechnicalFAQ`.

xantho09:
I was probably stupid and branched from `master` instead of `dev-release`. In that case, I'll leave the changes to the `TechnicalFAQ` as-is.

Also, this is a stupid question, but how do I fix the `.sln` file?

I've been playing around on the `dev-release` branch, trying to ensure that the `.sln` file remains unchanged. 
- I've tried configuring my `core.autocrlf` to `false`
- I've tried adding a `.sln text eol=crlf` line to the `.gitattributes` file
- I've tried manually replacing all the `CRLF` in the `.sln` file to `LF` (despite the initial file already using `CRLF`)

The `.sln` file is always shown to be "modified".

yuhongherald:
I'm also not very sure how to fix the problem. Maybe you can try looking at this?

https://docs.microsoft.com/en-us/visualstudio/ide/encodings-and-line-breaks?view=vs-2019

xantho09:
Closing this PR as #2021 is a more updated PR that solves the issue.

PR #2019, opened 2020-04-01
#1681 Optimize sync selection pane ui

### LINKED ISSUE #1681: SyncLab: Optimize UI for Sync Selection Pane
## Feature Request  
Optimize UI for Sync Selection Pane

#### Justification
We have quite a number of items that can by synced in SyncLab.
It might be a good idea to optimize the Pane, so that a user can more easily select the items to sync.

The existing pane looks like this:

#### Ideas
Maybe something with a wider width might work.

ccyccyccy:
Fixes #1681 partially.

**Outline of Solution**
Allows more item to be shown to the user by reducing the vertical margins of each item in the list.
Also made text images more sharp by upscaling the image.

Side by side comparison below. Left side is after this PR, right side is the original.

image

However, there might be better ways of allowing more items to be viewed at once, and other improvements as suggested in #1681.

### DIFF
--- a/PowerPointLabs/PowerPointLabs/SyncLab/SyncFormatConstants.cs
+++ b/PowerPointLabs/PowerPointLabs/SyncLab/SyncFormatConstants.cs
@@ -12,7 +12,7 @@ namespace PowerPointLabs.SyncLab.ObjectFormats
     public class SyncFormatConstants
     {
 
-        public static readonly Size DisplayImageSize = new Size(30, 30);
+        public static readonly Size DisplayImageSize = new Size(50, 50);
         
         // values for bevel display
         public static readonly float DisplayImageDepth = 15;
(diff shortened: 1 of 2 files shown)


iskandarzulkarnaien:
Relatively small change, but it does make the sync lab pane far less annoying to use since it reduces the amount of scrolling the user has to do.

Something I noticed is that the previews seem to be very low resolution.

image

Perhaps fixing this could be a quality of life improvement for the UI.

ccyccyccy:
> Something I noticed is that the previews seem to be very low resolution.

I just made some changes. Do you think it's better now?

iskandarzulkarnaien:
> 
> I just made some changes. Do you think it's better now?

Great! Looking a lot better now. Are you planning to add on anything else to this PR?

xantho09:
Looks good to me. The margins are still sufficiently large enough such that the items don't feel too cramped.

ccyccyccy:
> Great! Looking a lot better now. Are you planning to add on anything else to this PR?

Nope. I think other improvements require a more thorough overhaul of the design, which can be better done in other PRs.

PR #2020, opened 2020-04-02
#2014 Empty clipboard exception

### LINKED ISSUE #2014: ShapeLab: Library import fails if clipboard is empty
Importing a library in ShapeLab fails if the clipboard is empty.

ccyccyccy:
Fixes #2014 partially. This fix only affects ShapeLab for now.

**Outline of Solution**
Resolve issue by removing unneeded usage of pasting from clipboard.

There might be more places that needs this fix as mentioned in #2014.


iskandarzulkarnaien:
Works for me on 2016. Tests pass too.

Issue #2027, opened 2020-05-10
CropToShape does not work properly on PPT2010 on non-default ratio


daekoon:
#### Environment
PowerPoint 2010, Windows
#### Steps to replicate
1. Create a powerpoint slide with 16:9 ratio.
2. Try creating an image and using CropToShape on it.
![image](https://user-images.githubusercontent.com/42311564/81497973-b4ba3880-92f4-11ea-8885-32e92c8c2daf.png)

#### Expected behavior
CropToShape crops the area selected by the shape.
![image](https://user-images.githubusercontent.com/42311564/81497976-bab01980-92f4-11ea-9eea-a2c6660a0aaf.png)

#### Actual behavior
Area cropped is off-center by a bit. (To the right)
![image](https://user-images.githubusercontent.com/42311564/81497981-c0a5fa80-92f4-11ea-8a27-4e996c242ae9.png)

PR #2028, opened 2020-05-10
#1994 Fix CropLab CropToShape on PowerPoint 2010

### LINKED ISSUE #1994: Crop Lab FT failing on PowerPoint 2010
PPT 2010, Windows 10

`FT_AutoCropTest` does not pass on PPT2010. `FT_EffectsLabTest` has noticeably different behavior but still passes.

daekoon:
Fixes #1994

**Outline of Solution**
The CropToShape functionality on PowerPoint 2010 was not functioning properly. The cropped shape ended up being slightly off center, and the image had slightly different dimension as well. This resulted in the FT for AutoCropTest failing.

This has been fixed by implementing a multiplier to the height/width/X axis to minimize the difference caused. This fix should make the Crop functionality on PowerPoint 2010 work just like the later versions.

However, this fix does not fix the following:
- #2027
- EffectsLab having different behavior, which is mentioned in #1994


damithc:
Guys, this PR needs reviewing.

